PR #55, opened 2017-06-30
Rewrite AMQP common transport to use state machines

pierreca:
# Description of the problem
The current implementation of the common AMQP layer has no state management whatsoever for AMQP objects (connection, links, etc.).

# Description of the solution
I've rewritten large parts of the code to introduce state machines and added, where relevant, the ability to automatically connect or establish links if needed by the requested operation (For example, a call to `send` will now automatically connect and create the proper link.

This is the first step to lay down the foundations required to implement a solid retry logic, which will be at the client level.

A few notable things: 
- There should be no breaking changes at all in this PR.
- The `AmqpReceiver` object is retired in favor of the `ReceiverLink` object: we still export it as `AmqpReceiver` though so this is not a breaking change.
- I've introduced an `accept` method in the `ReceiverLink` which is what used to be called `complete`. `complete` still exists and redirects to `accept`. It's a terminology thing. I think we should do that all the way to the top of the transport layer and deprecate the incorrect `complete` terminology.  What I have not done (and probably should) is also rename `abandon` to `release`. Feedback more than welcome on those points, I'm not hell-bent on making this happen now...
- There is now a `SenderLink` object that is the symmetrical of the `ReceiverLink`
- The implementation of Claims Based Security (CBS) has been separated into its own state machine.


common/transport/amqp/devdoc/amqp_cbs_requirements.md
```diff
@@ -0,0 +1,100 @@
...
+    });
+  }
+});
+```
+
+## Public Interface
+
+### constructor(amqpClient)
+
+**SRS_NODE_AMQP_CBS_16_001: [** The `constructor` shall instanciate a `SenderLink` object for the `$cbs` endpoint using a custom policy `{encoder: function(body) { return body;}}` which forces the amqp layer to send the token as an amqp value in the body. **]**
```
anthonyvercolano: >instanciate [](start = 56, length = 11)

instantiate

common/transport/amqp/devdoc/amqp_requirements.md
```diff
@@ -117,44 +115,23 @@ Configures an AmqpReceiver object to use the endpoint passed as a parameter and
...
-**SRS_NODE_COMMON_AMQP_06_020: [** If given as an argument, the `initializeCBS` method shall call `initializeCBSCallback` with a null error object if successful. **]**
+# putToken(audience, token, putTokenCallback)
 
-# $cbs listener
+**SRS_NODE_COMMON_AMQP_16_040: [** The `putToken` method shall call the internal `ClaimsBasedSecurityAgent` instance `putToken` method with the same arguments. **]**
 
-**SRS_NODE_COMMON_AMQP_06_013: [** A put token response of 200 will invoke `putTokenCallback` with null parameters. **]**
-**SRS_NODE_COMMON_AMQP_06_014: [** A put token response not equal to 200 will invoke `putTokenCallback` with an error object of UnauthorizedError. **]**
-**SRS_NODE_COMMON_AMQP_06_012: [** All responses shall be completed. **]**
+**SRS_NODE_COMMON_AMQP_16_041: [** If the state machine is in a `detached` state, the `putToken` client shall connect the client and get into the `connected` state before proceeding with the nor **]**mal operations.
```
anthonyvercolano: >with the nor **]**mal operations. [](start = 182, length = 33)

You want to fix this.

common/transport/amqp/devdoc/receiver_link_requirements.md
```diff
@@ -0,0 +1,104 @@
...
+
+The `ReceiverLink` class is internal to the SDK and shall be considered private. It shall not be used directly by clients of the SDK, who should instead rely on the higher-level constructs (`Client` classes provided by `azure-iot-device` and `azure-iothub`).
+
+The `ReceiverLink` class implements a state machine that manages the underlying `amqp10` link object used to receive messages from IoT Hub. It can be attached and detached manually, and will try to attach automatically if not already attached when setting up a message listener or trying to settle an existing message.
+
+## Example usage
+
+```typescript
+import * as amqp10 from 'amqp10';
+import { AmqpMessage } from './ampq_message';
```
anthonyvercolano: >pq [](start = 33, length = 2)

transpose.

common/transport/amqp/devdoc/receiver_link_requirements.md
```diff
@@ -0,0 +1,104 @@
...
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_018: [** The `abandon` method shall use the link created by the underlying `amqp10.AmqpClient` to settle the specified `message` with IoT hub by abandoning it. **]**
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_024: [** The `abandon` method shall work whether a `callback` is specified or not, and call the callback with a `result.MessageAbandoned` object if a callback is specified. **]**
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_019: [** If the state machine is not in the `attached` state, the `abandon` method shall immediately fail with a `DeviceMessageLockLostError`. **]**
+
+### Events
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_011: [** If a `detached` or `errorReceived` event is emitted by the `ampq10` link object, the `ReceiverLink` object shall return to the `detached` state. **]**
```
anthonyvercolano: >pq [](start = 104, length = 2)

Transpose.

common/transport/amqp/devdoc/sender_link_requirements.md
```diff
@@ -0,0 +1,88 @@
...
+
+**SRS_NODE_AMQP_SENDER_LINK_16_012: [** If the message cannot be sent the `callback` shall be called with an `Error` object describing the AMQP error reported by the service. **]**
+
+**SRS_NODE_AMQP_SENDER_LINK_16_013: [** If the message is successfully sent, the `callback` shall be called with a first parameter (error) set to `null` and a second parameter of type `MessageEnqueued`. **]**
+
+**SRS_NODE_AMQP_SENDER_LINK_16_014: [** If the link is detached while a message is being sent, the `callback` shall be called with an `Error` object describing the AMQP error that caused the detach to happen in the first place. **]**
+
+### Events
+
+**SRS_NODE_AMQP_SENDER_LINK_16_015: [** If a `detached` or `errorReceived` event is emitted by the `ampq10` link object, the `SenderLink` object shall return to the `detached` state. **]**
```
anthonyvercolano: >pq [](start = 102, length = 2)

transpose

common/transport/amqp/devdoc/receiver_link_requirements.md
```diff
@@ -0,0 +1,104 @@
...
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_005: [** If the `amqp10.AmqpClient` emits an `errorReceived` event during the time the link is attached, the `callback` function shall be called with this error. **]**
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_006: [** The `ReceiverLink` object should subscribe to the `detached` event of the newly created `amqp10` link object. **]**
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_007: [** The `ReceiverLink` object should subscribe to the `errorReceived` event of the newly created `amqp10` link object. **]**
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_008: [** If the `amqp10.AmqpClient` fails to create the link the `callback` function shall be called with this error object. **]**
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_020: [** The `attach` method shall call the `callback` if the link was successfully attached **]**
```
anthonyvercolano: >`callback` [](start = 77, length = 10)

Specify null argument on success?

common/transport/amqp/devdoc/sender_link_requirements.md
```diff
@@ -0,0 +1,88 @@
...
+
+The `SenderLink` class is internal to the SDK and shall be considered private. It shall not be used directly by clients of the SDK, who should instead rely on the higher-level constructs (`Client` classes provided by `azure-iot-device` and `azure-iothub`).
+
+The `SenderLink` class implements a state machine that manages the underlying `amqp10` link object used to send messages to IoT Hub. It can be attached and detached manually, and will try to attach automatically if not already attached when calling `send`.
+
+## Example usage
+
+```typescript
+import * as amqp10 from 'amqp10';
+import { AmqpMessage } from './ampq_message';
```
anthonyvercolano: >pq [](start = 33, length = 2)

transpose

common/transport/amqp/src/amqp.ts
```diff
@@ -132,6 +88,185 @@ export class Amqp {
...
+    this._fsm = new machina.Fsm({
+      initialState: 'disconnected',
+      states: {
+        disconnected: {
+          connect: (connectCallback) => {
+            this._fsm.transition('connecting');
+            let connectError = null;
+            const connectErrorHandler = (err) => {
+              connectError = err;
+            };
```
anthonyvercolano: Has this function been tested?

common/transport/amqp/src/amqp.ts
```diff
@@ -132,6 +88,185 @@ export class Amqp {
...
+          },
+          disconnect: (callback) => callback(null, new results.Disconnected()),
+          attachSenderLink: (endpoint, linkOptions, callback) => {
+            this._fsm.handle('connect', (err) => {
+              if (err) {
+                callback(err);
+              } else {
+                this._fsm.handle('attachSenderLink', endpoint, linkOptions, callback);
+              }
+            });
```
anthonyvercolano: General comment.  I guess I would have done the above actual .connect call in the on handler of the connecting state.  I would have had all the appropriate handlers transition to the connecting and in connecting defer unit transition to connected.

common/transport/amqp/src/amqp.ts
```diff
@@ -132,6 +88,185 @@ export class Amqp {
...
+                  done(null, this._receivers[endpoint]);
+                }
+              });
+            } else {
+              /*Codes_SRS_NODE_COMMON_AMQP_16_009: [If a receiver for this endpoint has already been created, the getReceiver method should call the done() method with the existing instance as an argument.] */
+              this._safeCallback(done, null, this._receivers[endpoint]);
+            }
+          },
+          attachReceiverLink: (endpoint: string, linkOptions: any, done: GenericAmqpBaseCallback): void => {
+            let receiverFsm = new ReceiverLink(endpoint, linkOptions, this._amqp);
```
anthonyvercolano: >Fsm [](start = 24, length = 3)

You may call it an FSM, and it may have a private member that is an fsm, but that really doesn't seem to make this an fsm.

common/transport/amqp/src/amqp.ts
```diff
@@ -132,6 +88,185 @@ export class Amqp {
...
+            this._receivers[endpoint] = receiverFsm;
```
anthonyvercolano: >receiverFsm [](start = 40, length = 11)

And why not just do the new here and dispense with the local that has a very limited lifespan. (Like two lines).

common/transport/amqp/src/amqp.ts
```diff
@@ -205,38 +306,39 @@ export class Amqp {
...
-    let openLinks = [];
     for (let senderEndpoint in this._senders) {
       if (this._senders.hasOwnProperty(senderEndpoint)) {
-        openLinks.push(this._senders[senderEndpoint]);
+        this._senders[senderEndpoint].detach();
         delete this._senders[senderEndpoint];
       }
     }
-    self._senders = {};
+    this._senders = {};
```
anthonyvercolano: >this._senders = {}; [](start = 4, length = 19)

After the for wouldn't this really be unnecessary?

common/transport/amqp/src/amqp.ts
```diff
@@ -205,38 +306,39 @@ export class Amqp {
    * @param {Function}   done   Called when disconnected of if an error happened.
    */
   disconnect(done: GenericAmqpBaseCallback): void {
-    const self = this;
+    this._fsm.handle('disconnect', done);
+  }
 
+  _disconnect(done: GenericAmqpBaseCallback): void {
```
anthonyvercolano: >_disconnect [](start = 2, length = 11)

Why have this here rather than in the state machine?

common/transport/amqp/src/amqp_cbs.ts
```diff
@@ -0,0 +1,313 @@
...
+  private _errorHandler: (err: Error)  => void;
+  private _fsm: machina.Fsm;
+  private _senderLink: SenderLink;
+  private _receiverLink: ReceiverLink;
+  private _putToken: PutTokenStatus = new PutTokenStatus();
+  private _putTokenQueue: {
+    audience: string,
+    token: string,
+    callback: (err?: Error) => void
+  }[];
```
anthonyvercolano: interface?

common/transport/amqp/src/amqp_cbs.ts
```diff
@@ -0,0 +1,313 @@
...
+    if (!token) {
+      throw new ReferenceError('token cannot be \'' + token + '\'');
+    }
+
+    this._fsm.handle('putToken', audience, token, putTokenCallback);
+  }
+
+  private _removeExpiredPutTokens(): void {
+    const currentTime = Math.round(Date.now() / 1000);
+    let expiredPutTokens: PutTokenOperation[] = [];
```
anthonyvercolano: For later, but this timeout stuff should probably interact with whatever timeout modules we add in forthcoming pull requests.

common/transport/amqp/src/receiver_link.ts
```diff
@@ -0,0 +1,208 @@
...
+import { EventEmitter } from 'events';
+import { Message, errors, results } from 'azure-iot-common';
+import { AmqpMessage } from './amqp_message';
+import { AmqpLink } from './amqp_link_interface';
+
+const debug = dbg('ReceiverLink');
+
+/*Codes_SRS_NODE_AMQP_RECEIVER_LINK_16_002: [** The `ReceiverLink` class shall inherit from `EventEmitter`.]*/
+/*Codes_SRS_NODE_AMQP_RECEIVER_LINK_16_003: [** The `ReceiverLink` class shall implement the `AmqpLink` interface.]*/
+export class ReceiverLink  extends EventEmitter implements AmqpLink {
```
anthonyvercolano: >ReceiverLink [](start = 13, length = 12)

So whoever is using this ReceiverLink has no way of knowing that it got detached?

common/transport/amqp/src/amqp.ts
```diff
@@ -132,6 +88,185 @@ export class Amqp {
         forever: false
       }
     }, amqp10.Policy.EventHub));
+
+    this._fsm = new machina.Fsm({
+      initialState: 'disconnected',
+      states: {
+        disconnected: {
```
anthonyvercolano: >disconnected [](start = 8, length = 12)

So I note that you are not using a queue for handles like you do in the link code.  ?

common/transport/amqp/src/sender_link.ts
```diff
@@ -0,0 +1,207 @@
...
+    /*Codes_SRS_NODE_AMQP_SENDER_LINK_16_004: [The `attach` method shall use the stored instance of the `amqp10.AmqpClient` object to attach a new link object with the `linkAddress` and `linkOptions` provided when creating the `SenderLink` instance.]*/
+    this._amqp10Client.createSender(this._linkAddress, this._linkOptions)
+      .then((amqp10link) => {
+        if (!connectionError) {
+          debug('Sender object created for endpoint: ' + this._linkAddress);
+          this._linkObject = amqp10link;
+          /*Codes_SRS_NODE_AMQP_SENDER_LINK_16_006: [The `SenderLink` object should subscribe to the `detached` event of the newly created `amqp10` link object.]*/
+          this._linkObject.on('detached', this._detachHandler);
+          /*Codes_SRS_NODE_AMQP_SENDER_LINK_16_007: [The `SenderLink` object should subscribe to the `errorReceived` event of the newly created `amqp10` link object.]*/
+          this._linkObject.on('errorReceived', this._errorHandler);
```
anthonyvercolano: >this._linkObject.on('errorReceived', this._errorHandler) [](start = 10, length = 56)

Do you need to remove this at some point?

common/transport/amqp/src/sender_link.ts
```diff
@@ -0,0 +1,207 @@
...
+  private _linkOptions: any;
+  private _linkObject: any;
+  private _fsm: machina.Fsm;
+  private _amqp10Client: amqp10.AmqpClient;
+  private _unsentMessageQueue: MessageOperation[];
+  private _pendingMessageQueue: MessageOperation[];
+  private _detachHandler: (detachEvent: any) => void;
+  private _errorHandler: (err: Error) => void;
+
+  constructor(linkAddress: string, linkOptions: any, amqp10Client: amqp10.AmqpClient) {
```
anthonyvercolano: >constructor(linkAddress: string, linkOptions: any, amqp10Client: amqp10.AmqpClient) { [](start = 2, length = 85)

How do I create a sender link such that any messages I send over THAT link are send pre-settled?

common/transport/amqp/devdoc/receiver_link_requirements.md
```diff
@@ -0,0 +1,104 @@
+# ReceiverLink Requirements
+
+## Overview
+
+The `ReceiverLink` class is internal to the SDK and shall be considered private. It shall not be used directly by clients of the SDK, who should instead rely on the higher-level constructs (`Client` classes provided by `azure-iot-device` and `azure-iothub`).
+
+The `ReceiverLink` class implements a state machine that manages the underlying `amqp10` link object used to receive messages from IoT Hub. It can be attached and detached manually, and will try to attach automatically if not already attached when setting up a message listener or trying to settle an existing message.
```
anthonyvercolano: >implements [](start = 25, length = 10)

Surfacing this information in the spec is nice I suppose, but actually doesn't really do anything for the spec.

While the Recevier Link utilizes a fsm that has zero affect on the api surface supplied.

Frankly I would remove any mention of it here.

common/transport/amqp/devdoc/amqp_cbs_requirements.md
```diff
@@ -0,0 +1,100 @@
+# ReceiverLink Requirements
+
+## Overview
+
+The `ReceiverLink` class is internal to the SDK and shall be considered private. It shall not be used directly by clients of the SDK, who should instead rely on the higher-level constructs (`Client` classes provided by `azure-iot-device` and `azure-iothub`).
+
+The `ReceiverLink` class implements a state machine that manages the underlying `amqp10` link object used to receive messages from IoT Hub. It can be attached and detached manually, and will try to attach automatically if not already attached when setting up a message listener or trying to settle an existing message.
+
```
anthonyvercolano: Bit of a cut and paste problem here.  Receiver link in cbs interface.

common/transport/amqp/devdoc/amqp_cbs_requirements.md
```diff
@@ -0,0 +1,100 @@
...
+**SRS_NODE_AMQP_CBS_16_005: [** The `attach` method shall set up a listener for responses to put tokens on the `message` event of the receiver link. **]**
+
+**SRS_NODE_AMQP_CBS_16_006: [** If given as an argument, the `attach` method shall call `callback` with a standard `Error` object if any link fails to attach. **]**
+
+**SRS_NODE_AMQP_CBS_16_007: [** If given as an argument, the `attach` method shall call `callback` with a `null` error object if successful. **]**
+
+### detach()
+
+**SRS_NODE_AMQP_CBS_16_008: [** `detach` shall detach both sender and receiver links and return the state machine to the `detached` state. **]**
```
anthonyvercolano: >e state machine to the `detached` state. [](start = 98, length = 40)

You never mention states anywhere else.  Did you want to say that the constructor starts the fsm in the detached state?

You also never discuss leaving the detached state for any other state.

Is it worth putting in the transition diagram for the FSM?

common/transport/amqp/devdoc/receiver_link_requirements.md
```diff
@@ -0,0 +1,104 @@
...
+    }
+  });
+});
+```
+
+## Public Interface
+
+### constructor(linkAddress: string, linkOptions: any, amqp10Client: amqp10.AmqpClient)
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_001: [** The `ReceiverLink` internal state machine shall be initialized in the `detached` state. **]**
```
anthonyvercolano: >`detached`  [](start = 112, length = 11)

Do you need an FSM diagram? You really never discuss your state trasitions at all.

common/transport/amqp/devdoc/sender_link_requirements.md
```diff
@@ -0,0 +1,88 @@
...
+
+**SRS_NODE_AMQP_SENDER_LINK_16_014: [** If the link is detached while a message is being sent, the `callback` shall be called with an `Error` object describing the AMQP error that caused the detach to happen in the first place. **]**
+
+### Events
+
+**SRS_NODE_AMQP_SENDER_LINK_16_015: [** If a `detached` or `errorReceived` event is emitted by the `ampq10` link object, the `SenderLink` object shall return to the `detached` state. **]**
+
+### Internal state machine
+
+**SRS_NODE_AMQP_SENDER_LINK_16_016: [** If returning to the `detached` state because of an error that didn't happen while trying to attach the link or send a message, the sender link shall call emit an `error` event with that error. **]**
```
anthonyvercolano: >the sender link shall call emit an `error` event with that error.  [](start = 167, length = 66)

This is proving to be a difficult issue with the twin code.  Consider not emitting errors from the links.  Really seems as though they should be emitted from only one object and that is the client.

The link could emit the detach, but emitting errors is a pretty profound action which will likely result in program termination.

common/transport/amqp/src/amqp.ts
```diff
@@ -132,6 +88,185 @@ export class Amqp {
...
+
+            const amqpErrorHandler = (err) => {
+              debug('amqp10 error: ' + err.toString());
+              if (this._disconnectHandler) {
+                this._disconnectHandler(err);
+              }
+            };
+
+            this._amqp.on(_amqpClientError, connectErrorHandler);
+            this._amqp.connect(this.uri)
```
anthonyvercolano: >his._amqp.connect(this.uri) [](start = 13, length = 27)

You have a connecting state why not use it?

common/transport/amqp/src/amqp.ts
```diff
@@ -205,38 +306,39 @@ export class Amqp {
...
-          })
-          .catch((err) => {
-            /*SRS_NODE_COMMON_AMQP_16_005: [The disconnect method shall call the done callback and pass the error as a parameter if the disconnection is unsuccessful] */
-            self._safeCallback(done, err);
-          });
+    this._receivers = {};
+
+    if (this._cbs) {
+      this._cbs.detach();
+    }
```
anthonyvercolano: So, was it really best that the above detaches in the for loop killed the senders and receivers out from under the cbs before the cbs detach code did it?

common/transport/amqp/src/amqp.ts
```diff
@@ -132,6 +88,185 @@ export class Amqp {
...
+              disconnectCallback(err);
+            });
+          },
+          initializeCBS: (callback) => {
+            this._cbs = new ClaimsBasedSecurityAgent(this._amqp);
+            this._cbs.attach(callback);
+          },
+          putToken: (audience, token, callback) => {
+            this._cbs.putToken(audience, token, callback);
+          },
```
anthonyvercolano: Is it reasonable that all of the other operations that the code is doing in this connected state can happen if you're still waiting for the initializeCBS and the PutToken to finish?  Should you be in an Authenticated State?

common/transport/amqp/src/amqp.ts
```diff
@@ -132,6 +88,185 @@ export class Amqp {
...
+            let senderFsm = new SenderLink(endpoint, linkOptions, this._amqp);
+            this._senders[endpoint] = senderFsm;
+            this._senders[endpoint].on('error', (err) => {
+              debug('error on the sender link for endpoint ' + endpoint);
+              debug(err.toString());
+              delete(this._senders[endpoint]);
+            });
+            this._senders[endpoint].attach(done);
+          },
+          detachReceiverLink: (endpoint: string, detachCallback: GenericAmqpBaseCallback): void => {
```
anthonyvercolano: >detachReceiverLink: (endpoint: string, detachCallback: GenericAmqpBaseCallback): void => { [](start = 10, length = 90)

In general, I feel REALLY uneasy that we have these basic sender and receive links.  We then compose other objects made up of these links. (twins, method, CBS).  These composed objects though can have the links deleted out from under them. 

I'm not going to wait on it, because this is working.

anthonyvercolano:
:shipit:

common/transport/amqp/src/amqp.ts
pierreca: fair but we cant just defer until connected since connecting could go back to disconnected and in that case we'd still need to call the attachSenderLink callback with a failure. which we do here. I'll try and see if I can find a clever way to deal with this.

---
In reply to: [127617317](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r127617317) [](ancestors = 127617317)

common/transport/amqp/src/amqp.ts
pierreca: yes

---
In reply to: [127617005](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r127617005) [](ancestors = 127617005)

common/transport/amqp/devdoc/receiver_link_requirements.md
pierreca: actually, it's undefined. I've discovered recently that using null instead of undefined is a little frowned upon: [https://basarat.gitbooks.io/typescript/docs/tips/null.html](https://basarat.gitbooks.io/typescript/docs/tips/null.html)

---
In reply to: [127616011](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r127616011) [](ancestors = 127616011)

common/transport/amqp/src/sender_link.ts
pierreca: not sure, is this relevant to this PR specifically?

---
In reply to: [127619965](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r127619965) [](ancestors = 127619965)

common/transport/amqp/devdoc/sender_link_requirements.md
pierreca: the amqp10 folks came up with this idea and I remember a discussion with them telling me they regret it. it's our job to subscribe to the error event on the links, not the user's. to a certain extent, the link object itself shouldn't even be exposed to the user. therefore, i have no problem emitting the standard error event and differentiating based on type or error code. The client will handle that error (with a retry for example) and decide if it is necessary to bubble it up to the user or not.


---
In reply to: [128120964](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r128120964) [](ancestors = 128120964)

common/transport/amqp/devdoc/amqp_cbs_requirements.md
pierreca: diagrams would indeed be nice.

---
In reply to: [128119333](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r128119333) [](ancestors = 128119333)

common/transport/amqp/src/sender_link.ts
pierreca: yes.

---
In reply to: [127619882](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r127619882) [](ancestors = 127619882)

common/transport/amqp/src/sender_link.ts
anthonyvercolano: I currently have no ability to send a message pre-setttled.  This certainly is a "issue" in the twin code.  Not a big issue since the service side does the right thing anyway, but it does cause unnecessary traffic.

---
In reply to: [129156918](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r129156918) [](ancestors = 129156918,127619965)

pierreca: so it is a separate problem. a new feature. I'm thinking you need to look at amqp10, figure out what it takes to do that, and then suggest modifications to this interface while it's not yet public API

---
In reply to: [129158714](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r129158714) [](ancestors = 129158714,129156918,127619965)

common/transport/amqp/devdoc/receiver_link_requirements.md
pierreca: diagram would be nice but that sounds more like implementation notes than requirements. I'm also not sure about the necessity to talk about states in the requirements. I may remove this.

---
In reply to: [128120483](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r128120483) [](ancestors = 128120483)

anthonyvercolano: I could certainly support removing the state information.

---
In reply to: [129159537](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r129159537) [](ancestors = 129159537,128120483)

common/transport/amqp/src/amqp.ts
pierreca: you're right it's bothering me too. ideally we'd have only a list of the higher level constructs. unfortunately that list is split between the common AMQP object and the device/service AMQP object, which i'm not happy with. i hope i'll be able to tackle that when I tackle the higher level, but for now i don't have a good idea to solve this.

---
In reply to: [128132936](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r128132936) [](ancestors = 128132936)

anthonyvercolano: And this is why I approved.  I feel certain that this will be resolved in the fullness of time.

---
In reply to: [129160272](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r129160272) [](ancestors = 129160272,128132936)

common/transport/amqp/src/amqp.ts
pierreca: the problem is that authentication is managed at the higher layer (device/service AMQP class). mainly because the "authenticated" state varies from one to the other (one has x509 support but not the other for example). 
I think of this layer as mapping to the AMQP primitives. I wish I could displace the send to the link object and thenonly have APIs to create/destroy links and high level objects (CBS, etc) here and then let the higher layer deal with authentication completely but that'd be a breaking change.

it's clearly not perfect and I've gone back and forth on this quite a bit - i may still change stuff as I refactor the higher layer.

---
In reply to: [128129311](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r128129311) [](ancestors = 128129311)

common/transport/amqp/src/amqp_cbs.ts
pierreca: probably. right now it's nicely self-contained though, so not touching it.

---
In reply to: [127619196](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r127619196) [](ancestors = 127619196)

common/transport/amqp/devdoc/receiver_link_requirements.md
pierreca: I'll take care of it.

---
In reply to: [129159982](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r129159982) [](ancestors = 129159982,129159537,128120483)

common/transport/amqp/devdoc/amqp_cbs_requirements.md
pierreca: removing the state information everywhere.

---
In reply to: [129157953](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r129157953) [](ancestors = 129157953,128119333)

common/transport/amqp/src/receiver_link.ts
pierreca: When the link is detached it gets an error that we then forward in the _onEnter handler of the detached state. at that point an error event is emitted. so whoever is using the link needs to subscribe to the error event to know that it is detached. if the error event is not subscribed to the program will throw (which I think is the correct behavior).


---
In reply to: [127619355](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r127619355) [](ancestors = 127619355)

common/transport/amqp/src/amqp.ts
pierreca: we step into the connecting state in the connect function

---
In reply to: [128124325](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r128124325) [](ancestors = 128124325)

common/transport/amqp/src/amqp_cbs.ts
pierreca: I don't see a need to separate this interface declaration considering it's the only place where it's used but i'll do it if you feel it's better.

---
In reply to: [127618978](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r127618978) [](ancestors = 127618978)

common/transport/amqp/src/amqp.ts
pierreca: I've put a lot more thought to this today, trying to find a better design, because honestly this was designed lazily at first. turns out lazy might actually be good, here's what I have in mind to explain why: 

the main challenge here is that we want to call the action-specific callback on connection error or action error, but only on action success if connection succeeds: Leaving it here allows to group the error and success case of the connect action in a very readable manner, whereas using state transitions and queues split the call of the callback in multiple places (in _onEnter of the disconnected state for connection errors and in _onEnter of the connected state for connection success, where we'd handle the actual action). It also requires complicated queue logic (significantly more complicated than for link objects, because  we'd have many different types of actions in the same queue, whereas links have only one type of action).

whether we call "handle('connect', ...)" or "transition('connecting', ...)" and use the _onEnter handler is a different question: in that case, I feel both are mostly equivalent and valid ways to write that code. What tips the scale for me is that I like the idea that the transition to connecting/connected is all done in the body of the connect action handler, and I like how easy it is to read (looks more imperative as opposed to transitioning to the "connecting" state and assuming the reader knows that the connection happens just because we transitioned).

Lastly, there's the question of what happens when we start a bunch of actions in a row (eg. 2 attachSenderLink and a send): should we fail all of them if the first one fails to connect, or should we try to connect for each action that was requested by the caller? Since the retry logic implemented by the caller may be different for each type of action though, i'm not sure it makes sense to "group fail" everything if the first action's call to connect fails. To rephrase, intuitively, I think each action "deserves a shot" at connecting.
To the argument that we have a different pattern for link attach and connect, I think it's acceptable because connecting is a lot more error prone than attaching links: it depends on more external factors (if the connection is up and authenticated, attaching the link should succeed, attaching doesn't depend on wifi quality, etc: the connection does) so it also feels ok treating connect and attach differently.

It's all work in progress though and those arguments might not stand the test of reimplementing the client layer. I guess we'll see, but feel free to challenge those assumptions: it's a complex piece of code.


---
In reply to: [129133265](https://github.com/Azure/azure-iot-sdk-node/pull/55#discussion_r129133265) [](ancestors = 129133265,127617317)

PR #57, opened 2017-07-06
Fix member variable name typo in Result interface

kurtb:
# Checklist

# Description of the problem
The d.ts files included with this library fail to build on TypeScript 2.4 due to the new weak types (https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript). This change to TypeScript is a good one because it actually catches a bug with the current code. The base Result interface has a typo with its transportObj field. Derived classes don't have the typo. Older TypeScript compilers treated it as a second field declaration and built fine. But in 2.4 you need overlap with a weak type and so now it will fail to build given the typo.

# Description of the solution
Fixes the typo.

### DIFF
--- a/common/core/src/results.ts
+++ b/common/core/src/results.ts
@@ -5,7 +5,7 @@
 'use strict';
 
 export interface Result {
-    tranportObj?: any;
+    transportObj?: any;
 }
 
 /**


kurtb:
I submitted #57 which should fix the issue. The 2.4 release of the TypeScript compiler added some stricter checking which found a simple typo bug with the base Result interface.

anthonyvercolano:
Merging into master.

PR #61, opened 2017-07-10
Add support for CA signed certs.

anthonyvercolano:
# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/common/core/src/endpoint.ts
+++ b/common/core/src/endpoint.ts
@@ -4,7 +4,7 @@
 
 'use strict';
 
-export const apiVersion = '2016-11-14';
+export const apiVersion = '2017-06-30';
 
 export function devicePath(id: string): string {
   return '/devices/' + id;
--- a/e2etests/e2etests.js
+++ b/e2etests/e2etests.js
@@ -41,16 +41,25 @@ device_provision(hubConnectionString, function (err, provisionedDevices) {
     console.log('Unable to create the devices needed.');
   } else {
     provisionedDevices.forEach(function(deviceToTest) {
-      acknowledgementProtocols.forEach(function (protocolToTest) {
-        device_acknowledge_tests(hubConnectionString, protocolToTest, deviceToTest);
+      if (deviceToTest.authenticationDescription !== 'CA signed certificate') {
+        acknowledgementProtocols.forEach(function (protocolToTest) {
+          device_acknowledge_tests(hubConnectionString, protocolToTest, deviceToTest);
         });
-      generalProtocols.forEach(function(protocolToTest) {
-        device_service_tests(hubConnectionString, protocolToTest, deviceToTest);
-      });
+      }
+      if (deviceToTest.authenticationDescription !== 'CA signed certificate') {
+        generalProtocols.forEach(function(protocolToTest) {
+          device_service_tests(hubConnectionString, protocolToTest, deviceToTest);
+        });
+      }
     });
 
-    // In the interest of saving time, we only will perform the connection
-    // tests on the shared key device.
+    // CA certs don't work for http and ca cert chains don't work for web sockets.
+    assert.equal(provisionedDevices[3].authenticationDescription, 'CA signed certificate');
+    device_service_tests(hubConnectionString, deviceMqtt.Mqtt, provisionedDevices[3]);
+    device_service_tests(hubConnectionString, deviceAmqp.Amqp, provisionedDevices[3]);
+    device_acknowledge_tests(hubConnectionString, deviceAmqp.Amqp, provisionedDevices[3]);
+    //In the interest of saving time, we only will perform the connection
+    //tests on the shared key device.
     assert.equal(provisionedDevices[1].authenticationDescription, 'shared private key');
     c2d_disconnect(hubConnectionString, provisionedDevices[1]);
     d2c_disconnect(hubConnectionString, provisionedDevices[1]);
@@ -71,7 +80,7 @@ device_provision(hubConnectionString, function (err, provisionedDevices) {
   authentication_tests(hubConnectionString);
 
   device_teardown(hubConnectionString, provisionedDevices);
-  if (!provisionedDevices || provisionedDevices.length !== 3) {
+  if (!provisionedDevices || provisionedDevices.length !== 4) {
     describe('device creation did not', function() {
       it('completely work', function(done) {
           done(new Error(''));
--- a/e2etests/test/device_provision.js
+++ b/e2etests/test/device_provision.js
@@ -18,6 +18,9 @@ var chalk = require('chalk');
 var device_provision = function (hubConnectionString, done) {
   var provisionedDevices = [];
   var registry = Registry.fromConnectionString(hubConnectionString);
+  var CARootCert = new Buffer(process.env.IOTHUB_CA_ROOT_CERT, 'base64').toString('ascii');
+  var CARootCertKey = new Buffer(process.env.IOTHUB_CA_ROOT_CERT_KEY, 'base64').toString('ascii');
+
   var host = ConnectionString.parse(hubConnectionString).HostName;
 
   function setupDevice(deviceDescription, provisionDescription, done) {
@@ -31,7 +34,7 @@ var device_provision = function (hubConnectionString, done) {
       }
     });
   }
-  
+
   function createCertDevice(deviceId, done) {
     var certOptions = {
       selfSigned: true,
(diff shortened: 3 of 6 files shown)


common/core/src/endpoint.ts
```diff
@@ -4,7 +4,7 @@
 
 'use strict';
 
-export const apiVersion = '2016-11-14';
+export const apiVersion = '2017-06-30';
```
tameraw: >2017-06-30 [](start = 27, length = 10)

If that's not deployed yet, we need to wait before checking this in.

service/devdoc/registry_requirements.md
```diff
@@ -61,6 +61,26 @@ The `fromSharedAccessSignature` static method returns a new instance of the `Reg
 
 ## CRUD operation for the device registry
 
+### normalize authentication
+With 2017-06-30 api release crud device information MUST contain authentication specification
+
+**SRS_NODE_IOTHUB_REGISTRY_06_028: [** A device information with no authentication will be normalize with the following authentication:
```
tameraw: >normalize [](start = 91, length = 9)

nit/pic: normalized

tameraw:
:shipit:

common/core/src/endpoint.ts
anthonyvercolano: Yup.

---
In reply to: [127058566](https://github.com/Azure/azure-iot-sdk-node/pull/61#discussion_r127058566) [](ancestors = 127058566)

service/devdoc/registry_requirements.md
anthonyvercolano: Fiexed.

---
In reply to: [127059528](https://github.com/Azure/azure-iot-sdk-node/pull/61#discussion_r127059528) [](ancestors = 127059528)

e2etests/e2etests.js
```diff
@@ -41,12 +41,16 @@ device_provision(hubConnectionString, function (err, provisionedDevices) {
     console.log('Unable to create the devices needed.');
   } else {
     provisionedDevices.forEach(function(deviceToTest) {
-      acknowledgementProtocols.forEach(function (protocolToTest) {
-        device_acknowledge_tests(hubConnectionString, protocolToTest, deviceToTest);
+      if (deviceToTest.authenticationDescription !== 'shared private key') {
```
pierreca: The test matrix is becoming more complicated and by adding this kind of conditions we make it harder to maintain. The pattern that separates device provisioning from the tests themselves is starting to show cracks.

also, if we were to stick with this, string description sounds like a bad choice to solve a type problem - enum values would work better.

e2etests/test/device_provision.js
```diff
@@ -72,13 +76,58 @@ var device_provision = function (hubConnectionString, done) {
     });
   }
 
+  function createCACertDevice(deviceId, done) {
+    var certOptions = {
```
pierreca: doesn't look like this variable is used anywhere.

service/src/registry.ts
```diff
@@ -735,6 +739,36 @@ export class Registry {
...
+      ] */
+      deviceInfo.authentication = {
+          type: 'sas',
+          symmetricKey: {
+            primaryKey: '',
+            secondaryKey: ''
+          }
+        };
+    /* Codes_SRS_NODE_IOTHUB_REGISTRY_06_029: [** A device information with an authentication object that contains a `type` property is considered normalized.] */
+    } else if (!deviceInfo.authentication.hasOwnProperty('type')) {
```
pierreca: nitpicking: all those `!` everywhere make the code slightly harder to read - I'd rather see positive criteria evaluation than having to negate everything

service/test/_registry_test.js
```diff
@@ -796,9 +817,22 @@ describe('Registry', function() {
       <stringified array supplied by the argument devices annotated with importMode property and deviceId property replaced by id>
       ```
       ]*/
+      /* Tests_SRS_NODE_IOTHUB_REGISTRY_06_029: [** A device information with an authentication object that contains a `type` property is considered normalized.] */
+      /* Tests_SRS_NODE_IOTHUB_REGISTRY_06_031: [A device information with an authentication object that doesn't contain the x509Thumbprint property will be normalized with a `type` property with value "sas".] */
+      /* Tests_SRS_NODE_IOTHUB_REGISTRY_06_030: [A device information with an authentication object that contains the x509Thumbprint property will be normalized with a `type` property with value "selfSigned".] */
       it('constructs a valid HTTP request', function(testCallback) {
-        var addedDevices = [{deviceId: 'deviceone'},{deviceId: 'devicetwo'}];
-        var sentBody = [{id:'deviceone',importMode:'create'},{id:'devicetwo',importMode:'create'}];
+        var addedDevices = [
```
pierreca: please use only one array and make the sentBody a property of the device description rather than separate into two arrays and rely on the idea that the next developer will know that the content of these two arrays are tighly coupled.

e2etests/test/device_provision.js
anthonyvercolano: Removed

---
In reply to: [128300745](https://github.com/Azure/azure-iot-sdk-node/pull/61#discussion_r128300745) [](ancestors = 128300745)

service/test/_registry_test.js
anthonyvercolano: I respectfully disagree.

First, adding the sentBody into the device description really complicates from a reading perspective what we have here.

Second, each of these paired things is restricted to a single test.  It is easily read that via their contents that they are linked.

Third, this not something new just thought up.  This paradigm was already in the tests. 

---
In reply to: [128303235](https://github.com/Azure/azure-iot-sdk-node/pull/61#discussion_r128303235) [](ancestors = 128303235)

e2etests/e2etests.js
anthonyvercolano: One didn't have to add these conditions if one didn't mind adding 33% to the time taken to perform this section of the tests.  That's A LOT of time.

I minded.

This is an e2e test program.  Not a public API that we expect some third party to have to consume.  There is NO gain in adding the enumeration.

---
In reply to: [128297732](https://github.com/Azure/azure-iot-sdk-node/pull/61#discussion_r128297732) [](ancestors = 128297732)

pierreca: I like performance improvements as much as the next guy but i don't like bad/hard to maintain code and one doesn't have to come at the expense of the other, which is what is happening here. You may check this in as is but it doesn't change the fact that an overhaul of this part of the code is needed. I guess you just don't want to take care of it.

service/src/registry.ts
pierreca: please leave an explanation as to why this is a better design than the requested change. Being dismissive is not gonna help your case.

service/test/_registry_test.js
pierreca: First is just a code formatting issue - my comment is based on code structure, i'm sure you can find a creative and readable way to format this.

Second and third don't invalidate the fact that we have unnecessary tight coupling of array content - and that it will require more effort for the maintainer to understand what's going on if you don't make that change.

So please make that change.

service/src/registry.ts
anthonyvercolano: My reply is:

1) You prefaced your comment with nitpick.  I view nitpick as something of no great importance as viewed by the author of the comments.

2) The negative predicates are there because that is how the decisions (at least in my mind) needed to make this code work arise.  Therefore to me it is perfectly reasonable to write the code that way.

3) In this particular case these are EXACTLY the correct questions to ask with precisely the correct sequence.


---
In reply to: [128658002](https://github.com/Azure/azure-iot-sdk-node/pull/61#discussion_r128658002) [](ancestors = 128658002)

e2etests/e2etests.js
anthonyvercolano: I do not dispute that this area is ripe for rethinking.  That however is NOT what this pull request is about.  I would of course be happy to discuss this with you.  There are certainly good arguments to be made for simplification to especially foster faster e2e test runs.

---
In reply to: [128657867](https://github.com/Azure/azure-iot-sdk-node/pull/61#discussion_r128657867) [](ancestors = 128657867)

service/test/_registry_test.js
anthonyvercolano: It is not at though we are comparing individual elements of 1 array to elements of the other.  The entire ARRAY of sent body is compare to what is created from addedDevices.  To put these together makes to me incorrect.

I actually have no idea what you think would be a better layout of some combined structure.  Please use this particular test to provide me with what you think is better.  When you do, I will be happy to change the others.

---
In reply to: [128658351](https://github.com/Azure/azure-iot-sdk-node/pull/61#discussion_r128658351) [](ancestors = 128658351)

PR #80, opened 2017-09-09
Documentation cleanup

pierreca:
- Removed a bunch of classes that appeared in the docs even though they are not public API
- Improved some of the documentation comments.

### DIFF
--- a/device/core/src/twin.ts
+++ b/device/core/src/twin.ts
@@ -13,6 +13,20 @@ import { translateError } from './twin_errors';
 import { Client } from './client';
 import { TwinTransport } from './interfaces';
 
+/**
+ * A Device Twin is document describing the state of a device that is stored by an Azure IoT hub and is available even if the device is offline.
+ * It is built around 3 sections:
+ *   - Tags: key/value pairs only accessible from the service side
+ *   - Desired Properties: updated by a service and received by the device
+ *   - Reported Properties: updated by the device and received by the service.
+ *
+ * Note that although it is a possibility, desired and reported properties do not have to match
+ * and that the logic to sync these two collections, if necessary, is left to the user of the SDK.
+ *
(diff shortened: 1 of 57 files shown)


anthonyvercolano:
Looks good.  I realize that these are documentation updates, we are coming along and saying these are private.  Do we need to think of these as breaking changes?

device/core/src/twin.ts
```diff
@@ -21,18 +34,31 @@ export class Twin extends EventEmitter {
...
+   * The desired and reported properties dictionnaries (respectively in `properties.desired` and `properties.reported`).
+   */
   properties: any; // TODO: need type
 
   private _rid: number;
   private _client: Client;
   private _receiver: any; // TODO: need type
 
+  /**
+   * The constructor should not be used directly and instead the SDK user should use the {@link Client#getTwin} method to obtain a valid `Twin` object.
```
anthonyvercolano: >method to obtain a valid `Twin` object. [](start = 112, length = 39)

Although below you say gettwin is private.

device/transport/http/src/http_errors.ts
```diff
@@ -16,6 +19,9 @@ export class HttpTransportError extends Error {
  * - the body of  the HTTP response, containing the explanation of why the request failed
  * - the HTTP response object itself]
  */
+/**
+ * @private
```
anthonyvercolano: Seems odd.

pierreca: the `@private` comment just removes this from the web page. since this is not used by SDK users, no need to have it in the docs.

device/core/src/twin.ts
pierreca: ? Client.getTwin(), Registry.getTwin() and Twin.get() are all public.

pierreca:
No breaking changes since we're not hiding any APIs. we're just removing from the docs whatever was not exported from our packages and was just clutter for SDK users.

PR #81, opened 2017-09-12
Add common retry logic objects: RetryPolicy, RetryOperation, and default policies

pierreca:
# Description of the problem
Currently the SDK does not implement any sort of retry logic for operations that can fail because of transient errors such as network issues. This is the beginning of a series of commits aimed at solving that problem

# Description of the solution
Right now this only contains the objects that will be used by the high level clients to retry operations. Next up are the new versions of the transport packages.

### DIFF
--- a/common/core/devdoc/retry_policy_requirements.md
+++ b/common/core/devdoc/retry_policy_requirements.md
@@ -1,0 +1,76 @@
+# azure-iot-common.RetryPolicy Requirements
+
+## Overview
+
+The `RetryPolicy` interface and classes implement the necessary logic to compute retry intervals using different algorithms and parameters when an operation (such as sending a D2C message) fails.
+
+## Example
+
+```js
+var policy = new ExponentialBackoffWithJitter(new DefaultErrorFilter());
+var firstTimeout = policy.nextRetryTimeout(1, false);
+var secondTimeout = policy.nextRetryTimeout(2, false);
+
+// the second parameter indicates whether the algorithm should use normal parameters (false)
+// or parameters adapted to situations where the Azure IoT hub is throttling
+var timeoutWhenThrottled = policy.nextRetryTimeout(1, true);
+// timeoutWhenThrottled will be longer than firstTimeout.
+
+var errorFilter: {
+  TimeoutError: false
+};
+
+var policy = new ExponentialBackoffWithJitter(errorFilter);
+var shouldBeFalse = policy.shouldRetry(new errors.TimeoutError());
+```
+
+## Public API:
+
+### RetryPolicy interface
+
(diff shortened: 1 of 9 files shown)


common/core/devdoc/retry_policy_requirements.md
```diff
@@ -0,0 +1,76 @@
...
+#### shouldRetry(error)
+
+**SRS_NODE_COMMON_RETRY_POLICY_16_006: [** The `shouldRetry` method of the new instance shall use the error filter passed to the constructor when the object was instantiated. **]**
+
+#### getNextTimeout(currentRetryCount, isThrottled)
+
+**SRS_NODE_COMMON_RETRY_POLICY_16_007: [** The `getNextTimeout` method shall implement the following math formula to determine the next timeout value: `F(x) = min(Cmin+ (2^(x-1)-1) * rand(C * (1 – Jd), C*(1-Ju)), Cmax` **]**
+
+**SRS_NODE_COMMON_RETRY_POLICY_16_008: [** The `getNextTimeout` method shall return `0` instead of the result of the math formula if the following 3 conditions are met:
+- the `constructor` was called with the `immediateFirstTimeout` boolean set to `true`
```
anthonyvercolano: >the `constructor` was called with the `immediateFirstTimeout` boolean set to `true` [](start = 2, length = 83)

This pops up with no previous explanation in these docs.

pierreca: retry_policy.ts line 97

common/core/src/retry_policy.ts
```diff
@@ -0,0 +1,179 @@
...
+  private _errorFilter: ErrorFilter;
+
+  /**
+   * Initializes a new instance of the {@link azure-iot-common.ExponentialBackOffWithJitter} class.
+   * @param maxTimeout            maximum allowed timeout in milliseconds.
+   * @param immediateFirstRetry   boolean indicating whether the first retry should be immediate (default) or wait the first interval (c value).
+   */
+  constructor(immediateFirstRetry?: boolean, errorFilter?: ErrorFilter, ) {
+    this._errorFilter = errorFilter ? errorFilter : new DefaultErrorFilter();
+    this.immediateFirstRetry = immediateFirstRetry !== false ? true : false;    // should default to true if not specified.
```
anthonyvercolano: >true : false [](start = 63, length = 12)

Needed?

pierreca: its a bit awkward but the idea is that it should be true by default, unless it's specifically set to false. so "undefined", "null" or any other value really should count as true.

pierreca: shall be corrected with `immediateFirstRetry !== false` (and remove the ?: expression)

PR #83, opened 2017-09-16
State machines for amqp base package

pierreca:
Introduce state machines for the base AMQP layer. greatly increases connection stability and another step towards automated retries.


common/transport/amqp/package.json
```diff
@@ -7,24 +7,25 @@
   "main": "main.js",
   "typings": "index.d.ts",
   "dependencies": {
-    "amqp10": "3.5.0",
+    "@types/async": "^2.0.42",
+    "amqp10": "^3.5.3",
```
BertKleewein: >^3.5.3 [](start = 15, length = 6)

We should talk about use of the carat and tilde operator in package.json.  For stability, I wonder if we should lock to specific versions.  (Yes, I know how semver and these operators _should_ work, but real life isn't always as clean, especially with packages that have limited support)

common/transport/amqp/devdoc/receiver_link_requirements.md
```diff
@@ -0,0 +1,98 @@
...
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_006: [** The `ReceiverLink` object should subscribe to the `detached` event of the newly created `amqp10` link object. **]**
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_007: [** The `ReceiverLink` object should subscribe to the `errorReceived` event of the newly created `amqp10` link object. **]**
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_008: [** If the `amqp10.AmqpClient` fails to create the link the `callback` function shall be called with this error object. **]**
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_020: [** The `attach` method shall call the `callback` if the link was successfully attached **]**
+
+### detach(): void
```
BertKleewein: AmqpLink.detach accepts a callback according to amqp_link_interface.ts

common/transport/amqp/devdoc/receiver_link_requirements.md
```diff
@@ -0,0 +1,98 @@
...
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_009: [** The `detach` method shall detach the link created by the `amqp10.AmqpClient` underlying object. **]**
+
+### accept(message: AmqpMessage, callback: (err?: Error, result?: results.MessageAccepted) => void): void
```
BertKleewein: >callback [](start = 33, length = 8)

TS syntax question: If callback is optional, shouldn't this be callback?: (err...blah) ?

common/transport/amqp/devdoc/receiver_link_requirements.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Public Interface
+
+### constructor(linkAddress: string, linkOptions: any, amqp10Client: amqp10.AmqpClient)
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_002: [** The `ReceiverLink` class shall inherit from `EventEmitter`. **]**
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_003: [** The `ReceiverLink` class shall implement the `AmqpLink` interface. **]**
+
+### attach(callback: (err?: Error) => void): void
```
BertKleewein: >callback [](start = 11, length = 8)

is callback required?

common/transport/amqp/devdoc/receiver_link_requirements.md
```diff
@@ -0,0 +1,98 @@
...
+
+### accept(message: AmqpMessage, callback: (err?: Error, result?: results.MessageAccepted) => void): void
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_021: [** The `accept` method shall throw if the `message` argument is falsy. **]**
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_013: [** The `accept` method shall use the link created by the underlying `amqp10.AmqpClient` to settle the specified `message` with IoT hub by accepting it. **]**
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_022: [** The `accept` method shall work whether a `callback` is specified or not, and call the callback with a `result.MessageCompleted` object if a callback is specified. **]**
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_014: [** If the state machine is not in the `attached` state, the `accept` method shall immediately fail with a `DeviceMessageLockLostError`. **]**
```
BertKleewein: When a link drops, who is responsible for moving this state machine back into the detached state?  Who is responsible when the connection drops?

common/transport/amqp/devdoc/receiver_link_requirements.md
```diff
@@ -0,0 +1,98 @@
...
+
+### complete(message: AmqpMessage, callback: (err?: Error, result?: results.MessageAccepted) => void): void
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_015: [** The `complete` method shall call the `accept` method with the same arguments (it is here for backward compatibility purposes only). **]**
+
+### reject(message: AmqpMessage, callback: (err?: Error, result?: results.MessageRejected) => void): void
```
BertKleewein: >callback [](start = 33, length = 8)

optional?

common/transport/amqp/devdoc/receiver_link_requirements.md
```diff
@@ -0,0 +1,98 @@
...
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_021: [** The `accept` method shall throw if the `message` argument is falsy. **]**
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_013: [** The `accept` method shall use the link created by the underlying `amqp10.AmqpClient` to settle the specified `message` with IoT hub by accepting it. **]**
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_022: [** The `accept` method shall work whether a `callback` is specified or not, and call the callback with a `result.MessageCompleted` object if a callback is specified. **]**
+
+**SRS_NODE_AMQP_RECEIVER_LINK_16_014: [** If the state machine is not in the `attached` state, the `accept` method shall immediately fail with a `DeviceMessageLockLostError`. **]**
+
+### complete(message: AmqpMessage, callback: (err?: Error, result?: results.MessageAccepted) => void): void
```
BertKleewein: >callback [](start = 35, length = 8)

optional?

common/transport/amqp/devdoc/receiver_link_requirements.md
BertKleewein: missing syntax and/or requirements for optional callbacks all over the place in here.  I'm not going to mark them all unless you want me too.

---
In reply to: [139493462](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139493462) [](ancestors = 139493462)

common/transport/amqp/devdoc/sender_link_requirements.md
```diff
@@ -0,0 +1,78 @@
...
+
+**SRS_NODE_AMQP_SENDER_LINK_16_022: [** The `attach` method shall call the `callback` if the link was successfully attached. **]**
+
+**SRS_NODE_AMQP_SENDER_LINK_16_006: [** The `SenderLink` object should subscribe to the `detached` event of the newly created `amqp10` link object. **]**
+
+**SRS_NODE_AMQP_SENDER_LINK_16_007: [** The `SenderLink` object should subscribe to the `errorReceived` event of the newly created `amqp10` link object. **]**
+
+**SRS_NODE_AMQP_SENDER_LINK_16_008: [** If the `amqp10.AmqpClient` fails to create the link the `callback` function shall be called with this error object. **]**
+
+### detach(): void
```
BertKleewein: >(): [](start = 10, length = 3)

accepts callback

common/transport/amqp/devdoc/sender_link_requirements.md
```diff
@@ -0,0 +1,78 @@
...
+senderLink.send(new AmqpMessage(''), linkAddress, (err) => {
+  if (err) {
+    console.error(err.toString());
+  } else {
+    console.log('message successfully sent');
+  }
+});
+```
+
+## Public Interface
```
BertKleewein: >Interface [](start = 10, length = 9)

missing requirements for forceDetach function in here.

common/transport/amqp/devdoc/sender_link_requirements.md
```diff
@@ -0,0 +1,78 @@
...
+
+### detach(): void
+
+**SRS_NODE_AMQP_SENDER_LINK_16_009: [** The `detach` method shall detach the link created by the `amqp10.AmqpClient` underlying object. **]**
+
+### send(message: AmqpMessage, callback: (err?: Error, result?: results.MessageEnqueued) => void): void
+
+**SRS_NODE_AMQP_SENDER_LINK_16_010: [** The `send` method shall use the link created by the underlying `amqp10.AmqpClient` to send the specified `message` to the IoT hub. **]**
+
+**SRS_NODE_AMQP_SENDER_LINK_16_011: [** If the state machine is not in the `attached` state, the `SenderLink` object shall attach the link first and then send the message. **]**
```
BertKleewein: >If the state machine is not in the `attached` state, the `SenderLink` object shall attach the link first and then send the message. [](start = 40, length = 131)

This is inconsistent with the complete, reject, and abandon functions on the receiver link object.  Why does this attach the link, but those functions don't?

common/transport/amqp/devdoc/sender_link_requirements.md
```diff
@@ -0,0 +1,78 @@
...
+
+**SRS_NODE_AMQP_SENDER_LINK_16_013: [** If the message is successfully sent, the `callback` shall be called with a first parameter (error) set to `null` and a second parameter of type `MessageEnqueued`. **]**
+
+**SRS_NODE_AMQP_SENDER_LINK_16_014: [** If the link is detached while a message is being sent, the `callback` shall be called with an `Error` object describing the AMQP error that caused the detach to happen in the first place. **]**
+
+**SRS_NODE_AMQP_SENDER_LINK_16_019: [** While the link isn't attached, the messages passed to the `send` method shall be queued. **]**
+
+**SRS_NODE_AMQP_SENDER_LINK_16_020: [** When the link gets attached, the messages shall be sent in the order they were queued. **]**
+
+**SRS_NODE_AMQP_SENDER_LINK_16_021: [** If the link fails to attach and there are messages in the queue, the callback for each message shall be called with the error that caused the detach in the first place. **]**
```
BertKleewein: >callback for each message shall be called [](start = 109, length = 41)

I assume the messages get removed from the queue in this case, right?  Should that be spelled out here?

common/transport/amqp/devdoc/sender_link_requirements.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Events
+
+**SRS_NODE_AMQP_SENDER_LINK_16_016: [** If an error happened that caused the link to be detached, the sender link shall call emit an `error` event with that error. **]**
```
BertKleewein: In my mind, this is a transient error.  The link was detached, but sending a message will re-attach it, so I don't think any action is required by the caller in this case.  Should our event have a different name to indicate this?  Otherwise, the caller won't know what to do in response to this.

common/transport/amqp/devdoc/sender_link_requirements.md
```diff
@@ -0,0 +1,78 @@
...
+
+**SRS_NODE_AMQP_SENDER_LINK_16_018: [** If an error happened that caused the link to be detached while trying to attach the link or send a message, the `callback` for this function shall be called with that error. **]**
```
BertKleewein: >the `callback` for this function shall be called with that error. [](start = 148, length = 65)

Ambiguous: In this case, is the error event also fired?  I'm 99% sure the answer is no, but do we want to be explicit in here?

common/transport/amqp/src/amqp.ts
```diff
@@ -75,14 +30,13 @@ class PutTokenStatus {
  * @param   {String}    sdkVersionString        String identifying the SDK used (device or service).
```
BertKleewein: >(device or service) [](start = 80, length = 19)

confusing.  this implies that the value could be "device" or "service", but I think it's more qualified than that.

common/transport/amqp/src/amqp.ts
```diff
@@ -135,6 +88,280 @@ export class Amqp {
         forever: false
       }
     }, amqp10.Policy.EventHub));
+
+    this._fsm = new machina.Fsm({
+      namespace: sdkVersionString,
```
BertKleewein: >namespace: sdkVersionString, [](start = 6, length = 28)

Interesting choice.  Is this arbitrary or does it serve a purpose?

common/transport/amqp/src/amqp.ts
```diff
@@ -135,6 +88,280 @@ export class Amqp {
...
+      initialState: 'disconnected',
+      states: {
+        disconnected: {
+          _onEnter: (disconnectCallback, err, result) => {
+            if (err) {
+              this._safeCallback(disconnectCallback, err);
```
BertKleewein: >this._safeCallback(disconnectCallback, err); [](start = 14, length = 44)

(This is more of a meta question.  My examples in here my be not correct, but I suspect we'll see this pattern all over the code and I want to make sure it's right)
I may be wrong here, but the parameter "disconnectCallback" doesn't necessarily come from a call to disconnect(), right?  If the connection gets disconnected calling some other function, the code _might_ end up here with disconnectCallback being the callback that was passed into that other function.  If that's the case, should we pass results.Disconnected() as the third parameter to the callback so that the caller is able to tell the final result of that operation.  Or are they responsible for figuring out that their operation caused a disconnect via some other means (such as checking the connection status)?

common/transport/amqp/src/amqp.ts
```diff
@@ -135,6 +88,280 @@ export class Amqp {
...
+            } else {
+              this._safeCallback(disconnectCallback, null, new results.Disconnected());
+            }
+          },
+          amqpError: (err, callback) => { callback(); }, // we're already considering ourselves disconnected. it'd be a bug though?
```
BertKleewein: >it'd be a bug though? [](start = 110, length = 21)

This could either be a race condition (error arrives right after the disconnect completed), or a big screw up (lower levels are in a different state than upper levels think they are.  Code typically can't recover from this very nicely if we're not careful))

common/transport/amqp/src/amqp.ts
```diff
@@ -135,6 +88,280 @@ export class Amqp {
...
+          disconnect: (disconnectCallback) => {
+            this._fsm.transition('disconnecting', disconnectCallback);
+          },
+          initializeCBS: (callback) => {
+            this._cbs = new ClaimsBasedSecurityAgent(this._amqp);
+            this._cbs.attach(callback);
+          },
+          putToken: (audience, token, callback) => {
+            this._cbs.putToken(audience, token, callback);
+          },
```
BertKleewein: Should there be a separate state for CBS initializing and CBS initialized?

common/transport/amqp/src/amqp.ts
```diff
@@ -135,6 +88,280 @@ export class Amqp {
...
+
+          },
+          getReceiver: (endpoint: string, done: GenericAmqpBaseCallback<ReceiverLink>): void => {
+            /*Codes_SRS_NODE_COMMON_AMQP_16_010: [If a receiver for this endpoint doesn’t exist, the getReceiver method should create a new AmqpReceiver object and then call the done() method with the object that was just created as an argument.] */
+            if (!this._receivers[endpoint]) {
+              this._fsm.handle('attachReceiverLink', endpoint, null, (err) => {
+                if (err) {
+                  done(err);
+                } else {
+                  done(undefined, this._receivers[endpoint]);
```
BertKleewein: use _safeCallback?  I assume done is option because you use _safeCallback below.

common/transport/amqp/src/amqp.ts
```diff
@@ -135,6 +88,280 @@ export class Amqp {
...
+              .catch((err) => {
+                this._amqp.removeListener(_amqpClientError, connectErrorHandler);
+                this._fsm.transition('disconnected');
+                /*Codes_SRS_NODE_COMMON_AMQP_16_003: [The `connect` method shall call the `done` callback if the connection fails.] */
+                this._safeCallback(connectCallback, connectError || err);
+              });
+          },
+          amqpError: (err, callback) => this._fsm.transition('disconnecting', callback, err),
+          connect: () => this._fsm.deferUntilTransition(),
+          disconnect: () => this._fsm.deferUntilTransition(),
```
BertKleewein: redundant since the wildcard below should handle this

common/transport/amqp/src/amqp.ts
```diff
@@ -135,6 +88,280 @@ export class Amqp {
...
+              this._safeCallback(done, null, this._receivers[endpoint]);
+            }
+          },
+          attachReceiverLink: (endpoint: string, linkOptions: any, done: GenericAmqpBaseCallback<ReceiverLink>): void => {
+            debug('creating receiver link for: ' + endpoint);
+            this._receivers[endpoint] = new ReceiverLink(endpoint, linkOptions, this._amqp);
+            const attachErrorHandler = (err) => {
+              debug('receiver link error while attaching: ' + endpoint + ': ' + err.toString());
+              delete(this._receivers[endpoint]);
+              done(err);
```
BertKleewein: >done [](start = 14, length = 4)

_safeCallback?

common/transport/amqp/src/amqp.ts
```diff
@@ -135,6 +88,280 @@ export class Amqp {
...
+              delete(this._senders[endpoint]);
+              done(err);
+            };
+
+            this._senders[endpoint].on('error', attachErrorHandler);
+            debug('attaching sender link for: ' + endpoint);
+            this._senders[endpoint].attach((err) => {
+              if (err) {
+                debug('failed to attach sender link: ' + endpoint + ': ' + err.toString());
+                done(err);
```
BertKleewein: >done [](start = 16, length = 4)

should you delete(this._senders[endpoint]) here?

BertKleewein: Also, are we guaranteed that the link is not attached?  Do we need to call detach to make absolutely sure that the link is deached?

---
In reply to: [139526430](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139526430) [](ancestors = 139526430)

common/transport/amqp/package.json
pierreca: yes, we should fixate on a specific version. also, i have ideas how to integrate dependencies updates and vulnerability scanning in our CI

---
In reply to: [139485216](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139485216) [](ancestors = 139485216)

common/transport/amqp/src/amqp.ts
```diff
@@ -135,6 +88,280 @@ export class Amqp {
...
+            }
+          },
+          detachSenderLink: (endpoint: string, detachCallback: GenericAmqpBaseCallback<any>): void => {
+            this._detachLink(this._senders[endpoint], (err?) => {
+              delete(this._senders[endpoint]);
+              this._safeCallback(detachCallback, err);
+            });
+          }
+        },
+        disconnecting: {
```
BertKleewein: >disconnecting [](start = 8, length = 13)

This block is difficult to read.  2 things slowed me down:
1) names disconnectFunc and detachFunc are confusing.  I knew disconnectFunc has a function that had something to do with disconnect, but I didn't know if it was a callback or the function that actually disconnected.  Suggest naming detachFunc to detachLink following the verb-subject pattern.  Same for disconnectFunc -- disconnectConnection doesn't sound right, but I don't know what is better.
2) The ternary operators you used in disconnectFunc and detachFunc are unusual.  I would expect the comparison to happen inside the function -- this is more typical and easier (for me) to grok.

common/transport/amqp/src/amqp.ts
```diff
@@ -135,6 +88,280 @@ export class Amqp {
...
+              callback();
+            } : (callback) => {
+              this._amqp.removeAllListeners('disconnected');
+              /*Codes_SRS_NODE_COMMON_AMQP_16_004: [The disconnect method shall call the done callback when the application/service has been successfully disconnected from the service] */
+              this._amqp.disconnect().then(() => {
+                callback();
+                return null;
+              }).catch((err) => {
+                callback(err);
+              });
```
BertKleewein: If we're disconnecting because of an error, don't we want to do this?  Otherwise, the connection seems like it might be only partially taken down.

common/transport/amqp/src/amqp.ts
```diff
@@ -135,6 +88,280 @@ export class Amqp {
...
+              if (link) {
+                link.forceDetach(err);
+              }
+              callback();
+            } : (link, callback) => {
+              if (link) {
+                link.detach(callback);
+              } else {
+                callback();
+              }
```
BertKleewein: If we're disconnecting because of an error, don't we still want to detach all links?

BertKleewein: Also, shouldn't we be removing the links from this._senders and this._receivers here?  

---
In reply to: [139532134](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139532134) [](ancestors = 139532134)

common/transport/amqp/devdoc/receiver_link_requirements.md
pierreca: yes it should

---
In reply to: [139489242](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139489242) [](ancestors = 139489242)

common/transport/amqp/src/amqp.ts
```diff
@@ -135,6 +88,280 @@ export class Amqp {
...
+            detachFunc(this._cbs, () => {
+              let remainingLinks = this._getRemainingLinks();
+              /*Codes_SRS_NODE_COMMON_AMQP_16_034: [The `disconnect` method shall detach all open links before disconnecting the underlying AMQP client.]*/
+              async.each(remainingLinks, detachFunc, () => {
+                disconnectFunc((err) => {
+                  if (err) {
+                    this._fsm.transition('disconnected', disconnectCallback, err);
+                  } else {
+                    this._fsm.transition('disconnected', disconnectCallback);
+                  }
```
BertKleewein: style question:  This block operates exactly the same as if you just called transition with 3 parameters without checking the value of err.  Is it better to be explicit or terse? (I don't know the answer)

common/transport/amqp/devdoc/receiver_link_requirements.md
pierreca: The Amqp object in the package handles the connection and detach/forceDetach the links as appropriate using the SenderLink/ReceiverLink APIs. these APIs will return the state machine to its appropriate state.

if it's link errors only, the error event received from the amqp10 object will trigger a forceDetach() which in turn will trigger a transition back to detached.

---
In reply to: [139493305](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139493305) [](ancestors = 139493305)

common/transport/amqp/src/amqp.ts
```diff
@@ -251,44 +413,8 @@ export class Amqp {
...
-          } else {
-            sendAction(sender, amqpMessage, done);
-          }
-        });
-      } else {
-        sendAction(this._senders[endpoint], amqpMessage, done);
-      }
-    }
+  send(message: Message, endpoint: string, to: string, done: GenericAmqpBaseCallback<any>): void {
+    this._fsm.handle('send', message, endpoint, to, done);
```
BertKleewein: >handle [](start = 14, length = 6)

If the user calls send() without ever calling connect(), where does the send error out?  I'm guessing it will be somewhere cryptic that will be difficult for the caller to track down.  Do we want an explicit check to make sure the user called connect() at least once before calling any of the other functions?

common/transport/amqp/devdoc/receiver_link_requirements.md
pierreca: looks like they should be optional to maintain backward compat - i'll add that.

---
In reply to: [139532964](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139532964) [](ancestors = 139532964,139489242)

common/transport/amqp/src/amqp.ts
```diff
@@ -454,232 +485,60 @@ export class Amqp {
...
-        correlationId: amqpMessage.properties.messageId
-      };
-
-      this._putToken.outstandingPutTokens.push(outstandingPutToken);
-      //
-      // If this is the first put token then start trying to time it out.
-      //
-      if (this._putToken.outstandingPutTokens.length === 1) {
-        this._putToken.timeoutTimer = setTimeout(this._removeExpiredPutTokens.bind(this), this._putToken.putTokenTimeOutExaminationInterval);
+  private _getRemainingLinks(): AmqpLink[] {
```
BertKleewein: >_getRemainingLinks [](start = 10, length = 18)

suggest better name.  This does more than just "get".  It also removes the link from the internal collection.  Surely that's significant enough to warrant a different verb.  (but I don't know what that verb would be :))

common/transport/amqp/src/amqp.ts
```diff
@@ -454,232 +485,60 @@ export class Amqp {
...
-    }
+
+    return remainingLinks;
   }
 
   /*Codes_SRS_NODE_COMMON_AMQP_16_011: [All methods should treat the `done` callback argument as optional and not throw if it is not passed as argument.]*/
-  private _safeCallback(callback: GenericAmqpBaseCallback, error?: Error | null, result?: any): void {
+  private _safeCallback(callback: GenericAmqpBaseCallback<any>, error?: Error | null, result?: any): void {
     if (callback) {
       process.nextTick(() => callback(error, result));
```
BertKleewein: >process.nextTick [](start = 6, length = 16)

uh oh, this suggests a work-around and scares me.  Why do you do this?  All callbacks lose their stack and that bothers me.
If this must be here, suggest a better function name like _deferredSafeCallback() so the caller knows what's going on (I read this whole file without knowing that the nextTick call was in here, and I'm worried that this was significant in understanding how the code functions.)

common/transport/amqp/devdoc/receiver_link_requirements.md
pierreca: yeah I'd rather make callbacks mandatory on all our lower-level APIs and only make them optional on the outer layer, public APIs. Since this is a new API, this is by design.

---
In reply to: [139490418](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139490418) [](ancestors = 139490418)

common/transport/amqp/devdoc/receiver_link_requirements.md
pierreca: yes

---
In reply to: [139493516](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139493516) [](ancestors = 139493516)

common/transport/amqp/devdoc/receiver_link_requirements.md
pierreca: look at 16_023 and corresponding requirements "should work whether a callback is specified or not..." etc

---
In reply to: [139494149](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139494149) [](ancestors = 139494149,139493462)

common/transport/amqp/src/amqp_cbs.ts
```diff
@@ -0,0 +1,323 @@
...
+ *                                                              a monotonically increasing time ordering.  This is effected by the nature
+ *                                                              of inserting new elements at the end of the array.  Note that elements may
+ *                                                              be removed from the array at any index.
+ * @property   {Number}     numberOfSecondsToTimeout            Currently a fixed value.  Could have a set option if we want to make this configurable.
+ * @property   {Number}     putTokenTimeOutExaminationInterval  While there are ANY put token operations outstanding a timer will be invoked every
+ *                                                              10 seconds to examine the outstandingPutTokens array for any put tokens that may have
+ *                                                              expired.
+ * @property   {Number}     timeoutTimer                        Timer used to trigger examination of the outstandingPutTokens array.
+ */
+class PutTokenStatus {
```
BertKleewein: >PutTokenStatus [](start = 6, length = 14)

PutToken is an unfortunate name.  It's not clear that Put is not a verb.  (I first thought this was a function that put the token status).  There's probably not a better way to name this, but I still have to ask.

common/transport/amqp/src/amqp_cbs.ts
```diff
@@ -0,0 +1,323 @@
...
+    token: string,
+    callback: (err?: Error) => void
+  }[];
+
+  constructor(amqp10Client: amqp10.AmqpClient) {
+    this._errorHandler = (err: Error): void => {
+      this._fsm.handle('detach', undefined, err);
+    };
+
+    this._amqp10Client = amqp10Client;
```
BertKleewein: >amqp10Client [](start = 25, length = 12)

add falsy check?

common/transport/amqp/devdoc/sender_link_requirements.md
pierreca: because the message lock is lost if the link is detached and therefore complete/reject/abandon cannot possibly succeed, whereas send can.

---
In reply to: [139494826](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139494826) [](ancestors = 139494826)

common/transport/amqp/devdoc/sender_link_requirements.md
pierreca: the fact that there is a queue in internal implementation details imho, but I'm not against spelling it out if you think it's better

---
In reply to: [139495305](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139495305) [](ancestors = 139495305)

common/transport/amqp/src/amqp.ts
pierreca: arbitrary during a debug phase... suggestions? could be simply "amqp" since links are "senderlink" and "receiverlink"

---
In reply to: [139500399](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139500399) [](ancestors = 139500399)

common/transport/amqp/src/amqp_cbs.ts
```diff
@@ -0,0 +1,323 @@
...
+                      for (let i = 0; i < this._putToken.outstandingPutTokens.length; i++) {
+                        if (msg.correlationId === this._putToken.outstandingPutTokens[i].correlationId) {
+                          const completedPutToken = this._putToken.outstandingPutTokens[i];
+                          this._putToken.outstandingPutTokens.splice(i, 1);
+                          if (completedPutToken.putTokenCallback) {
+                            /*Codes_SRS_NODE_AMQP_CBS_16_019: [A put token response of 200 will invoke `putTokenCallback` with null parameters.]*/
+                            let error = null;
+                            if (msg.properties.getValue('status-code') !== 200) {
+                              /*Codes_SRS_NODE_AMQP_CBS_16_018: [A put token response not equal to 200 will invoke `putTokenCallback` with an error object of UnauthorizedError.]*/
+                              error = new errors.UnauthorizedError(msg.properties.getValue('status-description'));
```
BertKleewein: >error = new errors.UnauthorizedError(msg.properties.getValue('status-description')); [](start = 30, length = 84)

This will still transition into attached.  Is that what you want?

common/transport/amqp/src/amqp.ts
pierreca: yup, I don't want to crash though, but I'll add a debug log line

---
In reply to: [139507770](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139507770) [](ancestors = 139507770)

common/transport/amqp/src/amqp_cbs.ts
```diff
@@ -0,0 +1,323 @@
...
+                }
+              }
+            });
+          }
+        },
+        detaching: {
+          _onEnter: (forwardedCallback, err) => {
+            /*Codes_SRS_NODE_AMQP_CBS_16_008: [`detach` shall detach both sender and receiver links and return the state machine to the `detached` state.]*/
+            this._senderLink.detach(() => {
+              this._receiverLink.detach(() => {
```
BertKleewein: Should you handle null _senderLink or _receiverLink?  I don 't find any easy failures to get into this state, but this could result in an exception or broken state.

common/transport/amqp/src/amqp_cbs.ts
```diff
@@ -0,0 +1,323 @@
...
+              callback: callback
+            });
+            this._fsm.transition('attaching');
+          }
+         },
+        attaching: {
+          _onEnter: (callback) => {
+            this._senderLink.attach((err) => {
+              if (err) {
+                this._fsm.transition('detached', callback, err);
```
BertKleewein: >this._fsm.transition('detached', callback, err); [](start = 16, length = 48)

I can see why you're going straight into detached here, but it might be worth going through the detaching state just to future proof your code.  (quick eyes might not notice that a failure on _senderLink goes into a different state than the failure on _receiverLink)

common/transport/amqp/src/amqp_cbs.ts
```diff
@@ -0,0 +1,323 @@
...
+                  if (err) {
+                    this._fsm.transition('detaching', callback, err);
+                  } else {
+                    this._receiverLink.on('message', (msg) => {
+                      //
+                      // Regardless of whether we found the put token in the list of outstanding
+                      // operations, accept it.  This could be a put token that we previously
+                      // timed out.  Be happy.  It made it home, just too late to be useful.
+                      //
+                      /*Codes_SRS_NODE_AMQP_CBS_16_020: [All responses shall be completed.]*/
```
BertKleewein: >completed [](start = 80, length = 9)

Shouldn't this say "all responses should be accepted"?  (I don't remember if AMQP has "complete" and "accept" as 2 different dispositions, but using the right word here might be significant)

common/transport/amqp/src/amqp.ts
pierreca: CBS state is handled with the CBS state machine. the connection itself doesn't depend on the CBS state so I don't think so. Plus, CBS isn't always used.

On the higher layer, we will separate disconnected/connecting/authenticating/authenticated/disconnecting

---
In reply to: [139511781](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139511781) [](ancestors = 139511781)

common/transport/amqp/src/amqp_cbs.ts
```diff
@@ -0,0 +1,323 @@
...
+                      }
+                    });
+                    this._fsm.transition('attached', callback);
+                  }
+                });
+              }
+            });
+          },
+          detach: (callback, err) => this._fsm.transition('detaching', callback, err),
+          forceDetach: () => this._fsm.transition('detached'),
```
BertKleewein: >forceDetach: () => this._fsm.transition('detached'), [](start = 10, length = 52)

Links could be attached here.  Don't you want to forceDetach the links here.?

common/transport/amqp/src/amqp_cbs.ts
```diff
@@ -0,0 +1,323 @@
...
+            /*Codes_SRS_NODE_AMQP_CBS_16_012: [The `putToken` method shall send this message over the `$cbs` sender link.]*/
+            this._senderLink.send(amqpMessage, (err) => {
+              if (err) {
+                // Find the operation in the outstanding array.  Remove it from the array since, well, it's not outstanding anymore.
+                // Since we may have arrived here asynchronously, we simply can't assume that it is the end of the array.  But,
+                // it's more likely near the end.
+                for (let i = this._putToken.outstandingPutTokens.length - 1; i >= 0; i--) {
+                  if (this._putToken.outstandingPutTokens[i].correlationId === amqpMessage.properties.messageId) {
+                    const outStandingPutTokenInError = this._putToken.outstandingPutTokens[i];
+                    this._putToken.outstandingPutTokens.splice(i, 1);
```
BertKleewein: Nice find!  I would have assumed that the token was the front of the queue

common/transport/amqp/src/amqp_common_errors.ts
```diff
@@ -102,6 +102,13 @@ export function translateError(message: string, amqpError: Error): AmqpTransport
     }
   } else if (amqpError instanceof Amqp10Errors.AuthenticationError) {
     error = new errors.UnauthorizedError(message);
+  } else if ((<any>amqpError).code) {
+    switch ((<any>amqpError).code) {
+      case 'ENOENT':
+      case 'ENOTFOUND':
+      default:
```
BertKleewein: OK.  Why have a switch at all?

common/transport/amqp/src/receiver_link.ts
```diff
@@ -0,0 +1,212 @@
...
+                }
+                return null;
+              })
+              .catch((err) => {
+                debug('amqp10 failed to create receiver: ' + err.toString());
+                this._fsm.transition('detached', callback, err);
+              });
+          },
+          detach: (callback) => this._fsm.transition('detaching', callback),
+          forceDetach: (err) => this._fsm.transition('detached', undefined, err),
```
BertKleewein: >forceDetach: (err) => this._fsm.transition('detached', undefined, err), [](start = 10, length = 71)

Links could be present.  I think forceDetach should always _try_ to close and null out variables, otherwise this object will be in an indeterminate state.

BertKleewein: by "links could be present", I mean "the amqp10link object could be attached here.  Not likely, but still possible.

---
In reply to: [139543130](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139543130) [](ancestors = 139543130)

common/transport/amqp/src/receiver_link.ts
```diff
@@ -0,0 +1,212 @@
...
+          accept: (message, callback) => callback(new errors.DeviceMessageLockLostError()),
+          reject: (message, callback) => callback(new errors.DeviceMessageLockLostError()),
+          abandon: (message, callback) => callback(new errors.DeviceMessageLockLostError())
```
BertKleewein: defer until attached? (dangerous since we don't know if the attach is actually going to succeed).

common/transport/amqp/src/receiver_link.ts
```diff
@@ -0,0 +1,212 @@
...
+
+  private _removeListeners(): void {
+    this._linkObject.removeListener('detached', this._detachHandler);
+    this._linkObject.removeListener('message', this._messageHandler);
+    this._linkObject.removeListener('errorReceived', this._errorHandler);
+  }
+
+  private _safeCallback(callback: (err?: Error, result?: any) => void, error?: Error | null, result?: any): void {
+    if (callback) {
+      process.nextTick(() => callback(error, result));
```
BertKleewein: >process.nextTick(() => callback(error, result)); [](start = 6, length = 48)

same comments on _safeCallback that I made in some other file.

common/transport/amqp/src/amqp.ts
pierreca: callbacks are not optional here

---
In reply to: [139525728](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139525728) [](ancestors = 139525728)

common/transport/amqp/src/sender_link.ts
```diff
@@ -0,0 +1,229 @@
...
+
+            /*Codes_SRS_NODE_COMMON_AMQP_16_011: [All methods should treat the `done` callback argument as optional and not throw if it is not passed as argument.]*/
+            let _processPendingMessageCallback = (error?, result?) => {
+              const opIndex = this._pendingMessageQueue.indexOf(op);
+              if (opIndex >= 0) {
+                this._pendingMessageQueue.splice(opIndex, 1);
+                if (op.callback) {
+                  /*Codes_SRS_NODE_AMQP_SENDER_LINK_16_013: [If the message is successfully sent, the `callback` shall be called with a first parameter (error) set to `null` and a second parameter of type `MessageEnqueued`.]*/
+                  /*Codes_SRS_NODE_AMQP_SENDER_LINK_16_012: [If the message cannot be sent the `callback` shall be called with an `Error` object describing the AMQP error reported by the service.]*/
+                  process.nextTick(() => callback(error, result));
```
BertKleewein: do you mean to call op.callback instead of callback?

common/transport/amqp/src/amqp.ts
pierreca: we should delete. if the link is attached at that point it's a bug in the link state machine.. not convinced we should be that defensive. since error handling and detaching is kinda complicated already i'd rather not do that, might come back to bite us if we ever change the state machine logic of the link?

---
In reply to: [139526967](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139526967) [](ancestors = 139526967,139526430)

common/transport/amqp/src/amqp.ts
pierreca: callback is not optional here. 

---
In reply to: [139524981](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139524981) [](ancestors = 139524981)

common/transport/amqp/src/sender_link.ts
```diff
@@ -0,0 +1,229 @@
...
+            };
+
+            this._pendingMessageQueue.push(op);
+
+            /*Codes_SRS_NODE_COMMON_AMQP_16_011: [All methods should treat the `done` callback argument as optional and not throw if it is not passed as argument.]*/
+            let _processPendingMessageCallback = (error?, result?) => {
+              const opIndex = this._pendingMessageQueue.indexOf(op);
+              if (opIndex >= 0) {
+                this._pendingMessageQueue.splice(opIndex, 1);
+                if (op.callback) {
```
BertKleewein: >if (op.callback) { [](start = 16, length = 18)

Suggest adding a comment that we only call the callback when the message was actually found in the queue because it could have been removed by an error handler and we don't want to call the callback twice (this is important but not readily apparent).

common/transport/amqp/src/amqp_cbs.ts
```diff
@@ -0,0 +1,323 @@
...
+            this._putToken.outstandingPutTokens.push(outstandingPutToken);
+            //
+            // If this is the first put token then start trying to time it out.
+            //
+            if (this._putToken.outstandingPutTokens.length === 1) {
+              this._putToken.timeoutTimer = setTimeout(this._removeExpiredPutTokens.bind(this), this._putToken.putTokenTimeOutExaminationInterval);
+            }
+            /*Codes_SRS_NODE_AMQP_CBS_16_012: [The `putToken` method shall send this message over the `$cbs` sender link.]*/
+            this._senderLink.send(amqpMessage, (err) => {
+              if (err) {
```
BertKleewein: >if (err) [](start = 14, length = 8)

It feels like we should be doing something here if the send succeeds. Shouldn't we call putTokenCallback at the very least?

common/transport/amqp/src/amqp_cbs.ts
```diff
@@ -0,0 +1,323 @@
...
+                // Find the operation in the outstanding array.  Remove it from the array since, well, it's not outstanding anymore.
+                // Since we may have arrived here asynchronously, we simply can't assume that it is the end of the array.  But,
+                // it's more likely near the end.
```
BertKleewein: >t's more likely near the end. [](start = 20, length = 29)

suggest you add a comment that the token might not even be in the array (if it expired).  That's not immediately obvious.

common/transport/amqp/src/amqp_common_errors.ts
pierreca: I wanted to look at the list of errors and discriminate but looks like this will have to wait ([https://nodejs.org/api/errors.html#errors_common_system_errors](https://nodejs.org/api/errors.html#errors_common_system_errors))

---
In reply to: [139542426](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139542426) [](ancestors = 139542426)

common/transport/amqp/src/receiver_link.ts
pierreca: if the link has been detached, this can't succeed (the lock for the message was lost)

---
In reply to: [139543686](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139543686) [](ancestors = 139543686)

common/transport/amqp/src/amqp.ts
pierreca: would be fine if we could rely on amqp10 to do the right thing if the connection is already disconnected or partially disconnected, but it crashes. so we have to assume that if there is an error, amqp10 already took down the connection and we can't call disconnect again.

---
In reply to: [139531957](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139531957) [](ancestors = 139531957)

common/transport/amqp/src/amqp.ts
pierreca: we want to forceDetach, not detach. if the connection has an error and has been disconnected amqp10 will crash if we call detach.

---
In reply to: [139532385](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139532385) [](ancestors = 139532385,139532134)

pierreca: we delete everything in _getRemainingLinks (put everything in one array and delete the references in the _senders and _receivers objects)

---
In reply to: [139532385](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139532385) [](ancestors = 139532385,139532134)

common/transport/amqp/src/amqp.ts
pierreca: I'm not sure I understand what you mean

---
In reply to: [139532979](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139532979) [](ancestors = 139532979)

common/transport/amqp/src/amqp.ts
pierreca: the problem when calling callbacks directly from promises is that any error in one of those callbacks end up getting caught by the .catch() clause of said promise, which screws up the stack way more than using process.nextTick. happy to hear suggestions though

---
In reply to: [139535307](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139535307) [](ancestors = 139535307)

common/transport/amqp/src/sender_link.ts
pierreca: if the message isn't found in the queue how would we call the callback? i'm not sure I understand the comment.

---
In reply to: [139546179](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139546179) [](ancestors = 139546179)

common/transport/amqp/src/amqp_cbs.ts
pierreca: we call the callback only when we get the putToken response on the CBS receiver link.

---
In reply to: [139546668](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139546668) [](ancestors = 139546668)

common/transport/amqp/src/receiver_link.ts
pierreca: same answer

---
In reply to: [139544184](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139544184) [](ancestors = 139544184)

common/transport/amqp/src/receiver_link.ts
pierreca: I don't think it can, because 1/ the link object is nulled when entering detached 2/ before "saving" it in a private instance member, the callback checks that it's still in "attaching" state meaning if somehow forceDetach was called during the "attaching" state we would not have a link to forceDetach.

---
In reply to: [139543378](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139543378) [](ancestors = 139543378,139543130)

common/transport/amqp/src/amqp_cbs.ts
pierreca: "Accept" is the right word and you are right. i've been trying to remove our use of "Complete" everywhere

---
In reply to: [139538944](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139538944) [](ancestors = 139538944)

common/transport/amqp/src/amqp_cbs.ts
pierreca: yeah that's not a bad idea..

---
In reply to: [139538455](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139538455) [](ancestors = 139538455)

common/transport/amqp/src/amqp_cbs.ts
pierreca: technically still attached.

---
In reply to: [139537583](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139537583) [](ancestors = 139537583)

common/transport/amqp/src/amqp_cbs.ts
pierreca: since this is an internal API the compiler does that for us :)

---
In reply to: [139536227](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139536227) [](ancestors = 139536227)

common/transport/amqp/src/amqp_cbs.ts
pierreca: that's just code that moved, i really don't want to refactor that at the same time

---
In reply to: [139535819](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139535819) [](ancestors = 139535819)

common/transport/amqp/src/amqp.ts
pierreca: it's used only once I'll inline the function and remove the naming problem.

---
In reply to: [139534573](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139534573) [](ancestors = 139534573)

common/transport/amqp/src/amqp.ts
pierreca: I hate the fact that we have a send() method here. SenderLink.send() should be used instead (and at some point we'll be able to get rid of the link collections).

---
In reply to: [139533766](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139533766) [](ancestors = 139533766)

pierreca: we can try to connect and send using the FSM but it's likely going to fail if not authenticated.

---
In reply to: [139567812](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139567812) [](ancestors = 139567812,139533766)

common/transport/amqp/devdoc/sender_link_requirements.md
pierreca: it's up to the caller to decide. in the case of methods for example we'd need to tear both links down and reestablish them. which is why it's probably a good idea to fire that error anyway.

---
In reply to: [139496095](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139496095) [](ancestors = 139496095)

common/transport/amqp/devdoc/sender_link_requirements.md
pierreca: you're right, the event wouldn't be fired because I don't want to feed the error back to the caller twice in 2 different ways.

---
In reply to: [139496419](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139496419) [](ancestors = 139496419)

common/transport/amqp/src/amqp.ts
pierreca: they shouldn't care that the connection is disconnected. Also, events signaling the connection state will come in a later feature. so they will have a separate way to know. using results.Disconnected here kindof breaks the pattern of having either an error or a result but not both.

---
In reply to: [139503297](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139503297) [](ancestors = 139503297)

common/transport/amqp/devdoc/sender_link_requirements.md
BertKleewein: Maybe this is intuitive, but I'm talking about the fact that caller will have to re-submit the message if the link fails to attach.  The AMQP object doesn't save the message to send it the next time the link attaches.

---
In reply to: [139537136](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139537136) [](ancestors = 139537136,139495305)

common/transport/amqp/src/amqp.ts
BertKleewein: good enough.  Just curious.

---
In reply to: [139537523](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139537523) [](ancestors = 139537523,139500399)

common/transport/amqp/src/amqp.ts
BertKleewein: Fair enough.

---
In reply to: [139538015](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139538015) [](ancestors = 139538015,139507770)

common/transport/amqp/src/amqp.ts
BertKleewein: It's a silly thing, but as written, if err is defined, it calls 
  this._fsm.transition('disconnected', disconnectCallback, err)
If err is falsy, it calls 
  this._fsm.transition('disconnected', disconnectCallback).

You could get the same behavior by calling 
  this._fsm.transition('disconnected', disconnectCallback, err)

Because, if err is falsy, 
  this._fsm.transition('disconnected', disconnectCallback, err)
is the same as 
  this._fsm.transition('disconnected', disconnectCallback)

(but only if the transition function treats all falsy values the same, which isn't a horrible assumption.)

It's OK to ignore this comment.  I won't be offended.


---
In reply to: [139561201](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139561201) [](ancestors = 139561201,139532979)

common/transport/amqp/src/amqp.ts
BertKleewein: Interesting.  I learned something new today,.  I'll have to think about the implications on .catch more.

---
In reply to: [139561431](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139561431) [](ancestors = 139561431,139535307)

common/transport/amqp/src/receiver_link.ts
BertKleewein: OK, fair enough.  I was thinking paranoid, but I trust your opinion here.

---
In reply to: [139563612](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139563612) [](ancestors = 139563612,139543378,139543130)

common/transport/amqp/src/sender_link.ts
BertKleewein: I think I was slightly confused and this comment resolves to "I was confused reading this.  Maybe someone else will be too."

---
In reply to: [139562522](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139562522) [](ancestors = 139562522,139546179)

common/transport/amqp/devdoc/sender_link_requirements.md
pierreca: Ohh I see. well yes the idea with these changes is that the retry logic itself will be handled at the client level (which in that case is the caller).

---
In reply to: [139806076](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139806076) [](ancestors = 139806076,139537136,139495305)

common/transport/amqp/src/amqp.ts
pierreca: got it.

---
In reply to: [139807575](https://github.com/Azure/azure-iot-sdk-node/pull/83#discussion_r139807575) [](ancestors = 139807575,139561201,139532979)

PR #85, opened 2017-09-20
Refactor AmqpMethodClient and rename AmqpTwinReceiver for retry logic

pierreca:
# Description of the problem
- `AmqpDeviceMethodClient` was following old conventions that would prevent proper implementation of the retry logic. 
- `AmqpTwinReceiver` handles both sending and receiving.

# Description of the solution
- Update `AmqpDeviceMethodClient` to follow conventions that enable retry logic:
  - Proper errors when called while detached
  - Do not automatically and silently attach links
  - Add attach, detach and forceDetach methods to enable the future caller to control link states.
- Rename `AmqpTwinReceiver` to `AmqpTwinClient`


common/transport/amqp/src/amqp.ts
```diff
@@ -254,12 +254,14 @@ export class Amqp {
               done(err);
             };
 
+            this._amqp.on('client:errorReceived', attachErrorHandler);
```
anthonyvercolano: >this._amqp.on('client:errorReceived', attachErrorHandler); [](start = 12, length = 58)

Are you actually gaining anything here.  Are you going to have to face handling an error twice?  Will the error first come through on the receiver, and then be "resignalled" via the client?

device/transport/amqp/devdoc/amqp_device_method_client_requirements.md
```diff
@@ -44,14 +48,40 @@ client.sendMethodResponse(methodResponse, function(err, result) {
 
 **SRS_NODE_AMQP_DEVICE_METHOD_CLIENT_16_003: [** The `AmqpDeviceMethodClient` shall inherit from the `EventEmitter` class. **]**
 
+### attach(callback)
+
+**SRS_NODE_AMQP_DEVICE_METHOD_CLIENT_16_014: [** The `AmqpDeviceMethodClient` object shall set 2 properties of any AMQP link that it create:
+- `com.microsoft:api-version` shall be set to the current API version in use.
+- `com.microsoft:channel-correlation-id` shall be set to the identifier of the device (also often referred to as `deviceId`). **]**
```
anthonyvercolano: >channel-correlation-id [](start = 17, length = 22)

This must be a service requirement.  Hard to imagine that there isn't some other property of the link that contains the deviceId.  Not pointing out a problem here, it just seems odd.

device/transport/amqp/devdoc/amqp_twin_receiver_requirements.md
```diff
@@ -1,11 +1,11 @@
-# AmqpTwinReceiver Requirements
+# AmqpTwinClient Requirements
 
 ## Overview
 Object used to subscribe to the Cloud-to-Device messages for Twin
 
 ## Example
 ```javascript
-var receiver = new AmqpTwinReceiver(config: ClientConfig, client: any);
+var receiver = new AmqpTwinClient(config: ClientConfig, client: any);
```
anthonyvercolano: >AmqpTwinClient( [](start = 19, length = 15)

in the version string for the sdk blah/x.y.z

What gets updated subsequent to the change? x, y, or z?

device/transport/amqp/src/amqp_device_method_client.ts
```diff
@@ -71,74 +93,118 @@ export class AmqpDeviceMethodClient extends EventEmitter {
...
+            this._fsm.transition('detached');
+          }
+        },
+        detaching: {
+          _onEnter: (forwardedCallback, err) => {
+            /*Codes_SRS_NODE_AMQP_DEVICE_METHOD_CLIENT_16_022: [The `detach` method shall detach both Sender and Receiver links.]*/
+            const links = [this._senderLink, this._receiverLink];
+            async.each(links, (link, callback) => {
+              if (link) {
+                link.detach(callback);
```
anthonyvercolano: >link.detach(callback); [](start = 16, length = 22)

Should you be calling the common forceDetach here?

device/transport/amqp/src/amqp_device_method_client.ts
```diff
@@ -71,74 +93,118 @@ export class AmqpDeviceMethodClient extends EventEmitter {
...
+          detach: (callback) => this._fsm.transition('detaching', callback),
+          forceDetach: () => {
+            /*Codes_SRS_NODE_AMQP_DEVICE_METHOD_CLIENT_16_024: [The `forceDetach` method shall forcefully detach all links.]*/
+            this._senderLink.forceDetach();
+            this._receiverLink.forceDetach();
+            this._fsm.transition('detached');
+          }
+        },
+        detaching: {
+          _onEnter: (forwardedCallback, err) => {
```
anthonyvercolano: >forwardedCallback [](start = 21, length = 17)

I'm confused at this point.  The entry handler is getting this forwardedCallback but I don't see how/what the async.each function passes a callback.

device/transport/amqp/devdoc/amqp_twin_receiver_requirements.md
pierreca: This type is not exported outside of the module. doesn't make any difference for any dependency

---
In reply to: [140019410](https://github.com/Azure/azure-iot-sdk-node/pull/85#discussion_r140019410) [](ancestors = 140019410)

device/transport/amqp/src/amqp_device_method_client.ts
pierreca: [https://caolan.github.io/async/docs.html#each](https://caolan.github.io/async/docs.html#each)
each function is executed, for each of those functions the local callback (named callback) is used, then once all functions have called their local callback, the transition is called with the forwarded callback.

---
In reply to: [140033506](https://github.com/Azure/azure-iot-sdk-node/pull/85#discussion_r140033506) [](ancestors = 140033506)

common/transport/amqp/src/amqp.ts
pierreca: Yup, you don't get these errors in the links because they aren't attached yet. You want the links removed of the dictionary because you're going to hit 'disconnecting' next which is going to try and forceDetach the links, which will screw things up.
There is probably something to be done to simplify error management but amqp10 doesn't make it easy on us for now this passes all the tests :) 

---
In reply to: [140017498](https://github.com/Azure/azure-iot-sdk-node/pull/85#discussion_r140017498) [](ancestors = 140017498)

common/transport/amqp/src/amqp.ts
```diff
@@ -336,7 +340,7 @@ export class Amqp {
               }
 
               if (err) {
-                link.forceDetach(err);
+                link.forceDetach();
                 return callback();
```
BertKleewein: >return  [](start = 16, length = 7)

silly, but this return statement isn't necessary here.

device/transport/amqp/devdoc/amqp_twin_receiver_requirements.md
```diff
@@ -1,11 +1,11 @@
-# AmqpTwinReceiver Requirements
+# AmqpTwinClient Requirements
```
BertKleewein: >AmqpTwinClient [](start = 2, length = 14)

nit: please rename this file when convenient.

PR #86, opened 2017-09-20
requirements for HTTP provisioning object

BertKleewein:
# Description of the problem
Initial SRS for device provisioning HTTP transport module

### DIFF
--- a/provisioning/transport/http/devdoc/http_requirements.md
+++ b/provisioning/transport/http/devdoc/http_requirements.md
@@ -1,0 +1,133 @@
+# azure-iot-provisioning-http Requirements
+
+## Overview
+This module provides HTTP protocol support to communicate with the Azure device provisioning service
+
+```js
+'use strict';
+
+export class Http extends EventEmitter implements DeviceProvisioningTransport {
+  api-version: string;
+  operationStatusPollingInterval: number
+  constructor(idScope: string, registrationId: string) : void;
+  register(authorization: string, forceRegistration: boolean, callback: (err: Error, result?: Any) => void) : void;
+  disconnect(callback: (err) => void);
+}
+```
+
+## Example Usage
+```js
+'use strict';
+var Http = require('azure-iot-provisioning-http').Http;
+
+var http = new Http('__MYSCOPE__', '__MY_DEVICE_ID__');
+
+http.on('operationStatusPollingInterval', (body) => {
+  console.log('operation id ' + body.operationId + ' is currently in state ' + body.state);
+);
+
+http.register(undefined, false, function(err, body)) {
+  if (err) {
+    if (err instanceof 'UnauthorizedError') {
+      console.log('Unauthorized: ' + body.message); // expect to see 'Authorization required, resend request using supplied key'
+    } else {
+  s    console.log('unknown error: ' + err.toString());
+    }
+  } else {
+    console.log('Success: status = ' + body.status);
+  }
+}
+
+
+
+```
+
+## Public Interface
+
+### constructor(idScope: string, registrationId: string) : void
+The `constructor` creates an Http transport object used to communicate with the Azure device provisioning service
+
+**SRS_NODE_PROVISIONING_HTTP_18_001: [** The `Http` constructor shall accept the following properties:
+- `idScope` - a string specifiying the scope of the provisioning operations,
+- `registrationId` - the registration id for the specific device 
+**]**
+
+
+### register(authorization: string, forceRegistration: boolean, callback: (err: Error, result?: Any) => void) : void;
+`register` calls into the service to register the given device with the provisioning service.
+
+**SRS_NODE_PROVISIONING_HTTP_18_002: [** `register` shall throw a `ReferenceError` if `callback` is falsy. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_036: [** `register` shall call the `callback` with an `InvalidOperationError` if it is called while a previous registration is in progress. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_004: [** The registration request shall use Http 1.1. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_005: [** The registration request shall include the current `api-version` as a URL query string value named 'api-version'. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_006: [** The registration request shall specify the following in the Http header:
+  Accept: application/json
+  Content-Type: application/json; charset=utf-8 **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_007: [** If an `authorization` string is specifed, it shall be URL encoded and included in the Http Authorization header. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_008: [** If `forceRegistration` is specified, the registration request shall include this as a query string value named 'forceRegistration' **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_009: [** `register` shall POST the registration request to 'https://global.azure-devices-provisioning.net/{idScope}/registrations/{registrationId}' **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_010: [** `register` shall wait for the response to the POST request. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_011: [** If the registration request times out, `register` shall call the `callback` with a `TimeoutError` error **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_031: [** If `disconnect` is called while the registration request is in progress, `register` shall call the `callback` with an `OperationCancelledError` error.
+ **]**
+**SRS_NODE_PROVISIONING_HTTP_18_012: [** If the registration response contains a body, `register` shall deserialize this into an object. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_013: [** If registration response body fails to deserialize, `register` will throw an `InternalServerError` error. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_014: [** If the registration response has a failed status code, `register` shall use `translateError` to translate this to a common error object and pass this into the `callback` function along with the deserialized body of the response. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_015: [** If the registration response has a success code with a 'status' of 'Assigned', `register` call the `callback` with `err` == `null` and result `containing` the deserialized body **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_016: [** If the registration response has a success code with a 'status' of 'Assigning', `register` shall fire an `operationStatus` event with the deserialized body **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_017: [** If the registration response has a success code with a 'status' of 'Assigning', `register` shall start polling for operation updates **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_029: [** If the registration response has a success code with a 'status' that is not 'Assigned' or  'Assigning', `register` shall call the callback with a `SyntaxError` error. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_018: [** `register` shall poll for operation status every `operationStatusPollingInterval` milliseconds **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_032: [** If `disconnect` is called while the operation status request is in progress, `register` shall call the `callback` with an `OperationCancelledError` error. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_033: [** If `disconnect` is called while the register is waiting between polls, `register` shall call the `callback` with an `OperationCancelledError` error. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_019: [** The operation status request shall use Http 1.1 **]**
+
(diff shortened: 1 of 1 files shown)


provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -0,0 +1,89 @@
...
+### constructor(idScope: string, registrationId: string) : void
+The `constructor` creates an Http transport object used to communicate with the Azure device provisioning service
+
+**SRS_NODE_PROVISIONING_HTTP_18_001: [** The `Http` constructor shall accept the following properties:
+- `idScope` - a string specifiying the scope of the provisioning operations,
+- `registrationId` - the registration id for the specific device 
+**]**
+
+
+### register(authorization?: string, forceRegistration?: boolean, callback (err: Error, result?: Any) => void) : void;
```
pierreca: >callback [](start = 66, length = 8)

missing ':'. also you can't have a mandatory parameter after optional parameters

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -0,0 +1,89 @@
...
+
+**SRS_NODE_PROVISIONING_HTTP_18_004: [** The registration request shall use Http 1.1. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_005: [** The registration request shall include the current `api-version` as a URL query string value. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_006: [** The registration request shall specify the following in the Http header:
+  Accept: application/json
+  Content-Type: application/json; charset=utf-8 **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_007: [** If an `authorization` string is specifed, the registration request shall include this in the Http header. **]**
```
pierreca: >Http [](start = 134, length = 4)

maybe add "Http Authorization header" and the expected format?

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -0,0 +1,89 @@
...
+
+**SRS_NODE_PROVISIONING_HTTP_18_008: [** If `forceRegistration` is specified, the registration request shall include this as a query string value. **]**
```
pierreca: >value [](start = 140, length = 5)

what's the query string parameter name?

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -0,0 +1,89 @@
...
+
+**SRS_NODE_PROVISIONING_HTTP_18_009: [** `register` shall shall POST the registration request to 'https://global.azure-devices-provisioning.net/{idScope}/registrations/{registrationId}' **]**
```
pierreca: >shall [](start = 58, length = 5)

repeated 'shall'

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -0,0 +1,89 @@
...
+
+**SRS_NODE_PROVISIONING_HTTP_18_010: [** `register` shall wait for a registration response. **]**
```
pierreca: This one is not super explicit - maybe rephrase saying when the callback should be called? once we're done polling and the operation succeeded, or when the response to the POST request comes in?

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -0,0 +1,89 @@
...
+
+**SRS_NODE_PROVISIONING_HTTP_18_011: [** If the registration request times out, `register` shall call the `callback` with `TimeoutError` error **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_012: [** If the registration response contains a body, `register` shall deserialize this into a JSON object. **]**
```
pierreca: >JSON [](start = 128, length = 4)

Elio would say that JSON is a notation and by definition a string :) "into an object" is simple enough - and maybe needs a type?

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -0,0 +1,89 @@
...
+
+**SRS_NODE_PROVISIONING_HTTP_18_013: [** If registration response body fails to deserialize, `register` will call the `callback` with a `SyntaxError` error. **]**
```
pierreca: >SyntaxError [](start = 137, length = 11)

- does this error already exist in the SDK? 
- shouldn't we throw here? means it's a programmer error somewhere, likely in the SDK? I don't want to give the customer the impression that this is something they are responsible for by giving them the error in a callback when in fact we are.

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -0,0 +1,89 @@
...
+
+**SRS_NODE_PROVISIONING_HTTP_18_014: [** If the registration response has a failed status code, `register` shall use `translateError` to translate this to a common error object and pass this into the `callback` function along with the deserialized body of the response. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_015: [** If the registration response has a success code with a 'status' of 'Assigned', `register` call the `callback` with `err` == `null` and result `containing` the deserialized body **]**
```
pierreca: >`err` == `null` [](start = 156, length = 15)

I've learned recently that it's recommended to avoid using null when undefined can do. I'm torn between starting to use undefined and make this inconsistent with the rest of the SDK or keeping consistency (and to a certain extent, a bad practice) alive. not sure what to suggest but letting you know what's going on.

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -0,0 +1,89 @@
...
+
+**SRS_NODE_PROVISIONING_HTTP_18_016: [** If the registration response has a success code with a 'status' of 'Assigning', `register` shall fire an `operationStatus` event with the deserialized body **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_017: [** If the registration response has a success code with a 'status' of 'Assigning', `register` shall start polling for operation updates **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_029: [** If the registration response has a success code with a 'status' that is not 'Assigned' or  'Assigning', `register` shall call the callback with a `SyntaxError` error. **]**
```
pierreca: >SyntaxError [](start = 188, length = 11)

same as the other syntax error

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Overview
+This module provides HTTP protocol support to communicate with the Azure device provisioning service
+
+```js
+'use strict';
+
+export class Http extends EventEmitter implements DeviceProvisioningTransport {
+  api-version: string;
+  operationPollingIntervalInSeconds: Number
```
pierreca: >operationPollingIntervalInSeconds [](start = 2, length = 33)

intervals are usually in milliseconds in Javascript, not sure what to do here.
Also, 'operation' is not super descriptive of what we're actually polling for.
Last but not least, what are other SDKs doing for APIs to set this?

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -0,0 +1,89 @@
...
+
+**SRS_NODE_PROVISIONING_HTTP_18_018: [** `register` shall poll for operation status every `operationPollintervalInSeconds` seconds **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_019: [** The operation polling request shall use Http 1.1 **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_020: [** The operation polling request shall have the following in the Http header:
+  Accept: application/json
+  Content-Type: application/json; charset=utf-8 **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_021: [** If an `authorization` string is specifed, the operation polling request include it in the Http header. **]**
```
pierreca: >header [](start = 136, length = 6)

same: which header + expected format

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -0,0 +1,89 @@
...
+
+**SRS_NODE_PROVISIONING_HTTP_18_022: [** operation polling shall be a GET operation sent to 'https://global.azure-devices-provisioning.net/{idScope}/registrations/{registrationId}/operations/{operationId}' **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_023: [** If the operation polling request times out, `register` shall stop polling and call the `callback` with a `TimeoutError` error. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_024: [** `register` shall deserialize the body of the operation polling respone into a JSON object. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_025: [** If the body of the operation polling respone fails to deserialize, `register` will call the `callback` with a `SyntaxError` error. **]**
```
pierreca: >SyntaxError [](start = 152, length = 11)

same as other SyntaxError comments

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -0,0 +1,89 @@
+# azure-iot-provisioning-http Requirements
+
+## Overview
+This module provides HTTP protocol support to communicate with the Azure device provisioning service
+
+```js
+'use strict';
+
+export class Http extends EventEmitter implements DeviceProvisioningTransport {
```
pierreca: >DeviceProvisioningTransport [](start = 50, length = 27)

would be interesting to see what this is.

BertKleewein: I will actually flush this out when I have 2 protocols.  Right now, I'm assuming it will be register+disconnect+maybe some of the strings like api-version.

---
In reply to: [140063985](https://github.com/Azure/azure-iot-sdk-node/pull/86#discussion_r140063985) [](ancestors = 140063985)

provisioning/transport/http/devdoc/http_requirements.md
BertKleewein: I will compare to other SDKs shortly

---
In reply to: [140062992](https://github.com/Azure/azure-iot-sdk-node/pull/86#discussion_r140062992) [](ancestors = 140062992)

provisioning/transport/http/devdoc/http_requirements.md
BertKleewein: Good point.  I don't want to tie this to a type yet.  Even though the schema appears well defined at this point, I don't have a reason  to enforce that schema yet.

---
In reply to: [140060986](https://github.com/Azure/azure-iot-sdk-node/pull/86#discussion_r140060986) [](ancestors = 140060986)

provisioning/transport/http/devdoc/http_requirements.md
pierreca: sounds reasonable

---
In reply to: [140100915](https://github.com/Azure/azure-iot-sdk-node/pull/86#discussion_r140100915) [](ancestors = 140100915,140063985)

provisioning/transport/http/devdoc/http_requirements.md
pierreca: I suspect there will be a setOptions thing somewhere :-/

---
In reply to: [140107581](https://github.com/Azure/azure-iot-sdk-node/pull/86#discussion_r140107581) [](ancestors = 140107581,140062992)

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -0,0 +1,129 @@
...
+  Accept: application/json
+  Content-Type: application/json; charset=utf-8 **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_007: [** If an `authorization` string is specifed, it shall be URL encoded and included in the Http Authorization header. **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_008: [** If `forceRegistration` is specified, the registration request shall include this as a query string value named 'forceRegistration' **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_009: [** `register` shall POST the registration request to 'https://global.azure-devices-provisioning.net/{idScope}/registrations/{registrationId}' **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_010: [** `register` shall wait for the response to the POST request. **]**
```
pierreca: >[** `register` shall wait for the response to the POST request. **]** [](start = 37, length = 69)

so the idea is that register calls the callback after a successful POST request and then we use events to notify the user of progress, am I getting this right?  can the example usage reflect that?

provisioning/transport/http/devdoc/http_requirements.md
pierreca: "yet" nevertheless it'll make it easier for users to have autocompletion with at least an interface. even if for now you leave it as simple as a basic generic dictionnary, it will give us a clue not to forget it.

---
In reply to: [140107605](https://github.com/Azure/azure-iot-sdk-node/pull/86#discussion_r140107605) [](ancestors = 140107605,140060986)

provisioning/transport/http/devdoc/http_requirements.md
pierreca: not convinced by InternalServerError either. might be us screwing up somewhere. I'm not against creating a specific error for that but blaming it on the server doesn't seem right. throwing seems right though

---
In reply to: [140061844](https://github.com/Azure/azure-iot-sdk-node/pull/86#discussion_r140061844) [](ancestors = 140061844)

pierreca:
That's a very good starting point although a few things will need to be ironed out as other transports come online :shipit:

PR #87, opened 2017-09-21
Refactor AmqpReceiver into Amqp itself in preparation for the Amqp state machine logic

pierreca:
# Description of the problem
The AmqpReceiver object had turned into a weird beast grouping device method and c2d features and making their state and error handling separate from the main Amqp object, forcing these responsibilities on the caller (the device client).

# Description of the solution
Refactoring the `AmqpReceiver` code into the `Amqp` object while keeping public APIs and tests the same allows us to prepare for the big move to implementing a transport state machine without breaking any APIs. this is a good step towards the AMQP retry logic.


device/transport/amqp/src/amqp.ts
```diff
@@ -67,7 +70,59 @@ export class Amqp extends EventEmitter implements Client.Transport, StableConnec
...
+    };
+
+    const c2dMessageListener = (msg) => {
+      this.emit('message', msg);
+    };
+
+    /*Codes_SRS_NODE_DEVICE_AMQP_RECEIVER_16_008: [The `Amqp` object shall remove the listeners on `message` and `error` events of the underlying `ReceiverLink` when no-one is listening to its own `message` event.]*/
+    this.on('removeListener', (eventName, eventCallback) => {
+      if (eventName === 'message') {
+        if (this._c2dLink) {
```
anthonyvercolano: >this._c2dLink) [](start = 12, length = 14)

Do you need to null this on the remove?

device/transport/amqp/src/amqp.ts
```diff
@@ -295,6 +343,20 @@ export class Amqp extends EventEmitter implements Client.Transport, StableConnec
 
   /**
    * @private
+   */
+  /*Codes_SRS_NODE_DEVICE_AMQP_RECEIVER_16_007: [The `onDeviceMethod` method shall forward the `methodName` and `methodCallback` arguments to the underlying `Amqp[DeviceMethodClient` object.]*/
```
anthonyvercolano: >[ [](start = 162, length = 1)

? perhaps some other character.

device/transport/amqp/test/_amqp_receiver_test.js
```diff
@@ -41,153 +27,168 @@ describe('AmqpReceiver', function () {
...
+  describe('#on(<others>, callback)', function () {
+    it('does not forward event listeners other than message and errorReceived of the message event to the underlying AMQP receiver', function () {
+      var recv = new AmqpReceiver(fakeConfig);
+      var fakeReceiverLink = new EventEmitter();
+      sinon.spy(fakeReceiverLink, 'on');
+      recv._amqp = new FakeAmqp();
+      sinon.stub(recv._amqp, 'attachReceiverLink').callsArgWith(2, null, fakeReceiverLink);
+      recv.on('foo', function () {});
+      assert(fakeReceiverLink.on.notCalled);
+      assert(recv._amqp.attachReceiverLink.notCalled);
```
anthonyvercolano: > assert(recv._amqp.attachReceiverLink.notCalled) [](start = 5, length = 48)

You've commented in the past about exposing the underlying implementation.  This (and a few other places) doesn't do that?

device/transport/amqp/src/amqp.ts
pierreca: probably a good idea...

---
In reply to: [140297588](https://github.com/Azure/azure-iot-sdk-node/pull/87#discussion_r140297588) [](ancestors = 140297588)

device/transport/amqp/test/_amqp_receiver_test.js
pierreca: yes and no. technically this is asserting something on the low boundary of the blackbox: making sure we're calling the right call. but at the same time we should be able to inject _amqp instead of reaching inside the private variable. so ultimately, you're right, but it's for a good purpose. we should refactor this nevertheless. not now though.

---
In reply to: [140305164](https://github.com/Azure/azure-iot-sdk-node/pull/87#discussion_r140305164) [](ancestors = 140305164)

PR #88, opened 2017-09-22
Bertk move restapiclient

BertKleewein:
Moving RestApiClient from Service SDK into http base

### DIFF
--- a/service/package.json
+++ b/service/package.json
@@ -11,8 +11,7 @@
     "azure-iot-common": "1.1.12",
     "azure-iot-http-base": "1.1.12",
     "debug": "^2.6.3",
-    "lodash": "^4.15.0",
-    "uuid": "^2.0.1"
+    "lodash": "^4.15.0"
   },
   "devDependencies": {
     "chai": "^3.5.0",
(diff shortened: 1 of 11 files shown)


BertKleewein:
This change has successfully gone through a buddy build on Jenkins.

pierreca:
>     httpHeaders['User-Agent'] = packageJson.name + '/' + packageJson.version;

That is going to break telemetry. for now I'd suggest to lock it to 'azure-iothub/' + <packageJson.version but we need to modify the API to take this as an option (probably an option of the constructor?

---
Refers to: common/transport/http/src/rest_api_client.ts:88 in 0a2b7eb. [](commit_id = 0a2b7eb26ebcffcd3d3b9c302d432fa690cf8b44, deletion_comment = False)

service/package.json
```diff
@@ -24,7 +23,8 @@
     "tslint": "^5.1.0",
     "typescript": "2.2.2",
     "@types/node": "^7.0.5",
-    "@types/debug": "0.0.29"
+    "@types/debug": "0.0.29",
+	"uuid": "^2.0.1"
```
pierreca: >"uuid": "^2.0.1" [](start = 1, length = 16)

needs to stay in the dependencies (required by the TS file)

common/core/package.json
```diff
@@ -16,7 +16,8 @@
     "mocha": "^3.0.1",
     "sinon": "^3.2.1",
     "tslint": "^4.5.1",
-    "typescript": "2.0.6"
+    "typescript": "2.0.6",
+    "uuid": "^2.0.1"
```
pierreca: >"uuid": "^2.0.1" [](start = 4, length = 16)

why?

BertKleewein: mistake

---
In reply to: [140601320](https://github.com/Azure/azure-iot-sdk-node/pull/88#discussion_r140601320) [](ancestors = 140601320)

PR #89, opened 2017-09-22
Main AMQP State machine in preparation for the device client retry logic

pierreca:
# Description of the problem
Right now the transports are dumb and the device client is in charge of managing the connection. Since every transport has different idiosyncracies this makes it very complicated to implement a common retry logic. 

# Description of the solution
The logical step is to bake the connection/link state management into the transport itself and then make the device client only handle retries on operations, without caring for what is required of that operation (links, etc) and defer this responsibility to the transport. This PR is a huge step forward in that direction for AMQP.

### DIFF
--- a/device/transport/amqp/devdoc/device_amqp_requirements.md
+++ b/device/transport/amqp/devdoc/device_amqp_requirements.md
@@ -138,7 +138,7 @@ This method is deprecated. The `AmqpReceiver` object and pattern is going away a
 
 **SRS_NODE_DEVICE_AMQP_06_010: [** If the AMQP connection is established, the `updateSharedAccessSignature` method shall call the amqp transport `putToken` method with the first parameter `audience`, created from the `sr` of the shared access signature, the actual shared access signature, and a callback. **]**
 
-**SRS_NODE_DEVICE_AMQP_06_011: [** The `updateSharedAccessSignature` method shall call the `done` callback with a `null` error object and a `SharedAccessSignatureUpdated` object as a result, indicating the client does NOT need to reestablish the transport connection. **]**
+**SRS_NODE_DEVICE_AMQP_06_011: [** The `updateSharedAccessSignature` method shall call the `done` callback with a `null` error object and a `SharedAccessSignatureUpdated` object as a result, with the `needToReconnect` property set to `false`. **]**
 
 ### sendMethodResponse(methodResponse, callback)
 
@@ -194,7 +194,7 @@ An new Amqp message shall be instantiated.
 
 **SRS_NODE_DEVICE_AMQP_06_033: [** The `getTwinReceiver` method shall throw an `ReferenceError` if done is falsy **]**
 
-**SRS_NODE_DEVICE_AMQP_16_026: [** The `getTwinReceiver` method shall call the `done` callback with a `null` error argument and the `AmqpTwinClient` instance currently in use. **]**
+**SRS_NODE_DEVICE_AMQP_16_026: [** The `getTwinReceiver` method shall call the `done` callback with a `null` error argument and the `AmqpTwinClient` instance created when the `Amqp` object was instantiated. **]**
 
 **SRS_NODE_DEVICE_AMQP_16_027: [** The `getTwinReceiver` method shall connect and authenticate the AMQP connection if necessary. **]**
 
--- a/device/transport/amqp/devdoc/device_amqp_requirements.md
+++ b/device/transport/amqp/devdoc/device_amqp_requirements.md
@@ -3,7 +3,7 @@
 
 ## Overview
 `Amqp` provides transport functionality for devices that want to communicate with an Azure IoT Hub using the AMQP protocol. It provides an additional level of abstraction on top of the common Amqp class (`azure-iot-common.Amqp`) which is not specific to the device or service.
-Based on the configuration parameters given to the constructor, the Amqp object will build the SASL-Plain URL used to communicate with the IoT Hub instance, as well as the sending and receiving endpoints, and will instantiate a base Amqp object to use with these parameters.
+Based on the configuration parameters given to the constructor, the Amqp object will build the URL used to communicate with the IoT Hub instance, as well as the sending and receiving endpoints, and will instantiate a base Amqp object to use with these parameters.
 AmqpWs works exactly the same way and provides AMQP transport over websockets. Requirements are the same.
 
 Note that the `Amqp` class now implements what used to be `AmqpReceiver` and as such some of the requirements have transferred over and have been kept mostly intact.
@@ -64,33 +64,47 @@ amqp.on('errorReceived', function (err) {
 ### connect(done)
 The `connect` method establishes a connection with the Azure IoT Hub instance.
 
-**SRS_NODE_DEVICE_AMQP_16_008: [**The `done` callback method passed in argument shall be called if the connection is established**]**
-**SRS_NODE_DEVICE_AMQP_16_009: [**The `done` callback method passed in argument shall be called with an error object if the connection fails**]**
+**SRS_NODE_DEVICE_AMQP_16_008: [**The `done` callback method passed in argument shall be called if the connection is established and authenticated. **]**
+
+**SRS_NODE_DEVICE_AMQP_16_009: [**The `done` callback method passed in argument shall be called with an error object if the connection or authentication fails. **]**
+
 **SRS_NODE_DEVICE_AMQP_06_005: [** If x509 authentication is NOT being utilized then `initializeCBS` shall be invoked. **]**
+
 **SRS_NODE_DEVICE_AMQP_06_008: [** If `initializeCBS` is not successful then the client will remain disconnected and the callback will be called with an error per SRS_NODE_DEVICE_AMQP_16_009. **]**
-**SRS_NODE_DEVICE_AMQP_06_006: [**If `initializeCBS` is successful, `putToken` shall be invoked with the first parameter audience, created from the sr of the sas signature, the next parameter of the actual sas, and a callback. **]**
+
+**SRS_NODE_DEVICE_AMQP_06_006: [**If `initializeCBS` is successful, `putToken` shall be invoked with the first parameter `audience`, created from the `sr` field of the shared access signature, the actual shared access signature, and a callback. **]**
+
 **SRS_NODE_DEVICE_AMQP_06_009: [** If `putToken` is not successful then the client will remain disconnected and the callback will be called with an error per SRS_NODE_DEVICE_AMQP_16_009. **]**
 
 ### disconnect(done)
 The `disconnect` method terminates the connection with the Azure IoT Hub instance.
 
-**SRS_NODE_DEVICE_AMQP_16_010: [**The `done` callback method passed in argument shall be called when disconnected**]**
-**SRS_NODE_DEVICE_AMQP_16_011: [**The `done` callback method passed in argument shall be called with an error object if disconnecting fails**]**
+**SRS_NODE_DEVICE_AMQP_16_010: [**The `done` callback method passed in argument shall be called when disconnected. **]**
+
+**SRS_NODE_DEVICE_AMQP_16_011: [**The `done` callback method passed in argument shall be called with an error object if disconnecting fails. **]**
+
+**SRS_NODE_DEVICE_AMQP_16_022: [** The `disconnect` method shall detach all attached links. **]**
+
+**SRS_NODE_DEVICE_AMQP_16_023: [** The `disconnect` method shall forcefully detach all attached links if a connection error is the causing the transport to be disconnected. **]**
 
 ### sendEvent(message, done)
 
 The `sendEvent` method sends an event to the IoT Hub as the device indicated in the constructor argument.
 
+**SRS_NODE_DEVICE_AMQP_16_024: [** The `sendEvent` method shall connect and authenticate the transport if necessary. **]**
+
+**SRS_NODE_DEVICE_AMQP_16_025: [** The `sendEvent` method shall create and attach the d2c link if necessary. **]**
+
 **SRS_NODE_DEVICE_AMQP_16_002: [**The `sendEvent` method shall construct an AMQP request using the message passed in argument as the body of the message.**]**
+
 **SRS_NODE_DEVICE_AMQP_16_003: [**The `sendEvent` method shall call the `done` callback with a null error object and a MessageEnqueued result object when the message has been successfully sent.**]**
+
 **SRS_NODE_DEVICE_AMQP_16_004: [**If `sendEvent` encounters an error before it can send the request, it shall invoke the `done` callback function and pass the standard JavaScript Error object with a text description of the error (err.message). **]**
 
 ### sendEventBatch(messages, done)
-The `sendEventBatch` method sends a list of events to the IoT Hub as the device indicated in the constructor argument.
+Not implemented
 
-**SRS_NODE_DEVICE_AMQP_16_005: [**If `sendEventBatch` encounters an error before it can send the request, it shall invoke the `done` callback function and pass the standard JavaScript Error object with a text description of the error (err.message).**]**
-
-### getReceiver(done)
+### getReceiver(done) [deprecated]
 This method is deprecated. The `AmqpReceiver` object and pattern is going away and the `Amqp` object now implements the `Receiver` interface until we can completely get rid of it in the device client.
 
 **SRS_NODE_DEVICE_AMQP_16_021: [** The `getReceiver` method shall call the `done` callback with a first argument that is `null` and a second argument that it `this`, ie the current `Amqp` instance. **]**
@@ -122,9 +136,9 @@ This method is deprecated. The `AmqpReceiver` object and pattern is going away a
 
 **SRS_NODE_DEVICE_AMQP_16_015: [**The `updateSharedAccessSignature` method shall save the new shared access signature given as a parameter to its configuration.**]**
 
-**SRS_NODE_DEVICE_AMQP_06_010: [** The `updateSharedAccessSignature` method shall call the amqp transport `putToken` method with the first parameter audience, created from the sr of the sas signature, the next parameter of the actual sas, and a callback. *]**
+**SRS_NODE_DEVICE_AMQP_06_010: [** If the AMQP connection is established, the `updateSharedAccessSignature` method shall call the amqp transport `putToken` method with the first parameter `audience`, created from the `sr` of the shared access signature, the actual shared access signature, and a callback. **]**
 
-**SRS_NODE_DEVICE_AMQP_06_011: [** The `updateSharedAccessSignature` method shall call the `done` callback with a null error object and a SharedAccessSignatureUpdated object as a result, indicating the client does NOT need to reestablish the transport connection. **]**
+**SRS_NODE_DEVICE_AMQP_06_011: [** The `updateSharedAccessSignature` method shall call the `done` callback with a `null` error object and a `SharedAccessSignatureUpdated` object as a result, indicating the client does NOT need to reestablish the transport connection. **]**
 
 ### sendMethodResponse(methodResponse, callback)
 
@@ -135,24 +149,35 @@ This method is deprecated. The `AmqpReceiver` object and pattern is going away a
 ### sendTwinRequest(method, resource, properties, body, done)
 
 **SRS_NODE_DEVICE_AMQP_06_012: [** The `sendTwinRequest` method shall not throw `ReferenceError` if the `done` callback is falsy. **]**
+
 **SRS_NODE_DEVICE_AMQP_06_013: [** The `sendTwinRequest` method shall throw an `ReferenceError` if the `method` argument is falsy. **]**
+
 **SRS_NODE_DEVICE_AMQP_06_014: [** The `sendTwinRequest` method shall throw an `ReferenceError` if the `resource` argument is falsy. **]**
+
 **SRS_NODE_DEVICE_AMQP_06_015: [** The `sendTwinRequest` method shall throw an `ReferenceError` if the `properties` argument is falsy. **]**
+
 **SRS_NODE_DEVICE_AMQP_06_016: [** The `sendTwinRequest` method shall throw an `ReferenceError` if the `body` argument is falsy. **]**
+
 **SRS_NODE_DEVICE_AMQP_06_017: [** The `sendTwinRequest` method shall throw an `ArgumentError` if the `method` argument is not a string. **]**
+
 **SRS_NODE_DEVICE_AMQP_06_018: [** The `sendTwinRequest` method shall throw an `ArgumentError` if the `resource` argument is not a string. **]**
+
 **SRS_NODE_DEVICE_AMQP_06_019: [** The `sendTwinRequest` method shall throw an `ArgumentError` if the `properties` argument is not a an object. **]**
 
 An new Amqp message shall be instantiated.
 
 **SRS_NODE_DEVICE_AMQP_06_020: [** The `method` argument shall be the value of the amqp message `operation` annotation. **]**
+
 **SRS_NODE_DEVICE_AMQP_06_021: [** The `resource` argument shall be the value of the amqp message `resource` annotation. **]**
+
 **SRS_NODE_DEVICE_AMQP_06_031: [** If the `resource` argument terminates in a slash, the slash shall be removed from the annotation. **]**
+
 **SRS_NODE_DEVICE_AMQP_06_039: [** If the `resource` argument length is zero (after terminating slash removal), the resource annotation shall not be set. **]**
 
 **SRS_NODE_DEVICE_AMQP_06_028: [** The `sendTwinRequest` method shall throw an `ArgumentError` if any members of the `properties` object fails to serialize to a string. **]**
 
 **SRS_NODE_DEVICE_AMQP_06_022: [** All properties, except $rid, shall be set as the part of the properties map of the amqp message. **]**
+
 **SRS_NODE_DEVICE_AMQP_06_023: [** The $rid property shall be set as the `correlationId` in the properties map of the amqp message. **]**
 
 **SRS_NODE_DEVICE_AMQP_06_024: [** The `body` shall be value of the body of the amqp message. **]**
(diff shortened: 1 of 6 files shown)


device/transport/amqp/devdoc/device_amqp_requirements.md
```diff
@@ -122,9 +136,9 @@ This method is deprecated. The `AmqpReceiver` object and pattern is going away a
...
 
-**SRS_NODE_DEVICE_AMQP_06_010: [** The `updateSharedAccessSignature` method shall call the amqp transport `putToken` method with the first parameter audience, created from the sr of the sas signature, the next parameter of the actual sas, and a callback. *]**
+**SRS_NODE_DEVICE_AMQP_06_010: [** If the AMQP connection is established, the `updateSharedAccessSignature` method shall call the amqp transport `putToken` method with the first parameter `audience`, created from the `sr` of the shared access signature, the actual shared access signature, and a callback. **]**
 
-**SRS_NODE_DEVICE_AMQP_06_011: [** The `updateSharedAccessSignature` method shall call the `done` callback with a null error object and a SharedAccessSignatureUpdated object as a result, indicating the client does NOT need to reestablish the transport connection. **]**
+**SRS_NODE_DEVICE_AMQP_06_011: [** The `updateSharedAccessSignature` method shall call the `done` callback with a `null` error object and a `SharedAccessSignatureUpdated` object as a result, indicating the client does NOT need to reestablish the transport connection. **]**
```
BertKleewein: >indicating the client does NOT need to reestablish the transport connection. [](start = 191, length = 76)

I have a few questions/issues with this:
1) What is it specifically that indicates that the client doesn't need to re-establish the transport connection?  It sounds like there is an inference here that could be made explicit (e.g. add a reconectionNeeded parameter)
2) There's some religion that I'm missing here.  Why does the client need to re-establish the connection?  If it's because the device transport needs to re-establish links after reconnecting, why can't it happen automatically at this level?

device/transport/amqp/devdoc/device_amqp_requirements.md
```diff
@@ -135,48 +149,68 @@ This method is deprecated. The `AmqpReceiver` object and pattern is going away a
...
-**SRS_NODE_DEVICE_AMQP_06_034: [** If a twin receiver for this endpoint doesn't exist, the `getTwinReceiver` method should create a new `AmqpTwinClient` object. **]**
-**SRS_NODE_DEVICE_AMQP_06_035: [** If a twin receiver for this endpoint has already been created, the `getTwinReceiver` method should not create a new `AmqpTwinClient` object. **]**
-**SRS_NODE_DEVICE_AMQP_06_036: [** The `getTwinReceiver` method shall call the `done` method after it complete. **]**
-**SRS_NODE_DEVICE_AMQP_06_037: [** If a twin receiver for this endpoint did not previously exist, the `getTwinReceiver` method should return the a new `AmqpTwinClient` object as the second parameter of the `done` function with null as the first parameter. **]**
-**SRS_NODE_DEVICE_AMQP_06_038: [** If a twin receiver for this endpoint previously existed, the `getTwinReceiver` method should return the preexisting `AmqpTwinClient` object as the second parameter of the `done` function with null as the first parameter. **]**
+
+**SRS_NODE_DEVICE_AMQP_16_026: [** The `getTwinReceiver` method shall call the `done` callback with a `null` error argument and the `AmqpTwinClient` instance currently in use. **]**
```
BertKleewein: >and the `AmqpTwinClient` instance currently in use. [](start = 124, length = 51)

awkward wording.  'currently in use' implies that it's actively being used for something, but I don't think that's the case.

device/transport/amqp/src/amqp.ts
```diff
@@ -74,54 +83,331 @@ export class Amqp extends EventEmitter implements Client.Transport, StableConnec
...
+                    if (detachErr) {
+                      debug('error detaching the D2C link: ' + detachErr.toString());
+                    } else {
+                      debug('D2C link detached');
+                    }
+                    tmpD2CLink.removeListener('error', this._d2cErrorListener);
+                    if (!finalError && detachErr) {
+                      finalError = translateError('error while detaching the D2C link when disconnecting', detachErr);
+                    }
+                    callback(finalError);
```
anthonyvercolano: >callback(finalError); [](start = 20, length = 21)

Does this abort the series?  If so, what's the point of the async.series?

anthonyvercolano:
>    * @description     This methods sets the SAS token used to authenticate with the IoT Hub service.

The comment here  could be better.  It also performs the re-authorization if the device is connected.  It perhaps should also say it does NOT extend the authentication time of the sas token, and that the extension of the sas token is done by the CALLER of this function.

---
Refers to: device/transport/amqp/src/amqp.ts:522 in 0e15e36. [](commit_id = 0e15e36f14da1a7c061641939a7771ccb01441df, deletion_comment = False)

device/transport/amqp/test/_amqp_test.js
```diff
@@ -4,246 +4,1345 @@
...
+        var fakeMethodResponse = { status: 200, payload: null, requestId: uuid.v4() };
+        fakeBaseClient.connect = sinon.stub();
+        transport.connect();
+        transport.sendMethodResponse(fakeMethodResponse, function (err) {
+          assert.instanceOf(err, errors.NotConnectedError);
+          testCallback();
+        });
+      });
+
+      /*Tests_SRS_NODE_DEVICE_AMQP_16_020: [The `sendMethodResponse` response shall call the `AmqpDeviceMethodClient.sendMethodResponse` method with the arguments that were given to it.]*/
```
anthonyvercolano: >method with the arguments that were given to it [](start = 137, length = 47)

So what's the rational that this fails if disconnected but that onDeviceMethod will initiate a connect.  (At least that's how the requirments read).

device/transport/amqp/test/_amqp_test.js
```diff
@@ -4,246 +4,1345 @@
...
+        sinon.spy(transport._deviceMethodClient, 'sendMethodResponse');
+
+        transport.connect(function () {
+          transport.sendMethodResponse(fakeMethodResponse, fakeCallback);
+          assert.isTrue(transport._deviceMethodClient.sendMethodResponse.calledWith(fakeMethodResponse));
+          testCallback();
+        });
+      });
+
+      it('fails if called disconnecting', function (testCallback) {
```
anthonyvercolano: >fails if called disconnecting' [](start = 10, length = 30)

So what's the rational that this fails if disconnected but that onDeviceMethod will initiate a connect.  (At least that's how the requirments read).

device/transport/amqp/src/amqp.ts
```diff
@@ -74,54 +83,331 @@ export class Amqp extends EventEmitter implements Client.Transport, StableConnec
...
-            this._c2dLink = link;
-            this._c2dLink.on('error', c2dErrorListener);
-            this._c2dLink.on('message', c2dMessageListener);
+        if (this.listeners('message').length === 0) {
+          debug('automatically attaching the link after the first message listener was registered');
+          this._fsm.handle('attachC2DLink', (err) => {
+            debug('C2D link attached automatically');
+            if (err) {
+              debug('error attaching the C2D link automatically: ' + err.toString());
+              this.emit('errorReceived', err);
```
BertKleewein: >this.emit('errorReceived', err); [](start = 14, length = 32)

thinking out loud, it would be nice if this could emit something that could be acted upon.  'errorReceived' is generic and the response is unknown.  'linkConnectionFailure' might be easier to act upon (disconnect then reconnect or something like that).

device/transport/amqp/test/_amqp_test.js
```diff
@@ -4,246 +4,1345 @@
...
+          fakeBaseClient.connect = sinon.stub().callsFake(function (uri, options, done) {
+            connectCallback = done;
+          });
+
+          transport.connect(function (err) {
+            assert.strictEqual(err.amqpError, connectErr);
+            fakeBaseClient.connect = sinon.stub().callsArgWith(2, null, new results.Connected());
+          });
+
+          // now blocked in "connecting" state
```
anthonyvercolano: >// now blocked in "connecting" state [](start = 10, length = 36)

And you know this how?  And it's testable how?

device/transport/amqp/src/amqp.ts
```diff
@@ -74,54 +83,331 @@ export class Amqp extends EventEmitter implements Client.Transport, StableConnec
...
+                callback(err);
+              } else {
+                callback(null, result);
+              }
+            }
+          },
+          connect: (connectCallback) => this._fsm.transition('connecting', connectCallback),
+          disconnect: (disconnectCallback) => {
+            if (disconnectCallback) {
+              disconnectCallback(null, new results.Disconnected());
```
BertKleewein: >results.Disconnected() [](start = 43, length = 22)

Does this mean "I was already disconnected" or "I am now disconnected".  If the second, why doesn't _onEnter also pass this?

device/transport/amqp/src/amqp.ts
```diff
@@ -74,54 +83,331 @@ export class Amqp extends EventEmitter implements Client.Transport, StableConnec
...
+            /*Codes_SRS_NODE_DEVICE_AMQP_16_024: [The `sendEvent` method shall connect and authenticate the transport if necessary.]*/
+            this._fsm.handle('connect', (err, result) => {
+              if (err) {
+                sendCallback(err);
+              } else {
+                this._fsm.handle('sendEvent', message, sendCallback);
+              }
+            });
+          },
+          updateSharedAccessSignature: (token, callback) => {
```
BertKleewein: >updateSharedAccessSignature [](start = 10, length = 27)

naming feels wrong.  This isn't an action to update the signature.  This is an action that is taken after the signature was updated.  Maybe the generic handleSignatureUpdate?

device/transport/amqp/src/amqp.ts
```diff
@@ -74,54 +83,331 @@ export class Amqp extends EventEmitter implements Client.Transport, StableConnec
...
+            // nothing to do here: the SAS has been updated in the config object.
+            callback(null, new results.SharedAccessSignatureUpdated(false));
```
BertKleewein: >false [](start = 68, length = 5)

I think this is the same issue I commented on in the MD file, but false is meaningless to me.  A well named enumerated type might convey meaning better.

device/transport/amqp/src/amqp.ts
```diff
@@ -74,54 +83,331 @@ export class Amqp extends EventEmitter implements Client.Transport, StableConnec
...
+          },
+          connect: (connectCallback) => connectCallback(null, new results.Connected()),
+          disconnect: (disconnectCallback) => this._fsm.transition('disconnecting', disconnectCallback),
+          sendEvent: (message, sendCallback) => {
+            let amqpMessage = AmqpMessage.fromMessage(message);
+            amqpMessage.properties.to = this._d2cEndpoint;
+
+            /*Codes_SRS_NODE_DEVICE_AMQP_16_025: [The `sendEvent` method shall create and attach the d2c link if necessary.]*/
+            if (!this._d2cLink) {
+              this._amqp.attachSenderLink(this._d2cEndpoint, null, (err, link) => {
```
BertKleewein: >attachSenderLink [](start = 25, length = 16)

attachingSenderLink feels like it should be it's own state.  Because the attachSenderLink function is async, the client could do anything while this is happening.  Consider the client which uses Async to call sendEvent many times in parallel.  Each one of those sendEvent actions could end up inside this code at the same time, and I suspect it will cause problems.  (I wholeheartedly disagree with the assertion that you don't have to worry about concurrency because node is single threaded.)

device/transport/amqp/src/amqp.ts
```diff
@@ -74,54 +83,331 @@ export class Amqp extends EventEmitter implements Client.Transport, StableConnec
...
+            /*Codes_SRS_NODE_DEVICE_AMQP_16_026: [** The `getTwinReceiver` method shall call the `done` callback with a `null` error argument and the `AmqpTwinClient` instance currently in use.]*/
+            callback(null, this._twinClient);
+          },
+          sendTwinRequest: (method, resource, properties, body, sendTwinRequestCallback) => {
+            this._twinClient.sendTwinRequest(method, resource, properties, body, sendTwinRequestCallback);
+          },
+          attachC2DLink: (callback) => {
+            debug('attaching C2D link');
+            /*Codes_SRS_NODE_DEVICE_AMQP_16_030: [The `Amqp` object shall attach the C2D `ReceiverLink` object if necessary to start receiving messages.]*/
+            this._amqp.attachReceiverLink(this._c2dEndpoint, null, (err, receiverLink) => {
```
BertKleewein: >attachReceiverLink [](start = 23, length = 18)

same comment about concurrency.  attachingReceiverLink feels like it should be a state.

device/transport/amqp/src/amqp.ts
```diff
@@ -74,54 +83,331 @@ export class Amqp extends EventEmitter implements Client.Transport, StableConnec
...
+              if (err) {
+                debug('error creating a C2D link: ' + err.toString());
+                handleResult('AMQP Transport: Could not attach link', callback)(err);
+              } else {
+                debug('C2D link created and attached successfully');
+                this._c2dLink = receiverLink;
+                this._c2dLink.on('error', this._c2dErrorListener);
+                this._c2dLink.on('message', this._c2dMessageListener);
```
BertKleewein: I _think_ there's the possibility of attaching the same listener multiple times here, depending on how this._amqp.attachReceiverLink acts.  If that function returns an existing object if it's called twice, then these handlers get added to that object twice.

device/transport/amqp/src/amqp.ts
```diff
@@ -74,54 +83,331 @@ export class Amqp extends EventEmitter implements Client.Transport, StableConnec
...
+                callback();
+              }
+            });
+          },
+          detachC2DLink: (callback) => {
+            this._stopC2DListener(undefined, callback);
```
BertKleewein: >_stopC2DListener [](start = 17, length = 16)

My concurrency thoughts are getting out of hand.  stoppingC2DListener should be a separate state too if we're worried about multiple actions in parallel.  (I'm not saying we should fix this now, but this makes the code less than bulletproof.)

device/transport/amqp/src/amqp.ts
```diff
@@ -74,54 +83,331 @@ export class Amqp extends EventEmitter implements Client.Transport, StableConnec
...
+          },
+          onDeviceMethod: (methodName, methodCallback) => {
+            /*Codes_SRS_NODE_DEVICE_AMQP_16_022: [The `onDeviceMethod` method shall call the `onDeviceMethod` method on the `AmqpDeviceMethodClient` object with the same arguments.]*/
+            this._deviceMethodClient.onDeviceMethod(methodName, methodCallback);
+            this._deviceMethodClient.attach((err) => {
```
BertKleewein: >attach [](start = 37, length = 6)

silly question: why do we need to call attach here?  Shouldn't the deviceMethodClient object call it's own attach function as a side effect of the onDeviceMethod call?

device/transport/amqp/src/amqp.ts
```diff
@@ -74,54 +83,331 @@ export class Amqp extends EventEmitter implements Client.Transport, StableConnec
...
+              if (err) {
+                debug('error while attaching the device method client: ' + err.toString());
+                // no need to emit anything: the general error handler registered in the constructor will do that for us.
+              }
+            });
+          }
+        },
+        disconnecting: {
+          _onEnter: (disconnectCallback, err) => {
+            let finalError = err;
```
BertKleewein: >let finalError = err; [](start = 12, length = 21)

Not sure why you need to capture this here.

BertKleewein: never mind, figured it out.

---
In reply to: [141134345](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141134345) [](ancestors = 141134345)

device/transport/amqp/src/amqp.ts
```diff
@@ -74,54 +83,331 @@ export class Amqp extends EventEmitter implements Client.Transport, StableConnec
...
+                    callback(finalError);
+                  });
+                } else {
+                  callback();
+                }
+              },
+              (callback) => {
+                if (this._c2dLink) {
+                  /*Codes_SRS_NODE_DEVICE_AMQP_16_022: [The `disconnect` method shall detach all attached links.]*/
+                  this._stopC2DListener(err, callback);
```
BertKleewein: >this._stopC2DListener(err, callback); [](start = 18, length = 37)

If I understand async.series, then this will abort all the following actions on failure. right?  This means that a failure stopping the C2D listener will prevent this._amqp._disconnct from being called.  Is that what we want?

device/transport/amqp/test/_amqp_test.js
```diff
@@ -4,246 +4,1345 @@
...
+        fakeBaseClient.putToken = sinon.stub().callsFake(function (audience, token, done) {
+          authCallback = done;
+        });
+
+        transport.connect(function () {});
+
+        transport.sendEvent(new Message('test'), function () {
+          assert(fakeBaseClient.connect.calledOnce);
+          testCallback();
+        });
```
anthonyvercolano: Sometimes these things are just hard to follow.

device/transport/amqp/src/amqp.ts
```diff
@@ -244,7 +475,11 @@ export class Amqp extends EventEmitter implements Client.Transport, StableConnec
    */
   /*Codes_SRS_NODE_DEVICE_AMQP_16_013: [The ‘complete’ method shall call the ‘complete’ method of the C2D `ReceiverLink` object and pass it the message and the callback given as parameters.] */
   complete(message: Message, done?: (err?: Error, result?: results.MessageCompleted) => void): void {
```
BertKleewein: >complete [](start = 2, length = 8)

now that I'm completely out of control on the whole concurrency thing, should 'complete' be a FSM action and completingMessage be it's own state?  I think no, but I also wonder how bulletproof we want this to be.

BertKleewein: same for reject and abandon.

---
In reply to: [141136795](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141136795) [](ancestors = 141136795)

BertKleewein:
>    * @description     This methods sets the SAS token used to authenticate with the IoT Hub service.

Being a big believer in self-documenting code, should this function be called setNewSharedAccessSignature.  This code isn't updating the signature.  It's accepting an updated signature, with the assumption that outside code has already updated the signature.

---
In reply to: [332262149](https://github.com/Azure/azure-iot-sdk-node/pull/89#issuecomment-332262149) [](ancestors = 332262149)

---
Refers to: device/transport/amqp/src/amqp.ts:522 in 0e15e36. [](commit_id = 0e15e36f14da1a7c061641939a7771ccb01441df, deletion_comment = False)

device/transport/amqp/src/amqp.ts
```diff
@@ -388,7 +613,11 @@ export class Amqp extends EventEmitter implements Client.Transport, StableConnec
    * @throws {ArgumentError}  One of the parameters is an incorrect type
    */
   sendTwinRequest(method: string, resource: string, properties: { [key: string]: string }, body: any, done?: (err?: Error, result?: results.MessageEnqueued) => void): void {
-    this._twinClient.sendTwinRequest(method, resource, properties, body, done);
+    this._fsm.handle('sendTwinRequest', method, resource, properties, body, (err) => {
+      if (done) {
+        done(err);
+      }
```
BertKleewein: <bert has an inner scream about consistency.  Why are some callbacks optional and others not?>

BertKleewein:
>   getTwinReceiver(done?: (err?: Error, receiver?: AmqpTwinClient) => void): void {

not so optional according to the comment one line down.

---
Refers to: device/transport/amqp/src/amqp.ts:632 in 0e15e36. [](commit_id = 0e15e36f14da1a7c061641939a7771ccb01441df, deletion_comment = False)

device/transport/amqp/test/_amqp_test.js
```diff
@@ -4,246 +4,1345 @@
...
+          // now blocked in "connecting" state
+          transport.connect(function (err, result) {
+            assert.instanceOf(result, results.Connected);
+            assert(fakeBaseClient.connect.calledTwice);
+            testCallback();
+          });
+
+          disconnectCallback();
+        });
+      });
```
BertKleewein: General comment about this whole file, which I'm not going to review in detail:  All this yellow is awesome!  These kinds of tests make me so happy.  I am worried that we have all these tests which are verifying functionality that isn't captured in our requirements doc.  Yes, this has some internal knowledge (caller doesn't know what 'defer' means in this context and shouldn't care), but we could write reqs like
If connect is called while the transport is being disconnect, connect shall wait until the disconnection is complete and then re-connect.
or something like that.  

Not something we should do now, but I'm wondering if this practice would be useful in the future.

device/transport/amqp/devdoc/device_amqp_requirements.md
pierreca: historically the client would take care of reconnecting the transport so the transport needed to communicate the need to reconnect and reestablish links. now that we're pushing this logic down in the transport itself, the client itself doesn't need to take care of reconnecting anything.
I've updated the requirement to be explicit about the needToReconnect value.

---
In reply to: [140955441](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r140955441) [](ancestors = 140955441)

device/transport/amqp/devdoc/device_amqp_requirements.md
pierreca: The AmqpTwinClient is instantiated in the constructor. rephrasing the requirement.

---
In reply to: [140955784](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r140955784) [](ancestors = 140955784)

device/transport/amqp/src/amqp.ts
pierreca: correct. my bad, shouldn't call its callback with the finalError.

---
In reply to: [141106618](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141106618) [](ancestors = 141106618)

device/transport/amqp/src/amqp.ts
pierreca: nope. bug. fixed.

---
In reply to: [141135707](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141135707) [](ancestors = 141135707)

pierreca:
>    * @description     This methods sets the SAS token used to authenticate with the IoT Hub service.

added better comment. agree with that Bert's naming proposal is better but that'd be a breaking change and we're not accepting those (for now). keeping the name but keeping the idea in the back of my head.

---
In reply to: [332285433](https://github.com/Azure/azure-iot-sdk-node/pull/89#issuecomment-332285433) [](ancestors = 332285433,332262149)

---
Refers to: device/transport/amqp/src/amqp.ts:522 in 0e15e36. [](commit_id = 0e15e36f14da1a7c061641939a7771ccb01441df, deletion_comment = False)

device/transport/amqp/test/_amqp_test.js
pierreca: sending a method response requires using the same links that were used to receive the method request. so if we're disconnecting, the response will fail.
However, any call to onDeviceMethod while disconnected is OK because we're not responding, we're establishing new links.

---
In reply to: [141121863](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141121863) [](ancestors = 141121863)

device/transport/amqp/src/amqp.ts
pierreca: 2 things:
1/ i want to change this to the actual 'error' event. the trick is that any uncaught error is thrown so we have to make sure our error handling in the client is solid. this will happen later.
2/ this could be up for reconsideration (although it'd likely be a breaking change) but right now we're using the error type (and not the event name) to detect the type of failure. so the caller still knows what's going on, we're not throwing a generic thing that cannot be acted on.

---
In reply to: [141125512](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141125512) [](ancestors = 141125512)

device/transport/amqp/src/amqp.ts
pierreca: you are correct that _onEnter should use results.Disconnected().

---
In reply to: [141127383](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141127383) [](ancestors = 141127383)

device/transport/amqp/src/amqp.ts
pierreca: this is true, but we need to keep consistency for now. sadly wontfix.

---
In reply to: [141128283](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141128283) [](ancestors = 141128283)

device/transport/amqp/src/amqp.ts
pierreca: also true, also breaking. once we update how the client reacts to these, we can change that.

---
In reply to: [141128520](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141128520) [](ancestors = 141128520)

device/transport/amqp/test/_amqp_test.js
pierreca: because I stubbed the connect method on line 443.  I'll update to a better comment. 

---
In reply to: [141126023](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141126023) [](ancestors = 141126023)

device/transport/amqp/test/_amqp_test.js
pierreca: feel free to suggest a better pattern. adding a comment in the meantime.

---
In reply to: [141136711](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141136711) [](ancestors = 141136711)

device/transport/amqp/src/amqp.ts
pierreca: yeah it's been a nagging pain for me too. we initially had very thin layers where almost all callbacks were optional. Since most of the API calls are internal i'm trying to push this only on the outer layer (in the Client) and make all other callbacks (internals) mandatory to make sure we explicitely handle all error cases. but for the sake of not breaking APIs, well.. sometimes I have to keep those.

I hope i'll be able to do a scrub of all this once we update the client.

---
In reply to: [141137658](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141137658) [](ancestors = 141137658)

device/transport/amqp/src/amqp.ts
pierreca: that's why we have state machines in the SenderLink and ReceiverLink objects :) the link state machine will be created synchronously, cached and immediately switched to 'attaching'. then all subsequent calls will be directed to that object and handled by it. we should be good here.

---
In reply to: [141131461](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141131461) [](ancestors = 141131461)

device/transport/amqp/src/amqp.ts
pierreca: same answer.

---
In reply to: [141131946](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141131946) [](ancestors = 141131946)

device/transport/amqp/src/amqp.ts
pierreca: these are actions that are handled by the c2dlink state machine. 

---
In reply to: [141136795](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141136795) [](ancestors = 141136795)

device/transport/amqp/src/amqp.ts
pierreca: we could. The reason I don't want to is because onDeviceMethod doesn't have a way to surface errors if attaching the link automatically fails (except by using the error event which is a pain to handle). 
Separating the attach calls makes this (and therefore the retry logic) way easier.

---
In reply to: [141133862](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141133862) [](ancestors = 141133862)

device/transport/amqp/src/amqp.ts
pierreca: also handled by the C2D link state machine, we should be fine. unless what you're saying is that only one thing can go on on the AMQP connection at the same time and that nothing else at all should happen until we've received the detach frame. in that case, yes, we're in trouble...

---
In reply to: [141133451](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141133451) [](ancestors = 141133451)

device/transport/amqp/src/amqp.ts
pierreca: actually.. you're right, even though it is cached, a new link will be established. I'll investigate.

---
In reply to: [141166984](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141166984) [](ancestors = 141166984,141131461)

pierreca: technically I think it is OK, for 2 reasons:
1/ keeping links unique where necessary is the job of the caller (the various method/twin clients and the AMQP client which maintains the D2C and C2D link and ensure they are unique.
2/ ultimately I want to get rid of the link cache at base layer.

So i'm actually going to by design this until I realize that this is not possible, 

---
In reply to: [141170034](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141170034) [](ancestors = 141170034,141166984,141131461)

device/transport/amqp/src/amqp.ts
pierreca: the uniqueness of this link is left as an exercise for the caller for now. this is going to be refactored a bit once I get to the client logic, but for now I believe this works.

---
In reply to: [141132830](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141132830) [](ancestors = 141132830)

pierreca: the uniqueness of this link is left as an exercise for the caller for now. this is going to be refactored a bit once I get to the client logic, but for now I believe this works.

---
In reply to: [141132830](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141132830) [](ancestors = 141132830)

device/transport/amqp/test/_amqp_test.js
pierreca: I have been wondering how to tackle this. you're right, we need to think about it more. 

---
In reply to: [141139965](https://github.com/Azure/azure-iot-sdk-node/pull/89#discussion_r141139965) [](ancestors = 141139965)

pierreca:
rebased and merged here: https://github.com/Azure/azure-iot-sdk-node/commit/2972b90fda32dbdea846281c2863288c731a1a31

PR #90, opened 2017-09-25
New state machine for MqttBase

pierreca:
# Description of the problem
There is basically no state management for the connection in the MQTT transpor,t it's all left to the Client to manage the connection. This causes numerous issues (separation of concerns, transport implementation details leaking into the device client layer etc).

# Description of the solution
This is the first PR of a series that will introduce state management in the MQTT transport. the first step is to take care of MqttBase. In doing this we also cleaned up a couple of things with the receivers and Mqtt objects which were reaching in to manipulate the underlying MQTT client directly.

### DIFF
--- a/device/transport/mqtt/devdoc/mqtt_base_requirements.md
+++ b/device/transport/mqtt/devdoc/mqtt_base_requirements.md
@@ -33,6 +33,7 @@ base.receive(function (topic, msg) {
 The `Mqtt` constructor receives the configuration parameters to configure the MQTT.JS library to connect to an IoT hub.
 
 **SRS_NODE_COMMON_MQTT_BASE_16_004: [** The `Mqtt` constructor shall instanciate the default MQTT.JS library if no argument is passed to it. **]**
+
 **SRS_NODE_COMMON_MQTT_BASE_16_005: [** The `Mqtt` constructor shall use the object passed as argument instead of the default MQTT.JS library if it's not falsy. **]**
 
 ### MqttBase.connect(config, done)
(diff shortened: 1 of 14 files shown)


device/transport/mqtt/devdoc/mqtt_base_requirements.md
```diff
@@ -45,40 +46,61 @@ The `connect` method establishes a connection with the server using the config o
...
-**SRS_NODE_COMMON_MQTT_BASE_16_009: [** If the message has properties, the property keys and values shall be uri-encoded, then serialized and appended at the end of the topic with the following convention: `<key>=<value>&<key2>=<value2>&<key3>=<value3>(...)`. **]**
+**SRS_NODE_COMMON_MQTT_BASE_16_029: [** The `unsubscribe` method shall call the `callback` argument with no arguments if the operation succeeds. **]**
 
-**SRS_NODE_COMMON_MQTT_BASE_16_010: [** The `publish` method shall use QoS level of 1. **]**
+**SRS_NODE_COMMON_MQTT_BASE_16_030: [** The `unsubscribe` method shall call the `callback` argument with an `Error` if the operation fails. **]**
 
-**SRS_NODE_COMMON_MQTT_BASE_16_011: [** The `publish` method shall serialize the `messageId` property of the message as a key-value pair on the topic with the key `$.mid`. **]**
 
-**SRS_NODE_COMMON_MQTT_BASE_16_012: [** The `publish` method shall serialize the `correlationId` property of the message as a key-value pair on the topic with the key `$.cid`. **]**
+### MqttBase.updateSharedAccessSignature(sharedAccessSignature, callback)
```
anthonyvercolano: >e(sharedAccessSignature, callback) [](start = 39, length = 34)

I think there should be a requirement akin to the amqp: **SRS_NODE_DEVICE_AMQP_06_011: [** The `updateSharedAccessSignature` method shall call the `done` callback with a null error object and a SharedAccessSignatureUpdated object as a result, indicating the client does NOT need to reestablish the transport connection. **]**

device/transport/mqtt/src/mqtt_base.ts
```diff
@@ -3,48 +3,128 @@
...
+            } else {
+              if (err) {
+                this.emit('error', err);
+              }
+            }
+          },
+          connect: (callback) => this._fsm.transition('connecting', callback),
+          disconnect: (callback) => callback(),
+          /*Codes_SRS_NODE_COMMON_MQTT_BASE_16_020: [The `publish` method shall call the callback with a `NotConnectedError` if the connection hasn't been established prior to calling `publish`.]*/
+          publish: (topic, payload, options, callback) => callback(new errors.NotConnectedError()),
```
anthonyvercolano: >ew errors.NotConnectedError() [](start = 68, length = 29)

and this doesn't autoconnect because?

device/transport/mqtt/src/mqtt_base.ts
```diff
@@ -54,167 +134,126 @@ export class MqttBase {
...
-
-      self.client.on('connect', (connack) => {
-        debug('Device is connected');
-        debug('CONNACK: ' + JSON.stringify(connack));
+    this._mqttClient = this.mqttprovider.connect(uri, options);
+    this._mqttClient.on('message', messageCallback);
+    this._mqttClient.on('error', errorCallback);
+    this._mqttClient.on('close', closeCallback);
+    this._mqttClient.on('offline', offlineCallback);
+    this._mqttClient.on('disconnect', disconnectCallback);
```
anthonyvercolano: Shouldn't these be before the this.mqttprovider.connect call two lines above?

anthonyvercolano:
>       //       the error if the MQTT topic subscription fails. Fix this.

TODO!!!!!

---
Refers to: device/transport/mqtt/src/mqtt_receiver.ts:219 in b4c2e51. [](commit_id = b4c2e51518d9160f6a5ff728c85a89e5eca70788, deletion_comment = False)

device/transport/mqtt/src/mqtt_receiver.ts
```diff
@@ -207,15 +188,10 @@ export class MqttReceiver extends EventEmitter implements Receiver {
                         self._topics.message :
                         null;
       if (!!topic) {
+        debug('removing listener for topic: ' + JSON.stringify(topic));
         // decrement listeners count for this topic
         --(topic.listenersCount);
```
anthonyvercolano: >) [](start = 31, length = 1)

Did you REALLY want this in the parens?  Are you decrementing the value as opposed to the object?

pierreca:
>       //       the error if the MQTT topic subscription fails. Fix this.

yep, can't be done with the current interface though - we need changes in the higher layer and in the client logic. so for now it shall remain as is... unfortunately...

---
In reply to: [332027814](https://github.com/Azure/azure-iot-sdk-node/pull/90#issuecomment-332027814) [](ancestors = 332027814)

---
Refers to: device/transport/mqtt/src/mqtt_receiver.ts:219 in b4c2e51. [](commit_id = b4c2e51518d9160f6a5ff728c85a89e5eca70788, deletion_comment = False)

device/transport/mqtt/devdoc/mqtt_base_requirements.md
pierreca: Yes, this will be created as part of the higher-level MQTT transport refactoring. 

---
In reply to: [140887707](https://github.com/Azure/azure-iot-sdk-node/pull/90#discussion_r140887707) [](ancestors = 140887707)

device/transport/mqtt/src/mqtt_base.ts
pierreca: `this._mqttClient` is the returned value of the `connect` call, so we can't subscribe before we have the object. we have to bank on the fact that connect returns synchronously in the underlying mqtt library.

---
In reply to: [140911193](https://github.com/Azure/azure-iot-sdk-node/pull/90#discussion_r140911193) [](ancestors = 140911193)

device/transport/mqtt/src/mqtt_base.ts
pierreca: because the connection information is passed in the `connect` call and we can't guarantee that `connect` has been called before `publish` is called when we're in the `disconnected` state.

---
In reply to: [140908660](https://github.com/Azure/azure-iot-sdk-node/pull/90#discussion_r140908660) [](ancestors = 140908660)

device/transport/mqtt/devdoc/mqtt_base_requirements.md
```diff
@@ -45,40 +46,61 @@ The `connect` method establishes a connection with the server using the config o
...
 
-**SRS_NODE_COMMON_MQTT_BASE_16_013: [** The `publish` method shall serialize the `userId` property of the message as a key-value pair on the topic with the key `$.uid`. **]**
+**SRS_NODE_COMMON_MQTT_BASE_16_032: [** The `updateSharedAccessSignature` method shall throw a `ReferenceError` if the `sharedAccessSignature` argument is falsy. **]**
 
-**SRS_NODE_COMMON_MQTT_BASE_16_014: [** The `publish` method shall serialize the `to` property of the message as a key-value pair on the topic with the key `$.to`. **]**
+**SRS_NODE_COMMON_MQTT_BASE_16_033: [** The `updateSharedAccessSignature` method shall disconnect and reconnect the mqtt client with the new `sharedAccessSignature`. **]**
```
BertKleewein: >disconnect and reconnect [](start = 87, length = 24)

are you sure disconnect->reconnect belongs at this level?  None of the other operations at this level have any connection side effects.

device/transport/mqtt/devdoc/mqtt_requirements.md
```diff
@@ -80,37 +80,53 @@ The `sendEvent` method sends an event to an IoT hub on behalf of the device indi
 
 **SRS_NODE_DEVICE_MQTT_12_005: [** The `sendEvent` method shall call the publish method on `MqttTransport`. **]**
 
+**SRS_NODE_COMMON_MQTT_BASE_16_008: [** The `sendEvent` method shall use a topic formatted using the following convention: `devices/<deviceId>/messages/events/`. **]**
+
+**SRS_NODE_COMMON_MQTT_BASE_16_009: [** If the message has properties, the property keys and values shall be uri-encoded, then serialized and appended at the end of the topic with the following convention: `<key>=<value>&<key2>=<value2>&<key3>=<value3>(...)`. **]**
```
BertKleewein: ><key> [](start = 207, length = 5)

Is there a ? before the first key as in HTTP URIs?

device/transport/mqtt/src/mqtt.ts
```diff
@@ -36,21 +40,36 @@ export class Mqtt extends EventEmitter implements Client.Transport, StableConnec
...
    */
-  constructor(config: ClientConfig, provider?: MqttBase) {
+  constructor(config: ClientConfig, provider?: any) {
     super();
     this._config = config;
+    this._topicTelemetryPublish = 'devices/' + this._config.deviceId + '/messages/events/';
+    this._topicMessageSubscribe = 'devices/' + this._config.deviceId + '/messages/devicebound/#';
+    debug('topic publish: ' + this._topicTelemetryPublish);
+    debug('topic subscribe: ' + this._topicMessageSubscribe);
+    const sdkVersionString = encodeURIComponent('azure-iot-device/' + packageJson.version);
```
BertKleewein: >packageJson.version [](start = 70, length = 19)

This is the version of the mqtt transport, not the version of the sdk.  Is this what we want?

pierreca: Right now they are identical. TBH we probably want a combination of transport and client versions. I think Olivier and Yi have something in the works to improve telemetry so since this is a small change that is kindof out of scope for now, I'd say let's wait and fix this later. the most important part (azure-iot-device) is correct.

---
In reply to: [141152380](https://github.com/Azure/azure-iot-sdk-node/pull/90#discussion_r141152380) [](ancestors = 141152380)

device/transport/mqtt/devdoc/mqtt_requirements.md
pierreca: no apparently it's after a '/' directly that is included in the topic name, specified in SRS_NODE_COMMON_MQTT_BASE_16_008

---
In reply to: [141150761](https://github.com/Azure/azure-iot-sdk-node/pull/90#discussion_r141150761) [](ancestors = 141150761)

device/transport/mqtt/devdoc/mqtt_base_requirements.md
pierreca: I keep going back and forth on this one. elements to think about:
- in AMQP this is taken care of in the putToken call which is also on the base layer
- this also feels like it would be useful in the DPS transport, so it makes sense for it to be shared
- we're effectively replacing the mqtt client object hiding it below the MqttBase class but this is not supposed to trigger anything on the Mqtt object itself: unlike with AMQP there are no links to reestablish. since we set cleanSession to false, all subscribtions will be automatically reenabled.
- the only reason other methods don't have an effect on the connection is because we can't be sure to have a shared access signature when they are called. that case is specific since by definition this is what it's for. otherwise i'd happily handle auto-connect on publish in here too.

if you're thinking of a particular case where this would be detrimental please do let me know.

---
In reply to: [141150498](https://github.com/Azure/azure-iot-sdk-node/pull/90#discussion_r141150498) [](ancestors = 141150498)

pierreca:
rebased and merged here: https://github.com/Azure/azure-iot-sdk-node/commit/08ef708d412c851d0c8848b34d325b7ca1bade08

PR #92, opened 2017-09-27
Mqtt transport main state machine

pierreca:
# Description of the problem
A lot of transport-specific issues are dealt with at the Device Client level, when really the device client shoudn't care.

# Description of the solution
This state machine pushes the management of connection, subscriptions and topics down in the MQTT transport, in preparation for a drastic simplification of the client that will lead the way for retry logic.

### DIFF
--- a/device/transport/mqtt/devdoc/mqtt_requirements.md
+++ b/device/transport/mqtt/devdoc/mqtt_requirements.md
@@ -113,9 +113,11 @@ The `sendEvent` method sends an event to an IoT hub on behalf of the device indi
 
 **SRS_NODE_DEVICE_MQTT_16_024: [** The `sendEvent` method shall call its callback with an `Error` that has been translated using the `translateError` method if the `MqttBase` object fails to establish a connection. **]**
 
-**SRS_NODE_DEVICE_MQTT_16_025: [** The `sendEvent` method shall be deferred until either disconnected or connected if it is called while `MqttBase` is establishing the connection. **]**
+**SRS_NODE_DEVICE_MQTT_16_025: [** If `sendEvent` is called while `MqttBase` is establishing the connection, it shall wait until the connection is established and then send the event. **]**
 
-**SRS_NODE_DEVICE_MQTT_16_026: [** The `sendEvent` method shall be deferred until disconnected if it is called while `MqttBase` is disconnecting. **]**
+**SRS_NODE_DEVICE_MQTT_16_035: [** If `sendEvent` is called while `MqttBase` is establishing the connection, and `MqttBase` fails to establish the connection, then sendEvent shall fail. **]**
+
+**SRS_NODE_DEVICE_MQTT_16_026: [** If `sendEvent` is called while `MqttBase` is disconnecting, it shall wait until the disconnection is complete and then try to connect again and send the event.  **]**
 
 **SRS_NODE_DEVICE_MQTT_16_027: [** The `sendEvent` method shall call its callback with an `Error` that has been translated using the `translateError` method if the `MqttBase` object fails to publish the message. **]**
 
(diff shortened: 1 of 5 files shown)


device/transport/mqtt/devdoc/mqtt_requirements.md
```diff
@@ -96,24 +109,38 @@ The `sendEvent` method sends an event to an IoT hub on behalf of the device indi
 
 **SRS_NODE_COMMON_MQTT_BASE_16_015: [** The `sendEvent` method shall serialize the `expiryTimeUtc` property of the message as a key-value pair on the topic with the key `$.exp`. **]**
 
+**SRS_NODE_DEVICE_MQTT_16_023: [** The `sendEvent` method shall connect the Mqtt connection if it is disconnected. **]**
+
+**SRS_NODE_DEVICE_MQTT_16_024: [** The `sendEvent` method shall call its callback with an `Error` that has been translated using the `translateError` method if the `MqttBase` object fails to establish a connection. **]**
+
+**SRS_NODE_DEVICE_MQTT_16_025: [** The `sendEvent` method shall be deferred until either disconnected or connected if it is called while `MqttBase` is establishing the connection. **]**
```
BertKleewein: >deferred [](start = 67, length = 8)

I think I commented on a different CR, but I don 't like "deferred" here because that talks about internal implementation rather than behavior.  
Maybe:
If sendEvent is called while MqttBase is establishing the connection, it shall want until the connection is established and then send the event.
If sendEvent is called while MqttBase is establishing the connection, and MqttBase fails to establish the connection, then sendEvent shall fail.

device/transport/mqtt/devdoc/mqtt_requirements.md
```diff
@@ -210,7 +240,16 @@ The `sendTwinRequest` method sends the given body to the given endpoint on an Io
 
 **SRS_NODE_DEVICE_MQTT_18_017: [** If the `sendTwinRequest` method is successful, the first parameter to the `done` callback shall be null and the second parameter shall be a MessageEnqueued object. **]**
 
+**SRS_NODE_DEVICE_MQTT_16_029: [** The `sendTwinRequest` method shall connect the Mqtt connection if it is disconnected. **]**
+
+**SRS_NODE_DEVICE_MQTT_16_031: [** The `sendTwinRequest` method shall be deferred until either disconnected or connected if it is called while `MqttBase` is establishing the connection. **]**
+
+**SRS_NODE_DEVICE_MQTT_16_032: [** The `sendTwinRequest` method shall be deferred until disconnected if it is called while `MqttBase` is disconnecting. **]**
```
BertKleewein: same wording comment as 025and 026 above

device/transport/mqtt/src/mqtt.ts
```diff
@@ -70,6 +71,183 @@ export class Mqtt extends EventEmitter implements Client.Transport, StableConnec
...
+                callback(null, new results.MessageEnqueued(puback));
+              }
+            });
+          },
+          updateSharedAccessSignature: (sharedAccessSignature, callback) => {
+            /*Codes_SRS_NODE_DEVICE_MQTT_16_028: [The `updateSharedAccessSignature` method shall call the `updateSharedAccessSignature` method on the `MqttBase` object if it is connected.]*/
+            this._mqtt.updateSharedAccessSignature(sharedAccessSignature, (err) => {
+              if (err) {
+                /*Codes_SRS_NODE_DEVICE_MQTT_16_009: [The `updateSharedAccessSignature` method shall call the `done` method with an `Error` object if `MqttBase.updateSharedAccessSignature` fails.]*/
+                this._fsm.transition('disconnected', callback, err);
```
BertKleewein: >this._fsm.transition('disconnected', callback, err); [](start = 16, length = 52)

This assumes that the base was disconnected.  Do we want to force the base to disconnect here to avoid inconsistent states?

pierreca: errors are emitted only when the base is disconnected so this is assumed. happy to question that pattern though.

---
In reply to: [141443815](https://github.com/Azure/azure-iot-sdk-node/pull/92#discussion_r141443815) [](ancestors = 141443815)

PR #93, opened 2017-09-27
http provisioning object

BertKleewein:
I am not ready to push this yet.  I still need to finish the unit tests and write a simple test app that can roundtrip to a live DPS service.  The actual http.ts code is complete and open for comments.  Feel free to comment on anything and everything, but I'm specifically looking big things I may have missed (like dev-setup.cmd which I didn't miss) or naming conventions or file locations so I can get a head start on anything big.

### DIFF
--- a/build/dev-teardown.cmd
+++ b/build/dev-teardown.cmd
@@ -23,6 +23,12 @@ call npm rm azure-iot-device-amqp
 call npm rm azure-iot-device
 call npm rm azure-iot-common
 
+cd %node-root%\device\provisioning\http
+echo -- Removing links for %cd% --
+call npm rm
+call npm rm azure-iot-http-base
+call npm rm azure-iot-common
+
 cd %node-root%\device\transport\mqtt
 echo -- Removing links for %cd% --
 call npm rm
--- a/build/dev-teardown.sh
+++ b/build/dev-teardown.sh
@@ -18,6 +18,12 @@ npm rm azure-iot-device-amqp
 npm rm azure-iot-device
 npm rm azure-iot-common
 
+cd $node_root/provisioning/transport/http
+echo "-- Removing links for `pwd` --"
+npm rm
+npm rm azure-iot-http-base
+npm rm azure-iot-common
+
 cd $node_root/device/transport/mqtt
 echo "-- Removing links for `pwd` --"
 npm rm
--- a/common/core/src/errors.ts
+++ b/common/core/src/errors.ts
@@ -370,3 +370,18 @@ export class DeviceTimeoutError extends Error {
     Error.captureStackTrace(this, this.constructor);
   }
 }
+
+/**
+ * Error thrown when ant operation (local or remote) is cancelled
+ *
+ * @augments {Error}
+ */
+export class OperationCancelledError extends Error {
+  constructor(message?: string) {
+    super(message);
+    this.name = 'OperationCancelledError';
+    this.message = message;
+    Error.captureStackTrace(this, this.constructor);
+  }
+}
+
--- a/common/transport/http/src/rest_api_client.ts
+++ b/common/transport/http/src/rest_api_client.ts
@@ -3,9 +3,10 @@
 
 'use strict';
 
-import { anHourFromNow, errors, SharedAccessSignature } from 'azure-iot-common';
+import { anHourFromNow, errors, SharedAccessSignature, X509 } from 'azure-iot-common';
 import { Http as HttpBase } from './http';
 import  * as uuid from 'uuid';
+import { ClientRequest } from 'http';
 
 
 /**
@@ -40,11 +41,10 @@ export class RestApiClient {
   constructor(config: RestApiClient.TransportConfig, userAgent: string, httpRequestBuilder?: HttpBase) {
     /*Codes_SRS_NODE_IOTHUB_REST_API_CLIENT_16_001: [The `RestApiClient` constructor shall throw a `ReferenceError` if config is falsy.]*/
     if (!config) throw new ReferenceError('config cannot be \'' + config + '\'');
-    /*Codes_SRS_NODE_IOTHUB_REST_API_CLIENT_16_002: [The `RestApiClient` constructor shall throw an `ArgumentError` if config is missing a `host` or `sharedAccessSignature` property.]*/
+    /*Codes_SRS_NODE_IOTHUB_REST_API_CLIENT_16_002: [The `RestApiClient` constructor shall throw an `ArgumentError` if config is missing a `host` property.]*/
     if (!config.host) throw new errors.ArgumentError('config.host cannot be \'' + config.host + '\'');
-    if (!config.sharedAccessSignature) throw new errors.ArgumentError('config.sharedAccessSignature cannot be \'' + config.sharedAccessSignature + '\'');
     /*Codes_SRS_NODE_IOTHUB_REST_API_CLIENT_18_001: [The `RestApiClient` constructor shall throw a `ReferenceError` if `userAgent` is falsy.]*/
-    if (!userAgent) throw new ReferenceError('userAgent cannot be \'' + config + '\'');
+    if (!userAgent) throw new ReferenceError('userAgent cannot be \'' + userAgent + '\'');
 
     this._config = config;
     this._userAgent = userAgent;
@@ -85,7 +85,9 @@ export class RestApiClient {
     - Request-Id: <guid>
     - User-Agent: <version string>]*/
     let httpHeaders: any = headers || {};
-    httpHeaders.Authorization = (typeof(this._config.sharedAccessSignature) === 'string') ? this._config.sharedAccessSignature as string : (this._config.sharedAccessSignature as SharedAccessSignature).extend(anHourFromNow());
+    if (this._config.sharedAccessSignature) {
+      httpHeaders.Authorization = (typeof(this._config.sharedAccessSignature) === 'string') ? this._config.sharedAccessSignature as string : (this._config.sharedAccessSignature as SharedAccessSignature).extend(anHourFromNow());
+    }
     httpHeaders['Request-Id'] = uuid.v4();
     httpHeaders['User-Agent'] = this._userAgent;
 
@@ -115,8 +117,7 @@ export class RestApiClient {
       headers['Content-Length'] = requestBodyStringSizeInBytes;
     }
 
-    /*Codes_SRS_NODE_IOTHUB_REST_API_CLIENT_16_008: [The `executeApiCall` method shall build the HTTP request using the arguments passed by the caller.]*/
-    const request = this._http.buildRequest(method, path, httpHeaders, this._config.host, (err, responseBody, response) => {
+    let requestCallback: (err: Error, responseBody: any, response: any) => void = (err, responseBody, response) =>  {
       if (err) {
         if (response) {
           /*Codes_SRS_NODE_IOTHUB_REST_API_CLIENT_16_010: [If the HTTP request fails with an error code >= 300 the `executeApiCall` method shall translate the HTTP error into a transport-agnostic error using the `translateError` method and call the `done` callback with the resulting error as the only argument.]*/
@@ -130,7 +131,16 @@ export class RestApiClient {
         const result = responseBody ? JSON.parse(responseBody) : '';
         done(null, result, response);
       }
-    });
+    };
+
+    /*Codes_SRS_NODE_IOTHUB_REST_API_CLIENT_16_008: [The `executeApiCall` method shall build the HTTP request using the arguments passed by the caller.]*/
+    let request: ClientRequest;
+    if (!!this._config.x509) {
+      /* Codes_SRS_NODE_IOTHUB_REST_API_CLIENT_18_002: [ If an `x509` cert was passed into the constructor via the `config` object, `executeApiCall` shall use it to establish the TLS connection. ] */
+       request = this._http.buildRequest(method, path, httpHeaders, this._config.host, this._config.x509, requestCallback);
+    } else {
+       request = this._http.buildRequest(method, path, httpHeaders, this._config.host, requestCallback);
+    }
 
     /*Codes_SRS_NODE_IOTHUB_REST_API_CLIENT_16_030: [If `timeout` is defined and is not a function, the HTTP request timeout shall be adjusted to match the value of the argument.]*/
     if (timeout) {
@@ -250,8 +260,9 @@ export class RestApiClient {
 export namespace RestApiClient {
     export interface TransportConfig {
         host: string;
-        sharedAccessSignature: string | SharedAccessSignature;
+        sharedAccessSignature?: string | SharedAccessSignature;
+        x509?: X509;
     }
 
-    export type ResponseCallback = (err: Error, device?: any, response?: any) => void;
+    export type ResponseCallback = (err: Error, responseBody?: any, response?: any) => void;
 }
--- a/provisioning/transport/http/index.js
+++ b/provisioning/transport/http/index.js
@@ -1,0 +1,18 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+
+var Http = require('./lib/http.js').Http;
+
+/**
+ * The `azure-iot-provisioning-http` module provides support for the HTTPS protocol to the Azure Device Provisoning Service.
+ *
+ * @module azure-iot-provisioning-http
+ * @requires module:azure-iot-http-base
+ * @requires module:azure-iot-common
+ */
+
+module.exports = {
+  Http: Http
+};
--- a/provisioning/transport/http/package.json
+++ b/provisioning/transport/http/package.json
@@ -1,24 +1,26 @@
 {
-  "name": "azure-iot-http-base",
-  "version": "1.1.12",
-  "description": "HTTP operations used by Azure IoT device and service SDKs",
+  "name": "azure-iot-provisioning-http",
+  "version": "1.0.0",
+  "description": "HTTP transport for Azure Device Provisioning SDK",
   "author": "Microsoft Corporation",
   "license": "MIT",
   "main": "index.js",
   "typings": "index.d.ts",
   "dependencies": {
+    "azure-iot-http-base": "1.1.12",
     "azure-iot-common": "1.1.12",
-    "debug": "^2.2.0",
-    "uuid": "^2.0.1"
+    "debug": "^3.0.1",
+    "machina": "^2.0.0"
   },
   "devDependencies": {
     "chai": "^3.5.0",
     "istanbul": "^0.4.4",
     "jshint": "^2.9.2",
     "mocha": "^3.0.1",
+    "sinon": "^1.17.5",
     "tslint": "^5.1.0",
     "typescript": "2.2.2",
-    "@types/node": "^7.0.12"
+    "@types/node": "^7.0.5"
   },
   "scripts": {
     "lint": "tslint --type-check --project . -c ../../../tslint.json",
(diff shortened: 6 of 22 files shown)


provisioning/transport/http/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "azure-iot-provisioning-http",
+  "version": "1.0.0",
+  "description": "HTTP transport for Azure device provisoning SDK",
+  "author": "Microsoft Corporation",
+  "license": "MIT",
+  "main": "index.js",
+  "typings": "index.d.ts",
+  "dependencies": {
```
pierreca: >dependencies [](start = 3, length = 12)

missing azure-iot-common I think

build/dev-teardown.cmd
```diff
@@ -23,6 +23,11 @@ call npm rm azure-iot-device-amqp
 call npm rm azure-iot-device
 call npm rm azure-iot-common
 
+cd %node-root%\device\provisioning\http
+echo -- Removing links for %cd% --
+call npm rm
+call npm rm azure-iot-http-base
```
pierreca: +azure-iot-common

build/dev-teardown.sh
```diff
@@ -18,6 +18,11 @@ npm rm azure-iot-device-amqp
 npm rm azure-iot-device
 npm rm azure-iot-common
 
+cd $node_root/provisioning/transport/http
+echo "-- Removing links for `pwd` --"
+npm rm
```
pierreca: +azure-iot-common

common/core/src/errors.ts
```diff
@@ -370,3 +370,18 @@ export class DeviceTimeoutError extends Error {
     Error.captureStackTrace(this, this.constructor);
   }
 }
+
+/**
+ * Error thrown when an operation is cancelled
```
pierreca: >operation [](start = 24, length = 9)

is it generic or a specific type of DPS operation?

pierreca:
> http.on('operationStatusPollingInterval', (body) => {

operationStatusUpdate? should we emit for every interval or only when the state changes?

---
Refers to: provisioning/transport/http/devdoc/http_requirements.md:25 in f95de3f. [](commit_id = f95de3f2ef2e8397ca3c6254f026ed618b091b8a, deletion_comment = False)

pierreca:
> **SRS_NODE_PROVISIONING_HTTP_18_018: [** `register` shall poll for operation status every `operationStatusPollingInterval` milliseconds **]**

this is used as a constant and an event name, consider if there is an interest in differentiating them  - also do we have a reasonable default to configure (and have a requirement for it)?

---
Refers to: provisioning/transport/http/devdoc/http_requirements.md:95 in f95de3f. [](commit_id = f95de3f2ef2e8397ca3c6254f026ed618b091b8a, deletion_comment = False)

provisioning/transport/http/index.js
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+
+var Http = require('./lib/http.js').Http;
+
+/**
+ * The `azure-iot-device-http` module provides support for the HTTPS protocol to the device [client]{@link module:azure-iot-device.Client}.
```
pierreca: >device [client]{@link module:azure-iot-device.Client} [](start = 85, length = 53)

update the comment with something related to provisioning?

provisioning/transport/http/index.js
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * @module azure-iot-provisioning-http
+ * @requires module:azure-iot-http-base
```
pierreca: >base [](start = 35, length = 4)

and azure-iot-common

provisioning/transport/http/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "azure-iot-provisioning-http",
+  "version": "1.0.0",
+  "description": "HTTP transport for Azure device provisoning SDK",
```
pierreca: >device provisoning [](start = 43, length = 18)

needs to be double checked but if this is an official service name it probably needs to be capitalized.

provisioning/transport/http/readme.md
```diff
@@ -0,0 +1,16 @@
+#azure-iot-provisioning-http
+Communicate with Azure Device Provisioning Service over HTTP
+
+[![npm version](https://badge.fury.io/js/azure-iot-provisioning-http.svg)](https://badge.fury.io/js/azure-iot-provisioning-http)
+
+## Install
+
+`npm install -g azure-iot-provisioning-http@latest` to get the latest (pre-release) version.
```
pierreca: >latest [](start = 44, length = 6)

maybe will be revisited later but not sure if we'll use latest or something else for the tag

pierreca:
> 

shouldnt need that if you npm install @types/node instead

---
Refers to: provisioning/transport/http/typings.json:9 in f95de3f. [](commit_id = f95de3f2ef2e8397ca3c6254f026ed618b091b8a, deletion_comment = False)

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,271 @@
...
+
+  /**
+   * @private
+   * @constructor
+   * @param config The configuration object.
+   */
+  /* Codes_SRS_NODE_PROVISIONING_HTTP_18_001: [ The `Http` constructor shall accept the following properties:
+  - `idScope` - a string specifiying the scope of the provisioning operations,
+  - `registrationId` - the registration id for the specific device ] */
+  constructor(idScope: string, registrationId: string, httpBase?: Base) {
```
pierreca: >httpBase?: Base [](start = 55, length = 15)

feels like this should be RestApiClient and not Base

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,271 @@
...
+        callback(err, null);
+      } else {
+        callback(null, responseBody);
+      }
+    });
+
+    /* TODO: Codes_SRS_NODE_PROVISIONING_HTTP_18_032: [ If `disconnect` is called while the operation status request is in progress, `register` shall call the `callback` with an `OperationCancelledError` error. ] */
+  }
+
+  private _safeCallback(callback: (error2?: Error | null, result2?: any) => void, error?: Error | null, result?: any): void {
```
pierreca: >_safeCallback [](start = 10, length = 13)

do we need this? all callbacks should be mandatory. I would like to push the "optional callback" only on the highest layer of the user-accessible API

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,271 @@
...
+            this._safeCallback(callback);
+          }
+        },
+        registering: {
+          _onEnter: (callback) => {
+            this._sendRegistrationRequest((err, responseBody) => {
+              if (err) {
+                this._safeCallback(callback, err);
+                this._fsm.transition('idle');
+              } else if (this._isAssigning(responseBody)) {
```
pierreca: >this._isAssigning(responseBody)) [](start = 25, length = 32)

not strongly opposed to such a functional approach but makes me wonder if we need a class type for the responseBody

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,271 @@
...
+            this._currentHttpRequest = null;
+            this._restApiClient = null;
+            this._body = null;
+            this._forceRegistration = false;
+            this._operationId = '';
+          },
+          register: (authorization, forceRegistration, body, callback) => {
+            this._forceRegistration = forceRegistration;
+            this._body = body;
+            this._restApiClient = new RestApiClient({ 'host' : this._provisioningServer , 'sharedAccessSignature' : authorization}, 'azure-node-provisioning', httpBase);
```
pierreca: >azure-node-provisioning [](start = 133, length = 23)

need to land this with PMs and telemetry folks asap - talk with Yi?

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,271 @@
...
+          disconnect : (callback) => {
+            // nothing to do.
+            this._safeCallback(callback);
+          }
+        },
+        registering: {
+          _onEnter: (callback) => {
+            this._sendRegistrationRequest((err, responseBody) => {
+              if (err) {
+                this._safeCallback(callback, err);
```
pierreca: >this._safeCallback(callback, err); [](start = 16, length = 34)

since there's a process.nextTick in here we're going to have a weird moment when the callback is "on hold until next tick" while the state machine transitions to idle which is, in case of changes in machina, might lead to some weird side effects. 

Have you considered passing the callback and error in the transition('idle') call and dealing with those in the _onEnter() of the idle state?

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,271 @@
...
+              } else {
+                /* Codes_SRS_NODE_PROVISIONING_HTTP_18_029: [ If the registration response has a success code with a 'status' that is not 'Assigned' or  'Assigning', `register` shall call the callback with a `SyntaxError` error. ] */
+                this._safeCallback(callback, new SyntaxError('badly formed registration response'));
+                this._fsm.transition('idle');
+              }
+            });
+          },
+          register: (authorization, forceRegistration, body, callback) => {
+            /* Codes_SRS_NODE_PROVISIONING_HTTP_18_036: [ `register` shall call the `callback` with an `InvalidOperationError` if it is called while a previous registration is in progress. ] */
+            callback(new errors.InvalidOperationError('registration already in progress.'));
```
pierreca: >callback(new errors.InvalidOperationError('registration already in progress.')); [](start = 12, length = 80)

missed that requirement earlier - should this operation be idempotent if the values are the same or on the contrary cancel and start a new one? just thinking of potential retry logic in the future?

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,271 @@
...
+  register(authorization: string, forceRegistration: boolean, body: any, callback: ProvisioningResponseCallback): void {
+    /* Codes_SRS_NODE_PROVISIONING_HTTP_18_002: [ `register` shall throw a `ReferenceError` if `callback` is falsy. ] */
+    if (!callback) {
+      throw new ReferenceError();
+    }
+
+    this._fsm.handle('register', authorization, forceRegistration, body, callback);
+  }
+
+  disconnect(callback?: (err: Error) => void): void {
```
pierreca: >disconnect [](start = 2, length = 10)

not really disconnect - more like cancel() no?

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,271 @@
...
+          }
+        }
+      }
+    });
+  }
+
+  register(authorization: string, forceRegistration: boolean, body: any, callback: ProvisioningResponseCallback): void {
+    /* Codes_SRS_NODE_PROVISIONING_HTTP_18_002: [ `register` shall throw a `ReferenceError` if `callback` is falsy. ] */
+    if (!callback) {
+      throw new ReferenceError();
```
pierreca: error message explaining which argument is mandatory would be nice.
Also, the compiler takes care of most of those for us, since this won't be public API per se, maybe we don't need it at all.

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,271 @@
...
+    /* Codes_SRS_NODE_PROVISIONING_HTTP_18_034: [ disconnect will cancel any http operations that are in progress ] */
+    /* Codes_SRS_NODE_PROVISIONING_HTTP_18_035: [ disconnect will cause polling to cease ] */
+    if (this._currentHttpRequest != null) {
+      this._currentHttpRequest.abort();
+      this._currentHttpRequest = null;
+    }
+    if (this._pollingTimer != null) {
+      clearTimeout(this._pollingTimer);
+      this._pollingTimer = null;
+    }
```
pierreca: timer should be cancelled first I think, since the http request abort may not by synchronous - or is it?

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,271 @@
...
+  }
+
+  private _sendRegistrationRequest(callback: (err: Error, body?: any) => void): void {
+    /* Codes_SRS_NODE_PROVISIONING_HTTP_18_009: [ `register` shall POST the registration request to 'https://global.azure-devices-provisioning.net/{idScope}/registrations/{registrationId}' ] */
+    /* Codes_SRS_NODE_PROVISIONING_HTTP_18_005: [ The registration request shall include the current `api-version` as a URL query string value named 'api-version'. ] */
+    let path: string = '/' + this._idScope + '/registrations/' + this._registrationId + '?api-version=' + this._apiVersion;
```
pierreca: >let path: string = '/' + this._idScope + '/registrations/' + this._registrationId + '?api-version=' + this._apiVersion; [](start = 4, length = 119)

looks like this path is the same in _sendRegistrationRequest and send OperationStatusRequest - maybe needs to be factored in somewhere.

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,271 @@
...
+    /* Codes_SRS_NODE_PROVISIONING_HTTP_18_037: [ The operation status request shall include the current `api-version` as a URL query string value named 'api-version'. ] */
+    let path: string = '/' + this._idScope + '/registrations/' + this._registrationId + '?api-version=' + this._apiVersion;
+
+    /* Codes_SRS_NODE_PROVISIONING_HTTP_18_020: [ The operation status request shall have the following in the Http header:
+      Accept: application/json
+      Content-Type: application/json; charset=utf-8 ] */
+    let httpHeaders = {
+      'Accept' : 'application/json',
+      'Content-Type' : 'application/json; charset=utf-8'
+    };
```
pierreca: if headers are the same everywhere maybe we can factor that in somewhere?

common/transport/http/src/rest_api_client.ts
```diff
@@ -142,6 +145,8 @@ export class RestApiClient {
     }
 
     request.end();
+
+    return request;
```
pierreca: >return request; [](start = 4, length = 15)

I'm really not a fan of leaking the request object. The whole purpose of the RestApiClient is to abstract it. 

If there really is a need to be able to abstract request, let's try to think about a way to surface this without leaking the node API layer.

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,271 @@
...
+
+  disconnect(callback?: (err: Error) => void): void {
+    this._fsm.handle('disconnect', callback);
+  }
+
+  private _cancelCurrentOperation(callback: (err: Error) => void): void {
+    /* Codes_SRS_NODE_PROVISIONING_HTTP_18_034: [ disconnect will cancel any http operations that are in progress ] */
+    /* Codes_SRS_NODE_PROVISIONING_HTTP_18_035: [ disconnect will cause polling to cease ] */
+    if (this._currentHttpRequest != null) {
+      this._currentHttpRequest.abort();
```
pierreca: > this._currentHttpRequest.abort(); [](start = 5, length = 34)

what do we gain aborting the request? can't we just let it die and not care?

common/core/src/errors.ts
BertKleewein: generic.  updated comments to reflect this.

---
In reply to: [141476392](https://github.com/Azure/azure-iot-sdk-node/pull/93#discussion_r141476392) [](ancestors = 141476392)

provisioning/transport/http/src/http.ts
BertKleewein: I changed this to directly access the member.  I think we'll add a stronger typing of the response.

---
In reply to: [141480373](https://github.com/Azure/azure-iot-sdk-node/pull/93#discussion_r141480373) [](ancestors = 141480373)

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,271 @@
...
+            callback(new errors.InvalidOperationError('registration already in progress.'));
+          },
+          disconnect : (callback) => {
+            this._cancelCurrentOperation((err) => {
+              this._fsm.transition('idle');
+              this._safeCallback(callback);
+            });
+          }
+        },
+        waiting_to_poll: {
```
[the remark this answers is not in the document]
BertKleewein: Not sure what you mean by this.  I will discuss this in person.

---
In reply to: [141480867](https://github.com/Azure/azure-iot-sdk-node/pull/93#discussion_r141480867) [](ancestors = 141480867)

BertKleewein:
> http.on('operationStatusPollingInterval', (body) => {

I think once for every interval.  I suspect this will not be needed based on the evolution of the API but I'm going to leave this in for now.

---
In reply to: [332665785](https://github.com/Azure/azure-iot-sdk-node/pull/93#issuecomment-332665785) [](ancestors = 332665785)

---
Refers to: provisioning/transport/http/devdoc/http_requirements.md:25 in f95de3f. [](commit_id = f95de3f2ef2e8397ca3c6254f026ed618b091b8a, deletion_comment = False)

provisioning/transport/http/src/http.ts
BertKleewein: will take offline

---
In reply to: [141480500](https://github.com/Azure/azure-iot-sdk-node/pull/93#discussion_r141480500) [](ancestors = 141480500)

provisioning/transport/http/src/http.ts
BertKleewein: that was a typo.  the path was fixed.

---
In reply to: [141482380](https://github.com/Azure/azure-iot-sdk-node/pull/93#discussion_r141482380) [](ancestors = 141482380)

BertKleewein:
> **SRS_NODE_PROVISIONING_HTTP_18_018: [** `register` shall poll for operation status every `operationStatusPollingInterval` milliseconds **]**

the event name is "operationStatus".  default is 2000 ms, but we will also use the retry-after value

---
In reply to: [332666004](https://github.com/Azure/azure-iot-sdk-node/pull/93#issuecomment-332666004) [](ancestors = 332666004)

---
Refers to: provisioning/transport/http/devdoc/http_requirements.md:95 in f95de3f. [](commit_id = f95de3f2ef2e8397ca3c6254f026ed618b091b8a, deletion_comment = False)

provisioning/transport/http/src/http.ts
BertKleewein: I was wondering why you did the callback in the _onEnter that way.  I like it.  

---
In reply to: [141481212](https://github.com/Azure/azure-iot-sdk-node/pull/93#discussion_r141481212) [](ancestors = 141481212)

provisioning/transport/http/src/http.ts
BertKleewein: Good question, but I don't see any reason right now to change this.  

---
In reply to: [141481719](https://github.com/Azure/azure-iot-sdk-node/pull/93#discussion_r141481719) [](ancestors = 141481719)

provisioning/transport/http/src/http.ts
BertKleewein: agreed

---
In reply to: [141479907](https://github.com/Azure/azure-iot-sdk-node/pull/93#discussion_r141479907) [](ancestors = 141479907)

provisioning/transport/http/src/http.ts
BertKleewein: You're right.  That also allows us to not leak the ClientRequest from the RestApiClient interface.

---
In reply to: [141483242](https://github.com/Azure/azure-iot-sdk-node/pull/93#discussion_r141483242) [](ancestors = 141483242)

provisioning/transport/http/src/http.ts
BertKleewein: moot now that aborting the request is done.

---
In reply to: [141482210](https://github.com/Azure/azure-iot-sdk-node/pull/93#discussion_r141482210) [](ancestors = 141482210)

provisioning/transport/http/src/http.ts
BertKleewein: I suspect this will change after the transport interface takes shape (when  we add MQTT and AMQP)

---
In reply to: [141481857](https://github.com/Azure/azure-iot-sdk-node/pull/93#discussion_r141481857) [](ancestors = 141481857)

PR #96, opened 2017-09-29
Refactor MQTT Receiver into Mqtt class directly

pierreca:
# Description of the problem
The client/receiver pattern that we had in the transports makes state/connection handling complicated and does not help the retry logic scenario. As with AMQP, the MQTT receiver needs to be folded into the MQTT class

# Description of the solution
The code for the MqttReceiver class has been moved into the Mqtt class itself, which now returns itself when getReceiver is called, making this a non-breaking change. The tests have been updated to reflect changes to how we mocked thing but have been kept when it comes to functional requirements. the requirements themselves have been kept and applied to the Mqtt class.
In AMQP the device method client is separate - it is not the case here. separating it will be evaluated to understand if it makes sense or not (Mqtt doesn't require nearly as much state management as AMQP does)


device/transport/mqtt/devdoc/mqtt_receiver_requirements.md
```diff
@@ -2,26 +2,17 @@
...
 ## Overview
 Object used to subscribe to the Cloud-to-Device messages endpoint and receive messages for this device
+These requirements now apply to the `Mqtt` object as the receiver pattern is being deprecated.
 
 ## Example
 ```javascript
-var receiver = new MqttReceiver(
-        client,
-        'message_topic/to/subscribe/to'
-    );
```
anthonyvercolano: This example seems a bit lacking at this point.

device/transport/mqtt/src/mqtt.ts
```diff
@@ -547,4 +608,222 @@ export class Mqtt extends EventEmitter implements Client.Transport, StableConnec
...
     this._fsm.handle('getTwinReceiver', done);
   }
+
+  onDeviceMethod(methodName: string, callback: (methodRequest: DeviceMethodRequest, methodResponse: DeviceMethodResponse) => void): void {
+    this.on('method_' + methodName, callback);
+  }
+
+  private _setupSubscription(topic: TopicDescription): void {
+    debug('subscribe: ' + JSON.stringify(topic));
+    topic.subscribeInProgress = true;
```
anthonyvercolano: >topic.subscribeInProgress = true; [](start = 4, length = 33)

I realize that this is the old code.  But I get very leary of multiple booleans describing the state of a resource.  I wonder whether the case could be be to simply say subscribed = true at this point.  If the actual subscription fails then it could be set to false.  However, it would certainly seem as though the rest of the code base should be able to deal with something that says it's subscribed but then suddenly flips to false.  It could happen that way in the normal course of action, no?

Anyway, I don't expect you to change it at this point,  just something to think about.

device/transport/mqtt/src/mqtt.ts
```diff
@@ -547,4 +608,222 @@ export class Mqtt extends EventEmitter implements Client.Transport, StableConnec
...
+      // the beginning every time we use it.
+      const theTopic = this._topics[topicIndex];
+      theTopic.topicMatchRegex.lastIndex = 0;
+      if (theTopic.topicMatchRegex.test(topic)) {
+        targetTopic = theTopic;
+      }
+      return targetTopic !== null;
+    });
+    if (!!targetTopic) {
+      targetTopic.handler(topic, payload);
```
anthonyvercolano: The claim would be that this is self documenting code and that I should be able to understand exactly what is going on here.

It isn't, I don't.

Don't expect a change but, hey, descriptions about how something works conceptually are a good thing.

device/transport/mqtt/src/mqtt.ts
```diff
@@ -547,4 +608,222 @@ export class Mqtt extends EventEmitter implements Client.Transport, StableConnec
...
+function _parseMessage(topic: string, body: any): MethodMessage {
+  let url: URL.Url;
+  let path: string[];
+  let query: any;
+  try {
+    url = URL.parse(topic);
+    path = url.path.split('/');
+    query = querystring.parse(url.query);
+  } catch (err) {
+    return undefined;
```
anthonyvercolano: >  return undefined; [](start = 2, length = 19)

So, in _onDeviceMethod, things just silently fail if there is a problem HERE?

device/transport/mqtt/src/mqtt.ts
```diff
@@ -547,4 +608,222 @@ export class Mqtt extends EventEmitter implements Client.Transport, StableConnec
...
+  }
+
+  if (path.length > 0 && path[0] === '$iothub') {
+    let message = new MethodMessage();
+    if (path.length > 1 && path[1].length > 0) {
+      // create an object for the module; for example, $iothub/twin/...
+      // would result in there being a message.twin object
+      let mod = message[path[1]] = new MethodDescription();
+
+      // parse the request ID if there is one
```
anthonyvercolano: >parse the request ID if there is one [](start = 9, length = 36)

heh?  doesn't look like it's parsing to me.

device/transport/mqtt/src/mqtt.ts
```diff
@@ -547,4 +608,222 @@ export class Mqtt extends EventEmitter implements Client.Transport, StableConnec
...
+      // save the body
+      message.body = body;
+
+      // parse the verb
+      if (path.length > 2 && path[2].length > 0) {
+        mod.verb = path[2];
+
+        // This is a topic that looks like this:
+        //  $iothub/methods/POST/{method name}?$rid={request id}&{serialized properties}
+        // We parse the method name out.
```
anthonyvercolano: This is really helpful.  Thank you.

device/transport/mqtt/test/_mqtt_receiver_test.js
```diff
@@ -126,152 +91,129 @@ describe('MqttReceiver', function () {
...
         receiver.on('method_UpdateFirmware', function () { });
         receiver.on('method_Reboot', function () { });
         receiver.on('message', function () { });
         receiver.on('message', function () { });
 
         // assert
-        assert.isTrue(onSpy.calledOnce,
-          'mqttClient.on was called more than once');
-        assert.isTrue(onSpy.calledWith('message'),
+        assert.isTrue(fakeMqttBase.on.calledTwice,
```
anthonyvercolano: >fakeMqttBase.on.calledTwice [](start = 22, length = 27)

This seems less robust than the original.

device/transport/mqtt/src/mqtt.ts
pierreca: I believe things could indeed be simplified. and if we wanted to match AMQP we could also extract methods feature out in their own client. for now the goal is only to shuffle code around to enable the retry logic APIs. I think this feedback is worth addressing separately.

---
In reply to: [141919360](https://github.com/Azure/azure-iot-sdk-node/pull/96#discussion_r141919360) [](ancestors = 141919360)

device/transport/mqtt/test/_mqtt_receiver_test.js
pierreca: why? we're actually testing more stuff than before. 

---
In reply to: [141923278](https://github.com/Azure/azure-iot-sdk-node/pull/96#discussion_r141923278) [](ancestors = 141923278)

device/transport/mqtt/src/mqtt.ts
pierreca: added a few comments

---
In reply to: [141920173](https://github.com/Azure/azure-iot-sdk-node/pull/96#discussion_r141920173) [](ancestors = 141920173)

device/transport/mqtt/src/mqtt.ts
pierreca: haha fair. changed the comment.

---
In reply to: [141921865](https://github.com/Azure/azure-iot-sdk-node/pull/96#discussion_r141921865) [](ancestors = 141921865)

device/transport/mqtt/src/mqtt.ts
pierreca: if we receive something that can't be parsed, we swallow the exception (which would otherwise crash the program) and then return undefined. the return value of _parseMessage is checked by the caller, and if it is not undefined, an event to trigger the method handler is emitted. if it is undefined, nothing is emitted. Consider the fact that the server could be sending us garbage and we wouldn't want to crash. nevertheless we should log it so I added a log line here.

---
In reply to: [141921549](https://github.com/Azure/azure-iot-sdk-node/pull/96#discussion_r141921549) [](ancestors = 141921549)

PR #99, opened 2017-10-02
provisioning http object

BertKleewein:


build/dev-setup.sh
```diff
@@ -72,6 +72,14 @@ npm link
 npm run build
 [ $? -eq 0 ] || exit $?
 
+cd $node_root/provisioning/transport/http
+echo "\n-- Creating links for `pwd` --"
+npm link azure-iot-common
+npm link azure-iot-http-base
```
pierreca: >http [](start = 19, length = 4)

not sure if it matters but in other packages/files is linked http-base comes before common

common/transport/http/src/rest_api_client.ts
```diff
@@ -115,8 +117,7 @@ export class RestApiClient {
       headers['Content-Length'] = requestBodyStringSizeInBytes;
     }
 
-    /*Codes_SRS_NODE_IOTHUB_REST_API_CLIENT_16_008: [The `executeApiCall` method shall build the HTTP request using the arguments passed by the caller.]*/
-    const request = this._http.buildRequest(method, path, httpHeaders, this._config.host, (err, responseBody, response) => {
+    let requestCallback: (err: Error, responseBody: any, response: any) => void = (err, responseBody, response) =>  {
```
pierreca: >let [](start = 4, length = 3)

const?

common/transport/http/test/_rest_api_client_test.js
```diff
@@ -113,6 +113,23 @@ describe('RestApiClient', function() {
       client.executeApiCall('GET', '/fake/path', null, null, testCallback);
     });
 
+    /* Tests_SRS_NODE_IOTHUB_REST_API_CLIENT_18_002: [ If an `x509` cert was passed into the constructor via the `config` object, `executeApiCall` shall use it to establish the TLS connection. ] */
+    it('uses the x509 cert if passed', function(testCallback) {
+      var fakeHttpHelper = {
+        buildRequest: function(method, path, headers, host, x509, requestCallback) {
+          assert.equal('yes', x509.fake);
```
pierreca: >equal [](start = 17, length = 5)

we try to avoid .equal() now and instead use either .strictEqual() or .deepEqual(). In that case I believe you can just declare the fakeConfig object first, and just say assert.strictEqual(x509, fakeConfig.x509)

common/transport/http/test/_rest_api_client_test.js
```diff
@@ -113,6 +113,23 @@ describe('RestApiClient', function() {
       client.executeApiCall('GET', '/fake/path', null, null, testCallback);
     });
 
+    /* Tests_SRS_NODE_IOTHUB_REST_API_CLIENT_18_002: [ If an `x509` cert was passed into the constructor via the `config` object, `executeApiCall` shall use it to establish the TLS connection. ] */
+    it('uses the x509 cert if passed', function(testCallback) {
+      var fakeHttpHelper = {
+        buildRequest: function(method, path, headers, host, x509, requestCallback) {
```
pierreca: >function [](start = 22, length = 8)

super minor nitpick: i've read in style guides that for readability anonymous functions should have a space between 'function' and the opening '('. we've been inconsistent so far (and it's not a big issue) but I'd like us to slowly improve consistency by using 'function ()' with a space between 'function' and '()'

common/transport/http/test/_rest_api_client_test.js
```diff
@@ -113,6 +113,23 @@ describe('RestApiClient', function() {
...
+          assert.equal('yes', x509.fake);
+          testCallback();
+        }
+      };
+
+      var myFakeConfig = JSON.parse(JSON.stringify(fakeConfig));
+      myFakeConfig.x509 = { 'fake' : 'yes' };
+      var client = new RestApiClient(myFakeConfig, fakeAgent, fakeHttpHelper);
+      client.executeApiCall('GET', '/fake/path', null, null, function() {
+      });
```
pierreca: this can go on the previous line.

provisioning/transport/http/typings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "azure-iot-provisioning-http",
+  "main": "index.d.ts",
+  "dependencies": {},
+  "globalDependencies": {
+    "node": "registry:dt/node#6.0.0+20160813124416"
```
pierreca: >node": "registry:dt/node#6.0.0+20160813124416" [](start = 5, length = 46)

no need for a typings file if you have @types/node installed in your dependencies.

pierreca:
> 

since this is going public you should probably already add a warning saying that this package is not meant to be used directly but rather with a provisioning client that will be coming in the near future.

---
Refers to: provisioning/transport/http/readme.md:15 in ccd096c. [](commit_id = ccd096c38e8cd4241b6b68a6016a1fcb522be177, deletion_comment = False)

provisioning/transport/http/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "azure-iot-provisioning-http",
+  "version": "1.0.0",
```
pierreca: >1.0.0 [](start = 14, length = 5)

Since we're shooting for public preview before release, i'd set this to 0.0.1 right away (just to avoid publishing something by mistake)

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -72,14 +68,14 @@ The `constructor` creates an Http transport object used to communicate with the
 
 **SRS_NODE_PROVISIONING_HTTP_18_008: [** If `forceRegistration` is specified, the registration request shall include this as a query string value named 'forceRegistration' **]**
 
-**SRS_NODE_PROVISIONING_HTTP_18_009: [** `register` shall POST the registration request to 'https://global.azure-devices-provisioning.net/{idScope}/registrations/{registrationId}' **]**
+**SRS_NODE_PROVISIONING_HTTP_18_009: [** `register` shall PUT the registration request to 'https://global.azure-devices-provisioning.net/{idScope}/registrations/{registrationId}/register' **]**
+
+**SRS_NODE_PROVISIONING_HTTP_18_010: [** `register` shall wait for the response to the PUT request. **]**
```
pierreca: >wait [](start = 58, length = 4)

I'm ambivalent on this requirement because 'wait' may have the connotation of actively waiting. I'd rather stick to the simple "calls the callback with __ when ___"

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -72,14 +68,14 @@ The `constructor` creates an Http transport object used to communicate with the
...
 
-**SRS_NODE_PROVISIONING_HTTP_18_010: [** `register` shall wait for the response to the POST request. **]**
+**SRS_NODE_PROVISIONING_HTTP_18_011: [** If the registration request times out, `register` shall call the `callback` with with the lower level error **]**
```
pierreca: >with the lower level [](start = 122, length = 20)

if the error is fed directly to the transport-agnostic client (caller?) then the error should not be specific to the transport (the transport layer has the necessary knowledge to convert transport-specific to agnostic client). If it is necessary to also forward the internal error (for debugging purposes for example) i'd encourage making it a _field of the resulting transport-agnostic error

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -110,24 +106,22 @@ The `constructor` creates an Http transport object used to communicate with the
 
 **SRS_NODE_PROVISIONING_HTTP_18_022: [** operation status request polling shall be a GET operation sent to 'https://global.azure-devices-provisioning.net/{idScope}/registrations/{registrationId}/operations/{operationId}' **]**
 
-**SRS_NODE_PROVISIONING_HTTP_18_023: [** If the operation status request times out, `register` shall stop polling and call the `callback` with a `TimeoutError` error. **]**
+**SRS_NODE_PROVISIONING_HTTP_18_023: [** If the operation status request times out, `register` shall stop polling and call the `callback` with with the lower level error **]**
```
pierreca: >with the lower level [](start = 143, length = 20)

same thing here should be transport-agnostic I think (also duplicate 'with' word)

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -110,24 +106,22 @@ The `constructor` creates an Http transport object used to communicate with the
...
+**SRS_NODE_PROVISIONING_HTTP_18_025: [** If the body of the operation status response fails to deserialize, `register` will throw a `SyntaxError` error. **]**
 
 **SRS_NODE_PROVISIONING_HTTP_18_026: [** If the operation status response contains a failure status code, `register` shall stop polling and call the `callback` with an error created using `translateError`. **]**
 
 **SRS_NODE_PROVISIONING_HTTP_18_027: [** If the operation status response contains a success status code with a 'status' of 'Assigned', `register` shall stop polling and call the `callback` with `err` == null and the body containing the deserialized body. **]**
 
-**SRS_NODE_PROVISIONING_HTTP_18_028: [** If the operation status response contains a success status code with a 'status' that is not 'Assigned', `register` shall fire an `operationStatus` event with the deserialized body and continue polling. **]**
+**SRS_NODE_PROVISIONING_HTTP_18_028: [** If the operation status response contains a success status code with a 'status' that is 'Assigning', `register` shall fire an `operationStatus` event with the deserialized body and continue polling. **]**
```
pierreca: >continue polling [](start = 222, length = 16)

I don't see any requirement that includes using the Retry-After header ?

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  private _idScope: string;
+  private _registrationId: string;
+  private _body: any;
+  private _forceRegistration: boolean;
+  private _operationId: string;
+
+  private _fsm: machina.Fsm;
+  private _apiVersion: string = '2017-08-31-preview';
+  private _provisioningServer: string = 'global.azure-devices-provisioning.net';
```
pierreca: >'global.azure-devices-provisioning.net' [](start = 40, length = 39)

Do I remember some discussion about not having URLs hard coded just in case to support all clouds (public and not?) need to double check.

provisioning/transport/http/devdoc/http_requirements.md
```diff
@@ -110,24 +106,22 @@ The `constructor` creates an Http transport object used to communicate with the
...
 
-**SRS_NODE_PROVISIONING_HTTP_18_030: [** If the operation status response has a success code with a 'status' that is not 'Assigned' or  'Assigning', `register` shall call the callback with a `SyntaxError` error and stop polling. **]**
+**SRS_NODE_PROVISIONING_HTTP_18_030: [** If the operation status response has a success code with a 'status' that is any other value, `register` shall call the callback with a `SyntaxError` error and stop polling. **]**
 
 
 ### disconnect(callback: (err) => void);
```
pierreca: >disconnect [](start = 4, length = 10)

missing a few requirements here (depending on state)

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,251 @@
...
+            this._body = null;
+            this._forceRegistration = false;
+            this._operationId = '';
+            this._registrationCallback = null;
+            if (callback) {
+              callback(err, response);
+            }
+          },
+          register: (authorization, forceRegistration, body, callback) => {
+            this._registrationCallback = callback;
```
pierreca: >_registrationCallback [](start = 17, length = 21)

do you still need this if you're passing it into the transition call?

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,251 @@
...
+        idle: {
+          _onEnter: (callback, err, response) => {
+            this._pollingTimer = null;
+            this._restApiClient = null;
+            this._body = null;
+            this._forceRegistration = false;
+            this._operationId = '';
+            this._registrationCallback = null;
+            if (callback) {
+              callback(err, response);
```
pierreca: >response [](start = 28, length = 8)

nit: response is usually used for the HTTP response object and result for the actual result (usually the JSON.parsed body of the response)

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,251 @@
...
+            this._fsm.transition('registering', callback);
+          },
+          disconnect : (callback) => {
+            // nothing to do.
+            callback();
+          }
+        },
+        registering: {
+          _onEnter: (callback) => {
+            this._sendRegistrationRequest((err, responseBody) => {
```
pierreca: >responseBody [](start = 48, length = 12)

not sure due to naming: is it the object, or the JSON string that still needs to be parsed?

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,251 @@
...
+                    this._fsm.transition('idle', callback, err);
+                } else if (responseBody.status.toLowerCase() === 'assigning') {
+                  /* Codes_SRS_NODE_PROVISIONING_HTTP_18_016: [ If the registration response has a success code with a 'status' of 'Assigning', `register` shall fire an `operationStatus` event with the deserialized body ] */
+                  this.emit('operationStatus', responseBody);
+                  /* Codes_SRS_NODE_PROVISIONING_HTTP_18_017: [ If the registration response has a success code with a 'status' of 'Assigning', `register` shall start polling for operation updates ] */
+                  this._operationId = responseBody.operationId;
+                  this._fsm.transition('waiting_to_poll', callback);
+                } else if (responseBody.status.toLowerCase() === 'assigned') {
+                  /* Codes_SRS_NODE_PROVISIONING_HTTP_18_015: [ If the registration response has a success code with a 'status' of 'Assigned', `register` call the `callback` with `err` == `null` and result `containing` the deserialized body ] */
+                  this._fsm.transition('idle', callback, null, responseBody);
```
pierreca: >idle [](start = 40, length = 4)

so there is no "registered" state? we don't have a use for it? registration operations are idempotent right?

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,251 @@
...
+          disconnect : (callback) => {
+            this._cancelCurrentOperation((err) => {
+              this._fsm.transition('idle', callback, err);
+            });
+          }
+        },
+        polling: {
+          _onEnter: (callback) => {
+            this._sendOperationStatusRequest((err, body) => {
+              if (this._registrationInProgress()) { // make sure we weren't cancelled before doing something with the response
```
pierreca: >if (this._registrationInProgress()) [](start = 14, length = 35)

for these guys I believe it'd be nice to have an else clause with a debug message. generally speaking it'd be useful to add more debug() traces everywhere.

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,251 @@
...
+                  this.emit('operationStatus', body);
+                  this._fsm.transition('waiting_to_poll', callback);
+                } else if (body.status.toLowerCase() === 'assigned') {
+                  /* Codes_SRS_NODE_PROVISIONING_HTTP_18_027: [ If the operation status response contains a success status code with a 'status' of 'Assigned', `register` shall stop polling and call the `callback` with `err` == null and the body containing the deserialized body. ] */
+                  this._fsm.transition('idle', callback, null, body);
+                } else {
+                  /* Codes_SRS_NODE_PROVISIONING_HTTP_18_030: [ If the operation status response has a success code with a 'status' that is any other value, `register` shall call the callback with a `SyntaxError` error and stop polling. ] */
+                  this._fsm.transition('idle', callback, new SyntaxError('badly formed registration response'));
+                }
```
pierreca: looks very similar to the initial registration call response processing - maybe refactor inside one function? also, consider using a switch statement on body.status.toLowerCase()? technically it doesn't matter much but I feel it makes the code slightly more readable. it's a personal opinion though...

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,251 @@
...
+
+    /* Codes_SRS_NODE_PROVISIONING_HTTP_18_031: [ If `disconnect` is called while the registration request is in progress, `register` shall call the `callback` with an `OperationCancelledError` error. ] */
+    /* Codes_SRS_NODE_PROVISIONING_HTTP_18_033: [ If `disconnect` is called while the register is waiting between polls, `register` shall call the `callback` with an `OperationCancelledError` error. ] */
+    if (this._registrationCallback) {
+      let _callback: ProvisioningResponseCallback = this._registrationCallback;
+      this._registrationCallback = null;
+      _callback(new errors.OperationCancelledError());
+    }
+
+    callback();
```
pierreca: gut feeling that this should be handled by the FSM

provisioning/transport/http/src/http.ts
pierreca: found why. cancellation, right?

---
In reply to: [142297992](https://github.com/Azure/azure-iot-sdk-node/pull/99#discussion_r142297992) [](ancestors = 142297992)

provisioning/transport/http/src/http.ts
```diff
@@ -0,0 +1,251 @@
...
+        }
+      }
+    });
+  }
+
+  register(authorization: string | X509, forceRegistration: boolean, body: any, callback: ProvisioningResponseCallback): void {
+    this._fsm.handle('register', authorization, forceRegistration, body, callback);
+  }
+
+  disconnect(callback: (err: Error) => void): void {
```
pierreca: >disconnect [](start = 2, length = 10)

disconnect is kindof weird for a REST based thing. since we're using cancel in other places, wouldn't it be logical to call it cancel? or cancelCurrentOperation?

also, do other SDKs support cancelling a registration process? 

last but not least will we be able to do that with HTTP?

also, does the service cares to know we're cancelling?

provisioning/transport/http/typings.json
BertKleewein: You'll have to explain this to me later.

---
In reply to: [142263803](https://github.com/Azure/azure-iot-sdk-node/pull/99#discussion_r142263803) [](ancestors = 142263803)

provisioning/transport/http/devdoc/http_requirements.md
BertKleewein: Agreed.  I will flush this out when I add another transport and create an official interface.

---
In reply to: [142267085](https://github.com/Azure/azure-iot-sdk-node/pull/99#discussion_r142267085) [](ancestors = 142267085)

provisioning/transport/http/devdoc/http_requirements.md
BertKleewein: Yes, it's on my TODO list.

---
In reply to: [142272833](https://github.com/Azure/azure-iot-sdk-node/pull/99#discussion_r142272833) [](ancestors = 142272833)

provisioning/transport/http/devdoc/http_requirements.md
BertKleewein: sorted the requirements to show this better.  Taking this offline for thinking.

---
In reply to: [142297915](https://github.com/Azure/azure-iot-sdk-node/pull/99#discussion_r142297915) [](ancestors = 142297915)

provisioning/transport/http/src/http.ts
BertKleewein: Yes, will add a config object as part of the interface

---
In reply to: [142297735](https://github.com/Azure/azure-iot-sdk-node/pull/99#discussion_r142297735) [](ancestors = 142297735)

provisioning/transport/http/src/http.ts
BertKleewein: yes

---
In reply to: [142299672](https://github.com/Azure/azure-iot-sdk-node/pull/99#discussion_r142299672) [](ancestors = 142299672,142297992)

provisioning/transport/http/src/http.ts
BertKleewein: no registered state.  need to double check.  requests are not idempotent when forceRegistration is used.

---
In reply to: [142298471](https://github.com/Azure/azure-iot-sdk-node/pull/99#discussion_r142298471) [](ancestors = 142298471)

provisioning/transport/http/src/http.ts
BertKleewein: I agree.  I will sort this out after I think about the other protocols for a while.  The function name is a placeholder until I get a unified interface

---
In reply to: [142299914](https://github.com/Azure/azure-iot-sdk-node/pull/99#discussion_r142299914) [](ancestors = 142299914)

pierreca: please make sure for all TODO list items that you have associated PBIs. or else we're going to miss things.

---
In reply to: [142460536](https://github.com/Azure/azure-iot-sdk-node/pull/99#discussion_r142460536) [](ancestors = 142460536,142299914)

PR #100, opened 2017-10-04
transport interface, stubbed device client

BertKleewein:
# Description of the problem
The first provisioning transport (HTTP) was developed without defining a transport interface.  Now that I'm working on the second provisioning transport, I need to make this definition.

# Description of the solution
An interface file was created.  In order to do this, I created an initial provisioning client library to hold the interface.  Finally, the existing HTTP transport was updated to conform to the interface.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -214,6 +214,8 @@ device/transport/http/lib/
 device/transport/mqtt/lib/
 device/ts-samples/*.js
 
+provisioning/transport/http/lib/
+
 service/lib
 
 ts-e2e/lib
--- a/build/build.cmd
+++ b/build/build.cmd
@@ -93,6 +93,9 @@ if errorlevel 1 goto :cleanup
 call :lint-and-test %node-root%\service
 if errorlevel 1 goto :cleanup
 
+call :lint-and-test %node-root%\provisioning\transport\http
+if errorlevel 1 goto :cleanup
+
 if %e2e-tests%==1 (
   call :lint-and-test %node-root%\e2etests
   if errorlevel 1 goto :cleanup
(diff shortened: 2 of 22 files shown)


provisioning/device/src/transport_interface.ts
```diff
@@ -0,0 +1,59 @@
...
+'use strict';
+
+import { X509 } from 'azure-iot-common';
+
+export namespace DeviceProvisioningTransport {
+  /**
+   * @private
+   * Callback for responses from the previsioning service
+   */
+  export type ResponseCallback = (err?: Error, result?: any) => void;
```
anthonyvercolano: >result?: an [](start = 47, length = 11)

responses in our node packages typically have a third parameter that contains the raw response we get back from the service.  Utilized for debug purposes.

provisioning/device/typings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "azure-device-provisioning-client",
+  "main": "index.d.ts",
+  "dependencies": {},
+  "globalDependencies": {
+    "node": "registry:dt/node#6.0.0+20160813124416"
```
pierreca: >node [](start = 5, length = 4)

no need for typings (history: typings are a deprecated way of getting type definitions and where possible have been replaced by @types/<package> packages. in that case we already have the @types/node package so no need for this.

provisioning/device/src/transport_interface.ts
pierreca: true only for HTTP though, wouldn't surface this in the case of a transport-agnostic thing.

---
In reply to: [142788003](https://github.com/Azure/azure-iot-sdk-node/pull/100#discussion_r142788003) [](ancestors = 142788003)

PR #101, opened 2017-10-04
split mqtt into base and device

BertKleewein:
# Description of the problem
The MQTT protocol does not have a common/device split because the 2 modules were merged.  Because we're going to have a provisioning MQTT object in the near future, I needed to re-establish the split.

# Description of the solution
Moved code around.  Created a new module.


anthonyvercolano:
> provisioning/transport/http/lib/

One for http also?

---
Refers to: .gitignore:218 in bcce0ce. [](commit_id = bcce0ce0e25872e429295d6e1a4ad8e28b6741d2, deletion_comment = False)

anthonyvercolano:
> if errorlevel 1 goto :cleanup

one for mqtt also?

---
Refers to: build/build.cmd:100 in bcce0ce. [](commit_id = bcce0ce0e25872e429295d6e1a4ad8e28b6741d2, deletion_comment = False)

common/transport/mqtt/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+    "compilerOptions": {
+        "outDir": "./lib",
+        "target":"es5",
+        "sourceMap": true,
+        "declaration": true
+    },
+    "include": [
+        "typings/index.d.ts",
```
anthonyvercolano: > "typings/index.d.ts", [](start = 7, length = 22)

Needed?  I removed this from mine...

common/transport/mqtt/src/mqtt_base.ts
```diff
@@ -257,3 +256,13 @@ export class MqttBase extends EventEmitter {
...
 }
+
+export namespace MqttBase {
+  export interface TransportConfig {
+    host: string;
+    sharedAccessSignature?: string | SharedAccessSignature;
+    deviceId: string;
+    x509?: X509;
+  }
+}
```
anthonyvercolano: Do you really need to create a new type namespace and interface.  I was able to dispense with this in the provisioning code.

common/transport/mqtt/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "azure-iot-mqtt-base",
+  "version": "1.1.12",
```
pierreca: >"1.1.12" [](start = 13, length = 8)

we have a release at the end of the week so we need to get this right, right now. we need to double check if the interface is compatible with the previous version, or bump it to 2.0.0

common/transport/mqtt/tsconfig.json
pierreca: >"typings/index.d.ts", [](start = 8, length = 21)

no typings, you can remove this.

common/transport/mqtt/typings.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "azure-iot-mqtt-base",
+  "main": "index.d.ts",
+  "dependencies": {
```
pierreca: no typings used so you can remove this file.

device/transport/mqtt/test/_mqtt_twin_receiver_test.js
```diff
@@ -6,7 +6,7 @@
 var assert = require('chai').assert;
 
 var MqttTwinReceiver = require('../lib/mqtt_twin_receiver.js').MqttTwinReceiver;
-var MqttProvider = require('./_fake_mqtt.js');
+var MqttProvider = require('../../../../common/transport/mqtt/test/_fake_mqtt.js');
```
pierreca: >../../../../common/transport/mqtt/test/ [](start = 28, length = 39)

ugh that really isn't great

pierreca: I think i'd rather even duplicate the code. those fakes are old patterns we should do away with. Please add a PBI to take care of that and lets tackle it during the milestone. technical debt at its worst here.

---
In reply to: [142813241](https://github.com/Azure/azure-iot-sdk-node/pull/101#discussion_r142813241) [](ancestors = 142813241)

BertKleewein: I agree.  This ../ pattern was in device/transport/mqtt before the merge, so I didn't worry too much about removing it.  I like the idea of getting rid of this entirely.
PBI opened.

---
In reply to: [142813528](https://github.com/Azure/azure-iot-sdk-node/pull/101#discussion_r142813528) [](ancestors = 142813528,142813241)

BertKleewein:
> provisioning/transport/http/lib/

You mean MQTT, and, once  the library, yes

---
In reply to: [334302504](https://github.com/Azure/azure-iot-sdk-node/pull/101#issuecomment-334302504) [](ancestors = 334302504)

---
Refers to: .gitignore:218 in bcce0ce. [](commit_id = bcce0ce0e25872e429295d6e1a4ad8e28b6741d2, deletion_comment = False)

BertKleewein:
> if errorlevel 1 goto :cleanup

Yes, once I add it

---
In reply to: [334302611](https://github.com/Azure/azure-iot-sdk-node/pull/101#issuecomment-334302611) [](ancestors = 334302611)

---
Refers to: build/build.cmd:100 in bcce0ce. [](commit_id = bcce0ce0e25872e429295d6e1a4ad8e28b6741d2, deletion_comment = False)

common/transport/mqtt/package.json
BertKleewein: You know, even if the interface looks compatible, I don't want to support much older versions of the SDK with much newer versions of this file.  I'd rather just bump to 2.0.0 to show that this is a brand new thing that shouldn't be intermixed with old things.

---
In reply to: [142812748](https://github.com/Azure/azure-iot-sdk-node/pull/101#discussion_r142812748) [](ancestors = 142812748)

common/transport/mqtt/src/mqtt_base.ts
BertKleewein: Yeah, I'm not crazy about this, but I wanted to reduce the number of changes in the actual implementation. 

---
In reply to: [142805734](https://github.com/Azure/azure-iot-sdk-node/pull/101#discussion_r142805734) [](ancestors = 142805734)

PR #103, opened 2017-10-10
Add enableC2D, disableC2D, enableMethods and disableMethods to the Mqtt object

pierreca:
# Description of the problem
As with other transports we need to create a feedback channel from the transport to the client when enabling features that require persistent connections. 

# Description of the solution
This introduces methods and callbacks for C2D messaging and direct methods. Twin will come separately since it requires a bit of refactoring.

### DIFF
--- a/device/transport/mqtt/devdoc/mqtt_requirements.md
+++ b/device/transport/mqtt/devdoc/mqtt_requirements.md
@@ -274,6 +274,10 @@ The `getTwinReceiver` method creates a `MqttTwinReceiver` object for the twin re
 
 **SRS_NODE_DEVICE_MQTT_16_049: [** `enableC2D` shall subscribe to the MQTT topic for messages. **]**
 
+**SRS_NODE_DEVICE_MQTT_16_050: [** `enableC2D` shall call its callback with no arguments when the `SUBACK` packet is received. **]**
+
+**SRS_NODE_DEVICE_MQTT_16_052: [** `enableC2D` shall call its callback with an `Error` if subscribing to the topic fails. **]**
+
 ### enableMethods
 
 **SRS_NODE_DEVICE_MQTT_16_038: [** `enableMethods` shall connect the MQTT connection if it is disconnected. **]**
@@ -282,11 +286,17 @@ The `getTwinReceiver` method creates a `MqttTwinReceiver` object for the twin re
 
 **SRS_NODE_DEVICE_MQTT_16_040: [** `enableMethods` shall subscribe to the MQTT topic for direct methods. **]**
 
+**SRS_NODE_DEVICE_MQTT_16_051: [** `enableMethods` shall call its callback with no arguments when the `SUBACK` packet is received. **]**
+
+**SRS_NODE_DEVICE_MQTT_16_053: [** `enableMethods` shall call its callback with an `Error` if subscribing to the topic fails. **]**
+
 ### disableC2D
 
 **SRS_NODE_DEVICE_MQTT_16_041: [** `disableC2D` shall call its callback immediately if the MQTT connection is already disconnected. **]**
 
 **SRS_NODE_DEVICE_MQTT_16_042: [** `disableC2D` shall unsubscribe from the topic for C2D messages. **]**
+
+**SRS_NODE_DEVICE_MQTT_16_054: [** `disableC2D` shall call its callback with no arguments when the `UNSUBACK` packet is received. **]**
 
 **SRS_NODE_DEVICE_MQTT_16_043: [** `disableC2D` shall call its callback with an `Error` if an error is received while unsubscribing. **]**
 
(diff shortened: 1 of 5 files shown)


device/transport/mqtt/devdoc/mqtt_requirements.md
```diff
@@ -266,3 +266,34 @@ The `getTwinReceiver` method creates a `MqttTwinReceiver` object for the twin re
 **SRS_NODE_DEVICE_MQTT_18_007: [** If a twin receiver for this endpoint previously existed, the `getTwinReceiver` method should return the preexisting `MqttTwinReceiver` object as the second parameter of the `done` function with null as the first parameter. **]**
 
 
+### enableC2D
+
+**SRS_NODE_DEVICE_MQTT_16_047: [** `enableC2D` shall connect the MQTT connection if it is disconnected. **]**
```
BertKleewein: >MQTT connection if it is disconnected. [](start = 65, length = 38)

What's the contract for SUB/SUBACK.  Does this call done before or after the SUBACK is received?  Same for enableMethods.  I can't remember if there's an UNSUBACK, but this question may also apply to disable*

device/transport/mqtt/src/mqtt.ts
```diff
@@ -100,43 +100,68 @@ export class Mqtt extends EventEmitter implements Client.Transport, StableConnec
...
-                  eventName === 'message' ?
-                    this._topics.message :
-                    null;
-      if (!!topic) {
-        debug('removing listener for topic: ' + JSON.stringify(topic));
-        // decrement listeners count for this topic
+      if (eventName.indexOf('method_') === 0) {
+        const topic = this._topics.method;
+        --(topic.listenersCount);
+        if (topic.listenersCount === 0 && topic.subscribed === true) {
```
BertKleewein: >topic.subscribed === true [](start = 42, length = 25)

edge case: what if topic.subscribeInProgress is true ?

device/transport/mqtt/src/mqtt.ts
```diff
@@ -100,43 +100,68 @@ export class Mqtt extends EventEmitter implements Client.Transport, StableConnec
...
+              debug('error setting up subscription for ' + topic.name + ': ' + err.toString());
+            }
+          });
+        }
+      } else if (eventName === 'message') {
+        const topic = this._topics.message;
         --(topic.listenersCount);
-
-        // stop listening for MQTT events if our consumers stop listening for our events
         if (topic.listenersCount === 0 && topic.subscribed === true) {
```
BertKleewein: >topic.subscribed  [](start = 42, length = 17)

same question about topic.subscribeInProgress.

device/transport/mqtt/src/mqtt.ts
```diff
@@ -100,43 +100,68 @@ export class Mqtt extends EventEmitter implements Client.Transport, StableConnec
...
           // Codes_SRS_NODE_DEVICE_MQTT_RECEIVER_13_004: [ When there are no more listeners for the method event, the topic should be unsubscribed. ]
-          this._removeSubscription(topic);
+          this.disableC2D((err) => {
+            if (err) {
+              debug('error setting up subscription for ' + topic.name + ': ' + err.toString());
+            }
+          });
         }
+      } else {
+        debug('new listener for which there\'s nothing to subscribe to: ' + eventName);
```
BertKleewein: >new listener for which there\'s nothing to subscribe to [](start = 15, length = 55)

cut-and-paste error.

device/transport/mqtt/devdoc/mqtt_requirements.md
```diff
@@ -266,3 +266,34 @@ The `getTwinReceiver` method creates a `MqttTwinReceiver` object for the twin re
...
+
+**SRS_NODE_DEVICE_MQTT_16_048: [** `enableC2D` shall calls its callback with an `Error` object if it fails to connect. **]**
+
+**SRS_NODE_DEVICE_MQTT_16_049: [** `enableC2D` shall subscribe to the MQTT topic for messages. **]**
```
BertKleewein: >`enableC2D` shall subscribe to the MQTT topic for messages [](start = 35, length = 58)

What if the subscription fails or times out?  I'm not sure how likely this is, but both could theoretically happen.  Same enableMethods.  If there's an UNSUBACK, then this also applies to disableC2D and disableMethods.

device/transport/mqtt/src/mqtt.ts
pierreca: should be fine. we don't really care about sending multiple subscribe packets. if anything, I'd be worried that the first failed and we don't retry because it's already in progress.

we need better state management all around here though.. totally agree with the idea... out of scope for this PR though.

---
In reply to: [143822908](https://github.com/Azure/azure-iot-sdk-node/pull/103#discussion_r143822908) [](ancestors = 143822908)

PR #109, opened 2017-10-17
Remove Device Client State Machine

pierreca:
# Description of the problem
Now that all transports have state machines, so the client state machine becomes redundant and a source of bugs. Let's remove it before we can move on to adding the actual retry logic.

# Description of the solution
The state machine has been removed from the device client class, which now makes use of the enable/disable feature methods of the transports. this means we get all errors from the transports either through callbacks or the error event, which means we'll be able to introduce the retry logic next.

I've left the end to end tests untouched to prove that this is not breaking changes (not with the public client API anyway, we are indeed breaking some of the internal contracts between the client and the transports). Actually I've even re-enabled a couple of them to prove robustness.

### DIFF
--- a/device/core/src/client.ts
+++ b/device/core/src/client.ts
@@ -428,9 +428,9 @@ export class Client extends EventEmitter {
       throw new TypeError('callback\'s type is \'' + typeof(callback) + '\'. A function reference was expected.');
     }
 
-    // Codes_SRS_NODE_DEVICE_CLIENT_13_021: [ onDeviceMethod shall throw an Error if the underlying transport does not support device methods. ]
+    // Codes_SRS_NODE_DEVICE_CLIENT_13_021: [ onDeviceMethod shall throw an NotImplementedErrorError if the underlying transport does not support device methods. ]
     if (!((this._transport as DeviceMethodTransport).sendMethodResponse)) {
-      throw new Error('The transport for this client does not support device methods');
+      throw new errors.NotImplementedError('The transport for this client does not support device methods');
     }
 
     // Codes_SRS_NODE_DEVICE_CLIENT_13_023: [ onDeviceMethod shall throw an Error if a listener is already subscribed for a given method call. ]
--- a/device/core/src/client.ts
+++ b/device/core/src/client.ts
@@ -6,16 +6,15 @@
 import { Stream } from 'stream';
 import { EventEmitter } from 'events';
 import * as dbg from 'debug';
-const debug = dbg('azure-iot-device.Client');
-import * as machina from 'machina';
+const debug = dbg('azure-iot-device:Client');
 
-import { anHourFromNow, results, errors, Message, Receiver, TransportConfig } from 'azure-iot-common';
+import { anHourFromNow, results, errors, Message, TransportConfig } from 'azure-iot-common';
 import * as ConnectionString from './connection_string.js';
 import * as SharedAccessSignature from './shared_access_signature.js';
 import { BlobUploadClient } from './blob_upload';
 import { DeviceMethodRequest, DeviceMethodResponse } from './device_method';
 import { Twin } from './twin';
-import { StableConnectionTransport , DeviceMethodTransport, BatchingTransport, DeviceMethodReceiver, ClientConfig } from './interfaces';
+import { StableConnectionTransport , DeviceMethodTransport, ClientConfig } from './interfaces';
 
 function safeCallback(callback?: (err?: Error, result?: any) => void, error?: Error, result?: any): void {
   if (callback) callback(error, result);
@@ -49,12 +48,10 @@ export class Client extends EventEmitter {
   private _connectionString: string;
   private _useAutomaticRenewal: boolean;
   private _sasRenewalTimeout: number;
-  private _receiver: Receiver;
-  private _messageReceiverConnected: boolean;
   private _methodCallbackMap: any;
-  private _fsm: machina.Fsm;
   private _disconnectHandler: (err?: Error, result?: any) => void;
   private blobUploadClient: BlobUploadClient; // TODO: wrong casing/naming convention
+  private _c2dEnabled: boolean;
 
   /**
    * @constructor
@@ -69,6 +66,7 @@ export class Client extends EventEmitter {
     if (!transport) throw new ReferenceError('transport is \'' + transport + '\'');
 
     super();
+    this._c2dEnabled = false;
     /*Codes_SRS_NODE_DEVICE_CLIENT_16_026: [The Client constructor shall accept a connection string as an optional second argument] */
     this._connectionString = connStr;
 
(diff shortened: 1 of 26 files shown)


device/core/test/_client_test.js
```diff
@@ -942,157 +688,60 @@ describe('Client', function () {
...
-      });
-
-      client.on('message', function () {
+      client.on('error', function (err) {
+        assert.strictEqual(err, fakeError);
         testCallback();
-      });
+      })
+
+      // Calling 'on' twice to make sure it's called only once on the receiver.
```
BertKleewein: >'on' [](start = 17, length = 4)

This comment doesn't match the code.  Do you mean to call removeAllListeners twice?

device/core/src/client.ts
```diff
@@ -571,7 +323,11 @@ export class Client extends EventEmitter {
     /*Codes_SRS_NODE_DEVICE_CLIENT_16_016: [The ‘complete’ method shall throw a ReferenceError if the ‘message’ parameter is falsy.] */
     if (!message) throw new ReferenceError('message is \'' + message + '\'');
 
-    this._fsm.handle('complete', message, completeCallback);
+    this._transport.complete(message, (err, result) => {
+      if (completeCallback) {
```
BertKleewein: >completeCallback [](start = 10, length = 16)

Is it possible that the transport doesn't support completion?  If so, should we throw in this method and the following few methods?

device/core/src/client.ts
```diff
@@ -674,7 +441,7 @@ export class Client extends EventEmitter {
 
   private _addMethodCallback(methodName: string, callback: (request: DeviceMethodRequest, response: DeviceMethodResponse) => void): void {
     const self = this;
-    (this._receiver as DeviceMethodReceiver).onDeviceMethod(methodName, (message) => {
+    (this._transport as DeviceMethodTransport).onDeviceMethod(methodName, (message) => {
```
BertKleewein: >(this._transport as DeviceMethodTransport). [](start = 4, length = 43)

What will this do if the transport doesn't implement the DeviceMethodTransport interface?  Is that even possible?  Is the correct exception thrown?

device/core/src/client.ts
```diff
@@ -731,6 +503,31 @@ export class Client extends EventEmitter {
...
+      /*Codes_SRS_NODE_DEVICE_CLIENT_16_055: [The `close` method shall call the `closeCallback` function when done with either a single Error object if it failed or null and a results.Disconnected object if successful.]*/
+      safeCallback(closeCallback, err, result);
+    };
+
+    if (this._sasRenewalTimeout) {
+      clearTimeout(this._sasRenewalTimeout);
+    }
+
+    this._disableC2D(() => {
+      if (this._isImplementedInTransport('disconnect')) {
```
BertKleewein: >(this._isImplementedInTransport('disconnect')) [](start = 9, length = 46)

My HTTP provisioning transports has empty implementations for connect and disconnect rather than relying on the outer code to take care of it.  That way there's a single transport interface that always operates the same way.

Is there any value in updating my transports to have multiple interfaces depending on the need to connect/disconnect?  My gut says that adding that interface doesn't help me at all, but I"d like to hear your ideas.

device/transport/amqp/test/_amqp_test.js
```diff
@@ -925,7 +898,6 @@ describe('Amqp', function () {
         });
       });
 
-      // skipping until we have the logic in the device client to handle that
       it.skip('forwards errors from the D2C link', function (testCallback) {
```
BertKleewein: >it.skip [](start = 6, length = 7)

You removed the comment.  Did you mean to also remove the call to .skip?

BertKleewein:
>       receiver.setOptions({ interval: 5 });

I haven't seen anything yet to indicate that the transport is using a fake clock.  Does this have to be 5 seconds?  Wouldn't 1 second be just as useful but save some time on the unit tests?  Looks like a 1:10 savings.
There has to be a fake clock, right?

---
Refers to: device/transport/http/test/_http_test.js:241 in a68fcb0. [](commit_id = a68fcb0ddc3e007a662ba42e28c87ff3bb9f62cf, deletion_comment = False)

device/transport/http/test/_http_test.js
```diff
@@ -403,8 +377,32 @@ describe('HttpReceiver', function () {
       var received = 0;
       receiver.on('message', function () {
         received++;
-        if (received === msgCount) done();
+        if (received === msgCount) {
```
BertKleewein: >=== [](start = 21, length = 3)

recommend >= to catch the case where you receive more messages than expected.

device/core/src/client.ts
```diff
@@ -493,7 +235,15 @@ export class Client extends EventEmitter {
    */
   sendEventBatch(messages: Message[], sendEventBatchCallback?: (err?: Error, result?: results.MessageEnqueued) => void): void {
     /*Codes_SRS_NODE_DEVICE_CLIENT_05_008: [The sendEventBatch method shall send the list of events (indicated by the messages argument) via the transport associated with the Client instance.]*/
-    this._fsm.handle('sendEventBatch', messages, sendEventBatchCallback);
+    if (this._transport.sendEventBatch) {
```
anthonyvercolano: >(this._transport.sendEventBatch) [](start = 7, length = 32)

Two different ways to determine if a capability is present in the  transport?

device/transport/amqp/src/amqp.ts
```diff
@@ -259,11 +218,12 @@ export class Amqp extends EventEmitter implements Client.Transport, StableConnec
             callback();
           },
           onDeviceMethod: (methodName, callback) => {
+            // TODO - confirm this is still what we want now that we have enableMethods
```
anthonyvercolano: >// TODO - confirm this is still what we want now that we have enableMethods [](start = 12, length = 75)

!

device/core/test/_client_test.js
pierreca: no, I guess it's just old copy/pasta. removing the comment.

---
In reply to: [145237862](https://github.com/Azure/azure-iot-sdk-node/pull/109#discussion_r145237862) [](ancestors = 145237862)

device/core/src/client.ts
pierreca: the fact is that all transports have these methods, and it's the transport (MQTT in that case) that ends up throwing. it's a little different for other features (where some methods are missing from transports) but I haven't had time to unify that. Maybe later :)

---
In reply to: [145239756](https://github.com/Azure/azure-iot-sdk-node/pull/109#discussion_r145239756) [](ancestors = 145239756)

device/core/src/client.ts
pierreca: it's a little convoluted but the test for device methods support is in _validateDeviceMethodInputs. so an exception will be thrown alright, just from a weird place I guess.

---
In reply to: [145240275](https://github.com/Azure/azure-iot-sdk-node/pull/109#discussion_r145240275) [](ancestors = 145240275)

pierreca: I've modified the code to actually throw a NotImplementedError instead of a generic error though. thanks for pointing this out.

---
In reply to: [145266360](https://github.com/Azure/azure-iot-sdk-node/pull/109#discussion_r145266360) [](ancestors = 145266360,145240275)

device/core/src/client.ts
pierreca: yeah those tests are remnants of not having type safety and proper interfaces. I wish all transports implemented the same interface and would throw clear NotImplementedErrors themselves. probably a fix for later though, the scope of this PR is large enough already.

---
In reply to: [145241601](https://github.com/Azure/azure-iot-sdk-node/pull/109#discussion_r145241601) [](ancestors = 145241601)

device/transport/amqp/test/_amqp_test.js
pierreca: good catch, went back on that. I'm under the impression that we can ignore this error because any transport can reestablish the D2C link when sending the next message so we don't need to "retry" until we have something to send... 

---
In reply to: [145242932](https://github.com/Azure/azure-iot-sdk-node/pull/109#discussion_r145242932) [](ancestors = 145242932)

pierreca: I added a comment to clarify why we're skipping

---
In reply to: [145268291](https://github.com/Azure/azure-iot-sdk-node/pull/109#discussion_r145268291) [](ancestors = 145268291,145242932)

device/transport/http/test/_http_test.js
pierreca: doesn't hurt to be defensive...

---
In reply to: [145245315](https://github.com/Azure/azure-iot-sdk-node/pull/109#discussion_r145245315) [](ancestors = 145245315)

device/core/src/client.ts
pierreca: yeah. I hate that. I'll fix it in a separate PR though.

---
In reply to: [145258076](https://github.com/Azure/azure-iot-sdk-node/pull/109#discussion_r145258076) [](ancestors = 145258076)

pierreca:
>       receiver.setOptions({ interval: 5 });

it is a fake clock - it's set up in the beforeEach function line 226

---
In reply to: [337359136](https://github.com/Azure/azure-iot-sdk-node/pull/109#issuecomment-337359136) [](ancestors = 337359136)

---
Refers to: device/transport/http/test/_http_test.js:241 in a68fcb0. [](commit_id = a68fcb0ddc3e007a662ba42e28c87ff3bb9f62cf, deletion_comment = False)

device/transport/amqp/src/amqp.ts
pierreca: yup, should've removed that as well as the call the enableMethods below. very very good catch.

---
In reply to: [145259295](https://github.com/Azure/azure-iot-sdk-node/pull/109#discussion_r145259295) [](ancestors = 145259295)

PR #111, opened 2017-10-19
Align transport interfaces, introduce Mqtt Twin state machine and fix for #106

pierreca:
# Description of the problem
Right now we're operating with weird partial interfaces and have to check at the client level whether some features are implemented in transports or not. Also, see #106 
Also, the Mqtt Twin relies a lot on the Client calling the right things in the right order to subscribe to topics and forward messages. it'd be great to have a little less strong coupling between the Client Twin class and the transports.

# Description of the solution
Make all transports implement the same interface and have transports throw the `NotImplementedError` instead of the `Client`.
For #106: just check that the message property is populated before calling `indexOf` on it
Regarding the MQTT Twin, the first step is to introduce a state machine and higher-level APIs that are going to enable retries. Then we can clean up the contract between the `Client` and the transports.


device/transport/http/test/_http_test.js
```diff
@@ -46,6 +48,69 @@ describe('Http', function () {
...
+            testCallback();
+          });
+        });
+      });
+    });
+
+    /*Tests_SRS_NODE_DEVICE_HTTP_16_031: [The `disconnect` method shall call its callback with a `null` first argument and a `results.Disconnected` second argument after successfully disabling the C2D receiver (if necessary).]*/
+    it('calls its callback with no error and a results.Disconnected if the C2D receiver is not running', function (testCallback) {
+      var http = new Http();
+      http.disconnect(function (err, result) {
```
anthonyvercolano: >http.disconnect [](start = 6, length = 15)

Maybe test that disableC2D not called?

device/transport/mqtt/devdoc/mqtt_requirements.md
```diff
@@ -309,3 +319,13 @@ The `getTwinReceiver` method creates a `MqttTwinReceiver` object for the twin re
...
 
 **SRS_NODE_DEVICE_MQTT_16_046: [** `disableMethods` shall call its callback with an `Error` if an error is received while unsubscribing. **]**
+
+### disableTwin
+
+**SRS_NODE_DEVICE_MQTT_16_062: [** `disableTwin` shall call its callback immediately if the MQTT connection is already disconnected. **]**
+
+**SRS_NODE_DEVICE_MQTT_16_063: [** `disableTwin` shall unsubscribe from the topics for twin messages. **]**
+
+**SRS_NODE_DEVICE_MQTT_16_064: [** `disableTwin` shall call its callback with no arguments when the `UNSUBACK` packet is received. **]**
```
anthonyvercolano: >`UNSUBACK` [](start = 100, length = 10)

If no UNSUBACK received?

device/transport/mqtt/src/mqtt_twin_receiver.ts
```diff
@@ -47,80 +50,122 @@ export class MqttTwinReceiver extends EventEmitter {
...
+        subscribing: {
+          _onEnter: (callback) => {
+            debug('subscribing to twin topics');
+            this._mqtt.on('message', this._onMqttMessage.bind(this));
+            this._mqtt.subscribe(responseTopic, { qos: 0 }, (err, responseTopicResult) => {
+              if (err) {
+                this._fsm.transition('unsubscribed', callback, err);
+              } else {
+                debug('subscribed to response topic');
+                this._mqtt.subscribe(postTopic, { qos: 0 }, (err, postTopicResult) => {
```
anthonyvercolano: >subscribe( [](start = 27, length = 10)

If this fails but the response topic subscription succeeds you'll pop right to unsubscribed with an outstanding subscription to the response.  Is this ok?

device/transport/mqtt/src/mqtt_twin_receiver.ts
```diff
@@ -47,80 +50,122 @@ export class MqttTwinReceiver extends EventEmitter {
...
+          subscribe: (callback) => callback(),
+          unsubscribe: (callback) => this._fsm.transition('unsubscribing', callback)
+        },
+        unsubscribing: {
+          _onEnter: (callback) => {
+            /* Codes_SRS_NODE_DEVICE_MQTT_TWIN_RECEIVER_18_005: [** When there are no more listeners for the `response` event, the topic should be unsubscribed **]** */
+            let unsubscribeError;
+            this._mqtt.unsubscribe(responseTopic, (err) => {
+              unsubscribeError = err;
+              /* Codes_SRS_NODE_DEVICE_MQTT_TWIN_RECEIVER_18_021: [** When there are no more listeners for the post event, the topic should be unsubscribed. **]** */
```
anthonyvercolano: >should be unsubscribed. [](start = 133, length = 23)

If not unsuback?

device/core/src/client.ts
```diff
@@ -8,13 +8,13 @@ import { EventEmitter } from 'events';
 import * as dbg from 'debug';
 const debug = dbg('azure-iot-device:Client');
 
-import { anHourFromNow, results, errors, Message, TransportConfig } from 'azure-iot-common';
```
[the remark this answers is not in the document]
pierreca: yeah it's probably a mistake to remove it. I'll bring it back.

---
In reply to: [145744417](https://github.com/Azure/azure-iot-sdk-node/pull/111#discussion_r145744417) [](ancestors = 145744417)

device/transport/mqtt/devdoc/mqtt_requirements.md
pierreca: I feel it's the job of the mqtt library to come back either with unsuback or an error. if we have to introduce timeouts for this type of operations I want a cohesive approach, not one-offs on some specific APIs.

---
In reply to: [145747906](https://github.com/Azure/azure-iot-sdk-node/pull/111#discussion_r145747906) [](ancestors = 145747906)

device/transport/mqtt/src/mqtt_twin_receiver.ts
pierreca: probably not...

---
In reply to: [145750640](https://github.com/Azure/azure-iot-sdk-node/pull/111#discussion_r145750640) [](ancestors = 145750640)

device/transport/amqp/src/amqp_ws.ts
```diff
@@ -4,21 +4,21 @@
...
+import { Client } from 'azure-iot-device';
 
 /**
  * Constructs a transport object that can be used by the device {@link azure-iot-device.Client} to send and receive messages to and from an IoT Hub instance, using the AMQP protocol over secure websockets.
  * This class overloads the constructor of the base {@link azure-iot-device-amqp.Amqp} class from the AMQP transport, and inherits all methods from it.
  *
  * @augments module:azure-iot-device-amqp.Amqp
  */
-export class AmqpWs extends Amqp implements Client.Transport, StableConnectionTransport {
+export class AmqpWs extends Amqp implements Client.Transport {
```
BertKleewein: >implements Client.Transport [](start = 33, length = 27)

sillhy nit.  It inherits the Client.Transport implmementation from Amqp so this shouldn't be necessary

device/transport/mqtt/src/mqtt_ws.ts
```diff
@@ -13,13 +13,13 @@ import { Mqtt } from './mqtt';
 /*Codes_SRS_NODE_DEVICE_MQTT_12_001: [The constructor shall accept the transport configuration structure.]*/
 /*Codes_SRS_NODE_DEVICE_MQTT_12_002: [The constructor shall store the configuration structure in a member variable.]*/
 /*Codes_SRS_NODE_DEVICE_MQTT_12_003: [The constructor shall create an base transport object and store it in a member variable.]*/
-export class MqttWs extends Mqtt implements Client.Transport, StableConnectionTransport, TwinTransport {
+export class MqttWs extends Mqtt implements Client.Transport {
```
BertKleewein: >implements Client.Transpor [](start = 33, length = 26)

same nit as in amqp_ws

BertKleewein:
>   updateSharedAccessSignature (sharedAccessSignature: string, done?: (err?: Error, result?: any) => void): void {

remove the ? here

---
Refers to: device/transport/mqtt/src/mqtt.ts:475 in d779f78. [](commit_id = d779f784c2c3cdfa037617aa28877fd79023c096, deletion_comment = False)

device/transport/mqtt/src/mqtt_twin_receiver.ts
```diff
@@ -47,80 +50,136 @@ export class MqttTwinReceiver extends EventEmitter {
...
+
+    const messageHandler = this._onMqttMessage.bind(this);
+
+    this._fsm = new machina.Fsm({
+      namespace: 'device-mqtt',
+      initialState: 'unsubscribed',
+      states: {
+        unsubscribed: {
+          _onEnter: (callback, err) => {
+            this._mqtt.removeListener('message', messageHandler);
```
BertKleewein: >this._mqtt.removeListener('message', messageHandler); [](start = 12, length = 53)

I am so annoyed that a 'message' event could be an MQTT packet that is emitted by a lower layer or it could be a C2D message that is emitted at this layer.  I doubt if we want to change this right now, but AAARGH.

device/transport/mqtt/src/mqtt_twin_receiver.ts
```diff
@@ -47,80 +50,136 @@ export class MqttTwinReceiver extends EventEmitter {
...
+          _onEnter: (callback) => {
+            /* Codes_SRS_NODE_DEVICE_MQTT_TWIN_RECEIVER_18_021: [** When there are no more listeners for the post event, the topic should be unsubscribed. **]** */
+            this._mqtt.unsubscribe(postTopic, (err) => {
+              this._fsm.transition('unsubscribingFromResponseTopic', callback, err);
+            });
+          },
+          '*': () => this._fsm.deferUntilTransition()
+        },
+        unsubscribingFromResponseTopic: {
+          _onEnter: (callback, previousUnsubscribeError) => {
```
BertKleewein: >previousUnsubscribeError [](start = 31, length = 24)

naming: this could also be a subscribe error from subscripbingToPostTopic.

device/transport/amqp/src/amqp.ts
```diff
@@ -47,11 +47,11 @@ deviceId – (string) the identifier of a device registered with the IoT Hub
 sharedAccessSignature – (string) the shared access signature associated with the device registration.] */
 /*Codes_SRS_NODE_DEVICE_AMQP_RECEIVER_16_001: [The `Amqp` constructor shall implement the `Receiver` interface.]*/
 /*Codes_SRS_NODE_DEVICE_AMQP_RECEIVER_16_002: [The `Amqp` object shall inherit from the `EventEmitter` node object.]*/
-export class Amqp extends EventEmitter implements Client.Transport, StableConnectionTransport, Receiver {
```
BertKleewein: >implements Client.Transport, StableConnectionTransport, Receiver  [](start = 39, length = 65)

thanks for taking my prior feedback into account.

device/transport/mqtt/src/mqtt_twin_receiver.ts
pierreca: take a deep breath. it's all internal implementation details. FWIW I agree - 'common' message translation should happen only in the highest layer.

---
In reply to: [145828101](https://github.com/Azure/azure-iot-sdk-node/pull/111#discussion_r145828101) [](ancestors = 145828101)

PR #114, opened 2017-10-20
Add retry logic in the device client for C2D, D2C, device methods, file upload and SAS update. (Twin missing)

pierreca:
# Description of the problem
Now that the transports can maintain their own state, and the device client has been reduced to the bare minimum, we can start adding retry operations.

# Description of the solution
- new `setRetryPolicy` API
- use of `RetryOperation` and `RetryPolicy` objects
- `ExponentialBackoffWithJitter` by default
- the disconnect tests have been updated to reflect the new logic

# Known issues:
- cleanly terminating the AMQP connection isn't picked up anymore as a disconnect event. this is problematic but I have a hunch it has to do with me removing the listener for `connection:closed` on the `amqp10.Client` object. I'll bring this back into working order before checking in.
- There are a couple of unit tests that don't pass anymore in the AMQP transport and Client package. I still have to figure out if they are irrelevant now, or if that is hiding a true issue.
- The Twin feature requires massive refactoring to enable retry operations and is out of scope for this PR.

### DIFF
--- a/common/transport/amqp/src/amqp.ts
+++ b/common/transport/amqp/src/amqp.ts
@@ -14,7 +14,7 @@ import { AmqpLink } from './amqp_link_interface';
 
 import * as dbg from 'debug';
 import * as async from 'async';
-const debug = dbg('amqp-common:amqp');
+const debug = dbg('azure-iot-amqp-base:Amqp');
 
 const _amqpClientError = 'client:errorReceived';
 
@@ -90,6 +90,16 @@ export class Amqp {
       }
     }, amqp10.Policy.EventHub));
 
+    const amqpErrorHandler = (err) => {
+      debug('amqp10 client error: ' + err.toString());
+      this._fsm.handle('amqpError', err, () => {
+        if (this._disconnectHandler) {
+          debug('calling upper layer disconnect handler');
+          this._disconnectHandler(err);
+        }
+      });
+    };
+
     this._fsm = new machina.Fsm({
       namespace: 'amqp-base',
       initialState: 'disconnected',
@@ -155,15 +165,6 @@ export class Amqp {
             let connectError = null;
             const connectErrorHandler = (err) => {
               connectError = err;
-            };
-
-            const amqpErrorHandler = (err) => {
-              debug('amqp10 client error: ' + err.toString());
-              this._fsm.handle('amqpError', err, () => {
-                if (this._disconnectHandler) {
-                  this._disconnectHandler(err);
-                }
-              });
             };
 
             this._amqp.on(_amqpClientError, connectErrorHandler);
@@ -339,7 +340,7 @@ export class Amqp {
 
               if (err) {
                 debug('forceDetaching link');
-                link.forceDetach(err);
+                link.forceDetach();
                 callback();
               } else {
                 debug('cleanly detaching link');
@@ -367,6 +368,7 @@ export class Amqp {
               async.each(remainingLinks, detachLink, () => {
                 disconnect((disconnectError) => {
                   const finalError = err || disconnectError;
+                  this._amqp.removeListener(_amqpClientError, amqpErrorHandler);
                   this._fsm.transition('disconnected', disconnectCallback, finalError);
                 });
               });
@@ -408,9 +410,9 @@ export class Amqp {
    */
   setDisconnectHandler(disconnectCallback: GenericAmqpBaseCallback<any>): void {
     this._disconnectHandler = disconnectCallback;
-    this._amqp.on('connection:closed', () => {
-      this._disconnectHandler(new Error('amqp10: connection closed'));
-    });
+    // this._amqp.on('connection:closed', () => {
+    //   this._disconnectHandler(new errors.NotConnectedError('amqp10: connection closed'));
+    // });
   }
 
   /**
--- a/common/transport/amqp/src/amqp.ts
+++ b/common/transport/amqp/src/amqp.ts
@@ -6,7 +6,7 @@
 import * as amqp10 from 'amqp10';
 import * as machina from 'machina';
 import { AmqpMessage } from './amqp_message';
-import { results, Message } from 'azure-iot-common';
+import { errors, results, Message } from 'azure-iot-common';
 import { ClaimsBasedSecurityAgent } from './amqp_cbs';
 import { SenderLink } from './sender_link';
 import { ReceiverLink } from './receiver_link';
@@ -92,12 +92,7 @@ export class Amqp {
 
     const amqpErrorHandler = (err) => {
       debug('amqp10 client error: ' + err.toString());
-      this._fsm.handle('amqpError', err, () => {
-        if (this._disconnectHandler) {
-          debug('calling upper layer disconnect handler');
-          this._disconnectHandler(err);
-        }
-      });
+      this._fsm.handle('amqpError', err);
     };
 
     this._fsm = new machina.Fsm({
@@ -106,10 +101,15 @@ export class Amqp {
       states: {
         disconnected: {
           _onEnter: (disconnectCallback, err, result) => {
-            if (err) {
-              this._safeCallback(disconnectCallback, err);
-            } else {
-              this._safeCallback(disconnectCallback, null, new results.Disconnected());
+            if (disconnectCallback) {
+              if (err) {
+                this._safeCallback(disconnectCallback, err);
+              } else {
+                this._safeCallback(disconnectCallback, null, new results.Disconnected());
+              }
+            } else if (this._disconnectHandler) {
+              debug('calling upper layer disconnect handler');
+              this._disconnectHandler(err);
             }
           },
           amqpError: (err, callback) => {
@@ -168,6 +168,12 @@ export class Amqp {
             };
 
             this._amqp.on(_amqpClientError, connectErrorHandler);
+            this._amqp.on('disconnected', () => {
+              debug('amqp10Client disconnected while state machine is: ' + this._fsm.state);
+              if (this._fsm.state === 'connected') {
+                this._fsm.transition('disconnected', undefined, new errors.NotConnectedError('amqp10: connection closed'));
+              }
+            });
             this._amqp.connect(this.uri)
               .then((result) => {
                 debug('AMQP transport connected.');
@@ -178,9 +184,8 @@ export class Amqp {
               })
               .catch((err) => {
                 this._amqp.removeListener(_amqpClientError, connectErrorHandler);
-                this._fsm.transition('disconnected');
                 /*Codes_SRS_NODE_COMMON_AMQP_16_003: [The `connect` method shall call the `done` callback if the connection fails.] */
-                this._safeCallback(connectCallback, connectError || err);
+                this._fsm.transition('disconnected', connectCallback, connectError || err);
               });
           },
           amqpError: (err, callback) => this._fsm.transition('disconnecting', callback, err),
@@ -410,9 +415,6 @@ export class Amqp {
    */
   setDisconnectHandler(disconnectCallback: GenericAmqpBaseCallback<any>): void {
     this._disconnectHandler = disconnectCallback;
-    // this._amqp.on('connection:closed', () => {
-    //   this._disconnectHandler(new errors.NotConnectedError('amqp10: connection closed'));
-    // });
   }
 
   /**
--- a/common/transport/amqp/test/_amqp_test.js
+++ b/common/transport/amqp/test/_amqp_test.js
@@ -91,13 +91,15 @@ describe('Amqp', function () {
   });
 
   describe('#setDisconnectHandler', function() {
-    it('disconnect callback is called when the \'connection:closed\' event is emitted', function(testCallback) {
+    it('disconnect callback is called when the \'disconnected\' event is emitted', function(testCallback) {
       var amqp = new Amqp();
+      sinon.stub(amqp._amqp, 'connect').resolves();
       amqp.setDisconnectHandler(function() {
         testCallback();
       });
-
-      amqp._amqp.emit('connection:closed');
+      amqp.connect('uri', null, function () {
+        amqp._amqp.emit('disconnected');
+      });
     });
   });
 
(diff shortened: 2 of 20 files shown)


common/transport/amqp/src/amqp.ts
```diff
@@ -14,7 +14,7 @@ import { AmqpLink } from './amqp_link_interface';
 
 import * as dbg from 'debug';
 import * as async from 'async';
-const debug = dbg('amqp-common:amqp');
+const debug = dbg('azure-iot-amqp-base:Amqp');
```
BertKleewein: >'azure-iot-amqp-base:Amqp [](start = 18, length = 25)

I'm glad you're doing this.  I've been using this facility and I appreciate the consistency.

common/transport/amqp/src/amqp.ts
```diff
@@ -90,6 +90,16 @@ export class Amqp {
       }
     }, amqp10.Policy.EventHub));
 
+    const amqpErrorHandler = (err) => {
+      debug('amqp10 client error: ' + err.toString());
+      this._fsm.handle('amqpError', err, () => {
+        if (this._disconnectHandler) {
+          debug('calling upper layer disconnect handler');
+          this._disconnectHandler(err);
```
BertKleewein: Not your problem, but this bothers me.  The mechanism of setting a handler seems correct, but the idea of a lower level object calling an upper level function smells like an inverted dependency.  OK, it's probably technically not one, but it still smells like it.

Would it be better to emit an event instead?  That feels more node-ish.  Is that more node-ish?  The current code feels like a C developer added this.

Just asking the question for my own education/rumination.

common/transport/amqp/src/amqp.ts
```diff
@@ -408,9 +410,9 @@ export class Amqp {
    */
   setDisconnectHandler(disconnectCallback: GenericAmqpBaseCallback<any>): void {
     this._disconnectHandler = disconnectCallback;
-    this._amqp.on('connection:closed', () => {
-      this._disconnectHandler(new Error('amqp10: connection closed'));
-    });
+    // this._amqp.on('connection:closed', () => {
```
BertKleewein: >this._amqp.on('connection:closed', () => { [](start = 7, length = 42)

I'm guessing this is one of the issues you plan to fix, but could you please comment why you commented out in case you don't get to it or it becomes irrelevant.  (code that is commented out without explanation always scares me because I wonder why it wasn't removed and what terrible secret it's hiding.)

common/transport/amqp/test/_amqp_test.js
```diff
@@ -574,7 +574,7 @@ describe('Amqp', function () {
           });
         });
 
-        it('calls the done callback with an error if the connection fails while trying to attach the link', function(testCallback) {
+        it.only('calls the done callback with an error if the connection fails while trying to attach the link', function(testCallback) {
```
BertKleewein: >it.only [](start = 8, length = 7)

left over from a debugging session?  I don't think you mean to disable all the other tests in the file.

device/core/devdoc/device_client_requirements.md
```diff
@@ -229,6 +229,16 @@ interface DeviceMethodEventHandler {
...
 
+#### setRetryPolicy(policy)
+
+**SRS_NODE_DEVICE_CLIENT_16_083: [** The `setRetryPolicy` method shall throw a `ReferenceError` if the policy object is falsy. **]**
+
+**SRS_NODE_DEVICE_CLIENT_16_084: [** The `setRetryPolicy` method shall throw an `ArgumentError` if the policy object doesn't have a `shouldRetry` method. **]**
+
+**SRS_NODE_DEVICE_CLIENT_16_085: [** The `setRetryPolicy` method shall throw an `ArgumentError` if the policy object doesn't have a `nextRetryTimeout` method. **]**
+
+**SRS_NODE_DEVICE_CLIENT_16_086: [** Any operation happening after a `setRetryPolicy` call should use the policy set during that call. **]**
```
BertKleewein: >operation [](start = 41, length = 9)

This is crazy vague.  What's an operation anyway?

device/core/src/client.ts
```diff
@@ -445,7 +525,10 @@ export class Client extends EventEmitter {
 
   private _enableC2D(callback: (err?: Error) => void): void {
     if (!this._c2dEnabled) {
-      this._transport.enableC2D((err) => {
+      const retryOp = new RetryOperation(this._retryPolicy, this.maxOperationTimeout);
+      retryOp.retry((opCallback) => {
+      this._transport.enableC2D(opCallback);
```
BertKleewein: >this [](start = 6, length = 4)

indenting also in enableMethods

device/core/src/client.ts
```diff
@@ -292,12 +344,13 @@ export class Client extends EventEmitter {
     /*Codes_SRS_NODE_DEVICE_CLIENT_16_042: [The `setOptions` method shall throw a `ReferenceError` if the options object is falsy.]*/
     if (!options) throw new ReferenceError('options cannot be falsy.');
 
-    this._transport.setOptions(options, (err) => {
+    const retryOp = new RetryOperation(this._retryPolicy, this.maxOperationTimeout);
```
BertKleewein: >RetryOperation [](start = 24, length = 14)

It's not intuitive to me why setting options would need to be a retryOp.  I assume it's because this might trigger a transport reconnection?  Or is there some other reason?  It might be useful to add a comment.

device/core/src/client.ts
```diff
@@ -35,6 +42,12 @@ export class Client extends EventEmitter {
    */
   static sasRenewalInterval: number = 2700000;
 
+  /**
+   * Maximum timeout (in milliseconds) used to consider an operation failed.
+   * The operation will be retried according to the retry policy set with {@link azure-iot-device.Client.setRetryPolicy} method (or {@link azure-iot-common.ExponentialBackoffWithJitter} by default) until this value is reached.)
+   */
+  maxOperationTimeout: number;
```
BertKleewein: >maxOperationTimeout [](start = 2, length = 19)

I don't see anywhere for customers to change this.  Are we adding an API later?

device/core/test/_client_test.js
```diff
@@ -864,6 +864,59 @@ describe('Client', function () {
...
+      var client = new Client(new EventEmitter());
+      var badPolicy = {
+        shouldRetry: function () {}
+      };
+      assert.throws(function() {
+        client.setRetryPolicy(badPolicy);
+      }, errors.ArgumentError);
+    });
+
+    /*Tests_SRS_NODE_DEVICE_CLIENT_16_086: [Any operation happening after a `setRetryPolicy` call should use the policy set during that call.]*/
```
BertKleewein: >Any operation [](start = 44, length = 13)

debt for later: the SRS says "any operation" but you're only testing sendEvent.

e2etests/test/d2c_disconnect.js
```diff
@@ -182,8 +184,8 @@ var runTests = function (hubConnectionString, provisionedDevice) {
                 .catch(testCallback);
             });
 
-      doConnectTest(false)('Device client sends ' + numberOfD2CMessages + ' messages, when event hub client receives first, it ' + testConfiguration.closeReason + 'which is not seen by the iot hub device client', function (testCallback) {
-        this.timeout(20000);
+      doConnectTest(testConfiguration.testEnabled)('Device client sends ' + numberOfD2CMessages + ' messages, when event hub client receives first, it ' + testConfiguration.closeReason + 'which is not seen by the iot hub device client', function (testCallback) {
```
BertKleewein: nit: add space

e2etests/test/method_disconnect.js
```diff
@@ -92,7 +93,7 @@ var protocolAndTermination = [
     delayInSeconds: 2
   },
   {
-    testEnabled: true,
+    testEnabled: false,
```
BertKleewein: >testEnabled: false [](start = 4, length = 18)

please comment why this is disabled, because this is supposed to be temporary because of a bug.  I assume a bunch of the tests are disabled because of protocol limitations, but it would be nice to know when that's not the case.

common/transport/amqp/src/amqp.ts
pierreca: you're right, and that's because historically we have this 'disconnect' error that was tacked on top of the normal error-bubbling pipeline that we have to keep carrying around while trying not to "double-bubble-up" errors. this needs cleaning up. right now (for this PR) i'm hesitant to touch something that kinda works though.

---
In reply to: [145961017](https://github.com/Azure/azure-iot-sdk-node/pull/114#discussion_r145961017) [](ancestors = 145961017)

device/core/src/client.ts
pierreca: i've been going back and forth on this. technically it's a public property. I think having a method that acts as a setter is overkill since it's just a number.  I'm also not sure other SDKs let customers customize this. so right now I'm going with a minimalist approach. I'm open to feedback though. 

---
In reply to: [145965088](https://github.com/Azure/azure-iot-sdk-node/pull/114#discussion_r145965088) [](ancestors = 145965088)

pierreca: maybe we should make it private for now. also I need to double-check what the default value really should be.

---
In reply to: [146043848](https://github.com/Azure/azure-iot-sdk-node/pull/114#discussion_r146043848) [](ancestors = 146043848,145965088)

device/core/test/_client_test.js
pierreca: yeah and I was thinking sendEvent just as an example but it's true we may want to test all operations.. hmmm.

---
In reply to: [145965518](https://github.com/Azure/azure-iot-sdk-node/pull/114#discussion_r145965518) [](ancestors = 145965518)

common/transport/amqp/test/_amqp_test.js
pierreca: correct, thanks for catching that.

---
In reply to: [145962077](https://github.com/Azure/azure-iot-sdk-node/pull/114#discussion_r145962077) [](ancestors = 145962077)

device/core/devdoc/device_client_requirements.md
pierreca: added examples. not sure how to make that very precise.

---
In reply to: [145962392](https://github.com/Azure/azure-iot-sdk-node/pull/114#discussion_r145962392) [](ancestors = 145962392)

common/transport/amqp/src/amqp.ts
pierreca: yes, I want to fix this before we ship - and it's related to the deactivated e2e tests I believe.

---
In reply to: [145961633](https://github.com/Azure/azure-iot-sdk-node/pull/114#discussion_r145961633) [](ancestors = 145961633)

e2etests/test/method_disconnect.js
pierreca: see comment about the connection:closed event

---
In reply to: [145968166](https://github.com/Azure/azure-iot-sdk-node/pull/114#discussion_r145968166) [](ancestors = 145968166)

device/core/test/_client_test.js
pierreca: I think this can wait. (famous last words, I know...)

---
In reply to: [146044146](https://github.com/Azure/azure-iot-sdk-node/pull/114#discussion_r146044146) [](ancestors = 146044146,145965518)

device/core/src/client.ts
pierreca: I'm going to make it private for now - official default value is 4 minutes. if we need to make it public later we'll find a way.

---
In reply to: [146043956](https://github.com/Azure/azure-iot-sdk-node/pull/114#discussion_r146043956) [](ancestors = 146043956,146043848,145965088)

device/core/src/client.ts
```diff
@@ -112,10 +128,36 @@ export class Client extends EventEmitter {
     }
 
     this._disconnectHandler = (err) => {
-      this.emit('disconnect', new results.Disconnected(err));
+      debug('transport disconnect event: ' + (err ? err.toString() : 'no error'));
+      if (err && this._retryPolicy.shouldRetry(err)) {
+        if (this._c2dEnabled) {
+          this._c2dEnabled = false;
```
anthonyvercolano: Don't have the flow very well in my head yet, but, if did have something like c2dEnabled, and you attempt to do a retry that fails, will you then lose the fact that c2d was originally enabled and on the next retry  you want to start it back up?

device/core/src/client.ts
```diff
@@ -469,6 +552,35 @@ export class Client extends EventEmitter {
...
+          this._methodsEnabled = true;
+        }
+        callback(err);
+      });
+    } else {
+      callback();
+    }
+  }
+
+  // private _disableMethods(callback: (err?: Error) => void): void {
```
anthonyvercolano: >// [](start = 2, length = 2)

?

e2etests/test/c2d_disconnect.js
```diff
@@ -87,7 +88,7 @@ var protocolAndTermination = [
     delayInSeconds: 2
   },
   {
-    testEnabled: true,
+    testEnabled: false,
```
anthonyvercolano: >false, [](start = 17, length = 6)

?

device/core/src/client.ts
pierreca: retries are handled within the _enableC2D call - so retry will work fine until we can't retry and timeout (and at that point, I think we want this to be false.

---
In reply to: [146082947](https://github.com/Azure/azure-iot-sdk-node/pull/114#discussion_r146082947) [](ancestors = 146082947)

e2etests/test/c2d_disconnect.js
pierreca: fixed in iteration 3

---
In reply to: [146330356](https://github.com/Azure/azure-iot-sdk-node/pull/114#discussion_r146330356) [](ancestors = 146330356)

device/core/src/client.ts
pierreca: added a clarifying comment in iteration 3

---
In reply to: [146323974](https://github.com/Azure/azure-iot-sdk-node/pull/114#discussion_r146323974) [](ancestors = 146323974)

PR #115, opened 2017-10-20
transport agnostic state machine for DPS

BertKleewein:
# Description of the problem]

While working on the MQTT protocol, I discovered that MQTT, HTTP, and, I presume, AMQP have a ton in common, both in terms of how the protocols operate, but also in terms of the state machine I need to use.  (For example, MQTT registration queries for operation status instead of using a C2D topic to send status updates.  In other words, we have to poll with MQTT just like we poll with HTTP.)

While writing the MQTT state machine, I discovered that I was copying a bunch of code from the HTTP state machine.  Rather than copy/paste, I decided to refactor the state machine out of the transport implementation.  TransportStateMachine is effectively an abstract base class for all the transports.  It has the state machine and all the transport-specific operations are left to the derived object to implement.


provisioning/device/devdoc/transport_state_machine_requirements.md
```diff
@@ -0,0 +1,81 @@
+# transport_state_machine Requirements
+
+## Overview
+This module provides HTTP protocol support to communicate with the Azure device provisioning service
```
pierreca: >HTTP [](start = 21, length = 4)

probably not only HTTP right?

provisioning/device/devdoc/transport_state_machine_requirements.md
```diff
@@ -0,0 +1,81 @@
...
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_001: [** The `constructor` shall accept no arguments **]**
+
+
+### connect(callback: (err?: Error) => void): void
+Connect is used to connect the transport
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_002: [** `connect` shall call `_doConnectForFsm`. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_003: [** If `_doConnectForFsm` fails, then `connect` shall fail. **]**
```
pierreca: those look like internal implementation details - unnecessary

provisioning/device/devdoc/transport_state_machine_requirements.md
```diff
@@ -0,0 +1,81 @@
...
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_004: [** If the transport is already connected, then `connect` shall do nothing and call the `callback` immediately. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_005: [** If `connect` is called while a connection is in progress, it shall wait for that connection to complete before calling `callback`. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_006: [** If `connect` is called while the transport is executing the first registration request, it shall do nothing and call `callback` immediately. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_007: [** If `connect` is called while the transport is waiting between operation status queries, it shall do nothing and call `callback` immediately. **]**
```
pierreca: not possible to disconnect and reconnect between operation status queries?

provisioning/device/devdoc/transport_state_machine_requirements.md
```diff
@@ -0,0 +1,81 @@
...
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_008: [** If `connect` is called while the transport is executing an operation status query, it shall do nothing and call `callback` immediately. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_009: [** If `connect` is called while the transport is disconnecting, it shall wait for the disconnection to complete, then initiate the connection. **]**
+
+
+### register(registrationId: string, authorization: string | X509, requestBody: any, forceRegistration: boolean, callback: Provisioning.ResponseCallback): void
+Register round-trips one step of the registration process, not returning until the operation is complete
```
pierreca: >operation [](start = 83, length = 9)

the registration or the initial request?

provisioning/device/devdoc/transport_state_machine_requirements.md
```diff
@@ -0,0 +1,81 @@
...
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_010: [** `register` shall `connect` the transport if it is not connected. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_011: [** `register` shall fail if the connection fails. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_012: [** `register` shall call `_doFirstRegistrationRequestForFsm`. **]**
```
pierreca: >_doFirstRegistrationRequestForFsm [](start = 83, length = 33)

internal - not in requirements.

provisioning/device/devdoc/transport_state_machine_requirements.md
```diff
@@ -0,0 +1,81 @@
...
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_013: [** If the registration request fails, `register` shall fail. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_014: [** If the registration request returns with status==assigned, it shall call `callback` with null, the response body, and the protocol-specific result. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_015: [** If the registration request returns with status==Assigning, it shall begin polling for operation status requests. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_016: [** If the registration request returns with an unknown status, `register` shall fail with a `SyntaxError` and pass the response body and the protocol-specific result to the `callback`. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_017: [** If the registration request returns a successful, `register` shall emit an 'operationStatus' event passing the body of the response. **]**
```
pierreca: >successful [](start = 98, length = 10)

missing a word?

provisioning/device/devdoc/transport_state_machine_requirements.md
```diff
@@ -0,0 +1,81 @@
...
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_019: [** If `_doOperationStatusQueryForFsm` fails, `register` shall fail. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_020: [** If `_doOperationStatusQueryForFsm` succeeds with status==Assigned, `register` shall complete and pass the body of the response and the protocol-spefic result to the `callback`. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_021: [** If `_doOperationStatusQueryForFsm` succeeds with status==Assigned, `register` shall begin another polling interval. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_022: [** If `_doOperationStatusQueryForFsm` succeeds with an unknown status, `register` shall fail with a `SyntaxError` and pass the response body and the protocol-specific result to the `callback`. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_023: [** If `_doOperationStatusQueryForFsm` succeeds, `register` shall emit an 'operationStatus' event passing the body of the response. **]**
+
```
pierreca: same here. no requirements on internal implementation details.

provisioning/device/devdoc/transport_state_machine_requirements.md
```diff
@@ -0,0 +1,81 @@
...
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_024: [** If `register` is called while a different request is in progress, it shall fail with an `InvalidOperationError`. **]**
```
pierreca: >InvalidOperationError [](start = 149, length = 21)

Hopefully the error message will be very clear here because the error name doesn't suggest what's going on.

provisioning/device/devdoc/transport_state_machine_requirements.md
```diff
@@ -0,0 +1,81 @@
...
+
+
+### disconnect(callback: (err: Error) => void): void
+Disconnect is used to disconnect the transport
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_025: [** If `disconnect` is called while disconnected, it shall immediately call its `callback`. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_026: [** `disconnect` shall call `_doDisconnectForFsm` of it's called while the transport is connected. **]**
```
pierreca: >_doDisconnectForFsm [](start = 85, length = 19)

same as before

provisioning/device/devdoc/transport_state_machine_requirements.md
```diff
@@ -0,0 +1,81 @@
...
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_006: [** If `connect` is called while the transport is executing the first registration request, it shall do nothing and call `callback` immediately. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_007: [** If `connect` is called while the transport is waiting between operation status queries, it shall do nothing and call `callback` immediately. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_008: [** If `connect` is called while the transport is executing an operation status query, it shall do nothing and call `callback` immediately. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_009: [** If `connect` is called while the transport is disconnecting, it shall wait for the disconnection to complete, then initiate the connection. **]**
+
+
+### register(registrationId: string, authorization: string | X509, requestBody: any, forceRegistration: boolean, callback: Provisioning.ResponseCallback): void
```
anthonyvercolano: > register(r [](start = 3, length = 11)

Perhaps should mention that this function will emit the assigning message?

provisioning/device/devdoc/transport_state_machine_requirements.md
```diff
@@ -0,0 +1,81 @@
...
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_011: [** `register` shall fail if the connection fails. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_012: [** `register` shall call `_doFirstRegistrationRequestForFsm`. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_013: [** If the registration request fails, `register` shall fail. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_014: [** If the registration request returns with status==assigned, it shall call `callback` with null, the response body, and the protocol-specific result. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_015: [** If the registration request returns with status==Assigning, it shall begin polling for operation status requests. **]**
```
anthonyvercolano: > returns with status [](start = 87, length = 20)

emits?

provisioning/device/devdoc/transport_state_machine_requirements.md
```diff
@@ -0,0 +1,81 @@
...
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_016: [** If the registration request returns with an unknown status, `register` shall fail with a `SyntaxError` and pass the response body and the protocol-specific result to the `callback`. **]**
```
anthonyvercolano: >`SyntaxError` [](start = 149, length = 13)

!

provisioning/device/devdoc/transport_state_machine_requirements.md
anthonyvercolano: >emit an 'operationStatus [](start = 127, length = 24)

Not the overall callack of the function?

provisioning/device/devdoc/transport_state_machine_requirements.md
```diff
@@ -0,0 +1,81 @@
...
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_017: [** If the registration request returns a successful, `register` shall emit an 'operationStatus' event passing the body of the response. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_018: [** When the polling interval elapses, `register` shall call `_doOperationStatusQueryForFsm`. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_019: [** If `_doOperationStatusQueryForFsm` fails, `register` shall fail. **]**
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_020: [** If `_doOperationStatusQueryForFsm` succeeds with status==Assigned, `register` shall complete and pass the body of the response and the protocol-spefic result to the `callback`. **]**
```
anthonyvercolano: > body of the response [](start = 165, length = 21)

So the body comes back in two places? The emit and the callback?

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,282 @@
...
+import * as Provisioning from './transport_interface';
+import * as dbg from 'debug';
+const debug = dbg('azure-device-provisioning:transport-fsm');
+
+export interface TransportHandlers {
+  _doConnectForFsm(callback: (err?: Error) => void): void;
+  _doDisconnectForFsm(callback: (err?: Error) => void): void;
+  _doFirstRegistrationRequestForFsm(registrationId: string, authorization: SharedAccessSignature | X509 | string, requestBody: any, forceRegistration: boolean, callback: (err?: Error, responseBody?: any, result?: any, pollingInterval?: number) => void): void;
+  _doOperationStatusQueryForFsm(registrationId: string, operationId: string, callback: (err?: Error, responseBody?: any, result?: any, pollingInterval?: number) => void): void;
+  _getErrorFromResultForFsm(result: any): any;
```
pierreca: That is an very weird interface. by convention anything prefixed by underscore is private in javascript. in typescript you don't need to do that if you use the `private` access modifier. not using it makes me think you want these to be public. so they shouldn't be prefixed. 
Also the `forFsm` suffix suggest strong coupling between the FSM and the transport. I'm not sure if there is such a strong coupling... and think we could dispense with the Suffix.

pierreca: Suggesting: 
```js
export interface TransportHandlers {
  connect(callback: (err?: Error) => void): void;
  disconnect(callback: (err?: Error) => void): void;
  registrationRequest(registrationId: string, authorization: SharedAccessSignature | X509 | string, requestBody: any, forceRegistration: boolean, callback: (err?: Error, responseBody?: any, result?: any, pollingInterval?: number) => void): void;
  queryRegistrationStatus(registrationId: string, operationId: string, callback: (err?: Error, responseBody?: any, result?: any, pollingInterval?: number) => void): void;
  getErrorResult(result: any): any;
}
```

---
In reply to: [146399275](https://github.com/Azure/azure-iot-sdk-node/pull/115#discussion_r146399275) [](ancestors = 146399275)

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,282 @@
...
+  private _fsm: machina.Fsm;
+  private _pollingTimer: any;
+  private _registrationCallback: Provisioning.ResponseCallback;
+
+  /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_001: [ The `constructor` shall accept no arguments ] */
+  constructor() {
+    super();
+
+    this._fsm = new machina.Fsm({
+      namespace: 'Provisioning',
```
pierreca: provisioning-transport :) we'll have a client, and security state machines probably too

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,282 @@
...
+      initialState: 'disconnected',
+      states: {
+        disconnected: {
+          _onEnter: (callback, err) => {
+            debug('entering disconnected state');
```
pierreca: >debug('entering disconnected state'); [](start = 12, length = 37)

use the on('transition', data) of the Fsm object instead.

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,282 @@
...
+            }
+          },
+          connect: (callback) => {
+            this._fsm.transition('connecting', callback);
+          },
+          register: (callback, registrationId, authorization, requestBody, forceRegistration) => {
+            /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_010: [ `register` shall `connect` the transport if it is not connected. ] */
+            this._fsm.handle('connect', (err) => {
+              if (err) {
+                this._fsm.transition('disconnecting', callback, err);
```
pierreca: >this._fsm.transition('disconnecting', callback, err); [](start = 16, length = 53)

connect itself should handle the transitions. these should only handle the call to the callback

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,282 @@
...
+
+export interface TransportHandlers {
+  _doConnectForFsm(callback: (err?: Error) => void): void;
+  _doDisconnectForFsm(callback: (err?: Error) => void): void;
+  _doFirstRegistrationRequestForFsm(registrationId: string, authorization: SharedAccessSignature | X509 | string, requestBody: any, forceRegistration: boolean, callback: (err?: Error, responseBody?: any, result?: any, pollingInterval?: number) => void): void;
+  _doOperationStatusQueryForFsm(registrationId: string, operationId: string, callback: (err?: Error, responseBody?: any, result?: any, pollingInterval?: number) => void): void;
+  _getErrorFromResultForFsm(result: any): any;
+}
+
+export class TransportStateMachine extends EventEmitter implements Provisioning.Transport, TransportHandlers{
```
pierreca: nit: missing space

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,282 @@
...
+    if (this._registrationInProgress()) {
+      debug('_cancelCurrentOperation: operation is in progress.  Cancelling.');
+      let _callback = this._registrationCallback;
+      this._registrationCallback = null;
+      _callback(new errors.OperationCancelledError());
+    }
+
+    callback();
+  }
+  private _handleOperationStatusResponse(registrationId: string,  responseBody: any, result: any, pollingInterval: number, callback: any): void {
```
pierreca: >private [](start = 2, length = 7)

nit: missing newline

provisioning/device/devdoc/transport_state_machine_requirements.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Public Interface
+
+### constructor()
+The constructor creates a TransportStateMachine object
+
+**SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_001: [** The `constructor` shall accept no arguments **]**
+
+
+### connect(callback: (err?: Error) => void): void
```
pierreca: >connect [](start = 4, length = 7)

is there any reason why we'd want to connect without registering? if ultimately register and getOperationStatus can handle connection themselves, no need to connect and then do nothing, right? so maybe it shouldn't be public?

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,282 @@
...
+          _onEnter: (callback, registrationId, authorization, requestBody, forceRegistration) => {
+            debug('entering sendingRegistrationRequest state');
+            this._registrationCallback = callback;
+            debug('calling transport _doFirstRegistrationRequestForFsm');
+            /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_012: [ `register` shall call `_doFirstRegistrationRequestForFsm`. ] */
+            this._doFirstRegistrationRequestForFsm(registrationId, authorization, requestBody, forceRegistration, (err, responseBody, result, pollingInterval) => {
+              if (this._registrationInProgress()) { // make sure we weren't cancelled before doing something with the response
+                if (err) {
+                  /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_013: [ If the registration request fails, `register` shall fail. ] */
+                  this._fsm.transition('connected', callback, err);
```
anthonyvercolano: >('connected', callback, err) [](start = 38, length = 28)

Not disconnect?

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,282 @@
...
+        sendingRegistrationReqest: {
+          _onEnter: (callback, registrationId, authorization, requestBody, forceRegistration) => {
+            debug('entering sendingRegistrationRequest state');
+            this._registrationCallback = callback;
+            debug('calling transport _doFirstRegistrationRequestForFsm');
+            /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_012: [ `register` shall call `_doFirstRegistrationRequestForFsm`. ] */
+            this._doFirstRegistrationRequestForFsm(registrationId, authorization, requestBody, forceRegistration, (err, responseBody, result, pollingInterval) => {
+              if (this._registrationInProgress()) { // make sure we weren't cancelled before doing something with the response
+                if (err) {
+                  /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_013: [ If the registration request fails, `register` shall fail. ] */
```
anthonyvercolano: > `register` shall fail [](start = 117, length = 22)

How does this percolate out?

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,282 @@
...
+          },
+        },
+        sendingRegistrationReqest: {
+          _onEnter: (callback, registrationId, authorization, requestBody, forceRegistration) => {
+            debug('entering sendingRegistrationRequest state');
+            this._registrationCallback = callback;
+            debug('calling transport _doFirstRegistrationRequestForFsm');
+            /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_012: [ `register` shall call `_doFirstRegistrationRequestForFsm`. ] */
+            this._doFirstRegistrationRequestForFsm(registrationId, authorization, requestBody, forceRegistration, (err, responseBody, result, pollingInterval) => {
+              if (this._registrationInProgress()) { // make sure we weren't cancelled before doing something with the response
```
pierreca: >_registrationInProgress [](start = 23, length = 23)

this looks like an inner state within a state machine. Suggest to handle the response with a state machine action, and modify our list of states handle the response based on the state of the state machine.

provisioning/device/src/transport_state_machine.ts
pierreca: guessing registration might fail without disconnecting the transport? this being said, the error might be a disconnection so maybe falling back to connected could be, in some situations, misleading?

---
In reply to: [146401341](https://github.com/Azure/azure-iot-sdk-node/pull/115#discussion_r146401341) [](ancestors = 146401341)

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,282 @@
...
+
+  disconnect(callback: (err: Error) => void): void {
+    debug('disconnect called');
+    this._fsm.handle('disconnect', callback);
+  }
+
+  private _cancelCurrentOperation(callback: (err?: Error) => void): void {
+    if (this._pollingTimer != null) {
+      debug('_cancelCurrentOperation: stopping polling timer');
+      clearTimeout(this._pollingTimer);
```
pierreca: >clearTimeout(this._pollingTimer); [](start = 6, length = 33)

Doesn't this belong in the state machine?

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,282 @@
...
+  private _handleOperationStatusResponse(registrationId: string,  responseBody: any, result: any, pollingInterval: number, callback: any): void {
+    if (responseBody) {
+      debug('received response from service:' + JSON.stringify(responseBody));
+      switch (responseBody.status.toLowerCase()) {
+        case 'assigned': {
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_017: [ If the registration request returns a successful, `register` shall emit an 'operationStatus' event passing the body of the response. ] */
+          this.emit('operationStatus', responseBody);
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_014: [ If the registration request returns with status==assigned, it shall call `callback` with null, the response body, and the protocol-specific result. ] */
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_020: [ If `_doOperationStatusQueryForFsm` succeeds with status==Assigned, `register` shall complete and pass the body of the response and the protocol-spefic result to the `callback`. ] */
+          this._fsm.transition('connected', callback, null, responseBody, result);
```
pierreca: >this._fsm.transition('connected' [](start = 10, length = 32)

calling transition from outside the state machine feels sketchy - you should be calling handle('something') and let the state machine handle it correctly depending on what state it's in.

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,282 @@
...
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_021: [ If `_doOperationStatusQueryForFsm` succeeds with status==Assigned, `register` shall begin another polling interval. ] */
+          this._fsm.transition('waiting_to_poll', callback, registrationId, responseBody.operationId, pollingInterval);
+          break;
+        }
+        default: {
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_016: [ If the registration request returns with an unknown status, `register` shall fail with a `SyntaxError` and pass the response body and the protocol-specific result to the `callback`. ] */
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_022: [ If `_doOperationStatusQueryForFsm` succeeds with an unknown status, `register` shall fail with a `SyntaxError` and pass the response body and the protocol-specific result to the `callback`. ] */
+          let err = new SyntaxError('status is ' + responseBody.status);
+          (err as any).Result = result;
```
pierreca: >Result [](start = 23, length = 6)

property names should start with a lowercase letter.

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,282 @@
...
+          (err as any).ResponseBody = responseBody;
+          this._fsm.transition('connected', callback, err);
+          break;
+        }
+      }
+    } else {
+      let err = this._getErrorFromResultForFsm(result);
+      (err as any).Result = result;
```
pierreca: >Result [](start = 19, length = 6)

same here (camelCase)

provisioning/e2e/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "azure-device-provisioning-integration",
```
pierreca: >integration [](start = 37, length = 11)

integration or e2e? folder name or package name?

provisioning/e2e/_transport_integration.js
```diff
@@ -0,0 +1,58 @@
...
+var registrationId = '__';
+var x509 = {};
+var registrationBody = {
+
+};
+
+var idScope = process.env.IOTHUB_PROVISIONING_INTEGRATION_IDSCOPE;
+var registrationId = process.env.IOTHUB_PROVISIONING_INTEGRATION_REGISTRATION_ID;
+var x509Key = fs.readFileSync(process.env.IOTHUB_PROVISIONING_INTEGRATION_X509_KEY, 'utf-8');
+var x509Cert = fs.readFileSync(process.env.IOTHUB_PROVISIONING_INTEGRATION_X509_CERT, 'utf-8');
```
pierreca: should dynamically provision those instead of relying of env variable unless absolutely necessary. the road to hell is paved with env variables.

provisioning/e2e/_transport_integration.js
```diff
@@ -0,0 +1,58 @@
...
+
+provisioningProtocols.forEach(function (Protocol) {
+
+  describe(Protocol.name, function() {
+    this.timeout(10000);
+
+    it ('can roundtrip x509', function(callback) {
+      var transport = new Protocol(transportConfig);
+      transport.on('operationStatus', function(response) {
+        console.dir(response);
```
pierreca: >console.dir(response); [](start = 8, length = 22)

can't have that in tests/build logs.

provisioning/device/src/transport_state_machine.ts
pierreca: >TransportStateMachine [](start = 13, length = 21)

This looks a lot like the beginning of the actual device client FSM.
 
A couple of things worry me:
1. this state machine doesn't have the appropriate abstraction level to deal with protocol specificities: MQTT errors, AMQP CBS links, etc. so it's likely you'll need another state machine in the lower transport layers

2. Looks like a case where inheritance and composition both work, but making it an inheritance thing makes it harder to test (which makes me lean in favor of composition). If we can get away with transports not having their own state machines, we could have separate "dumb" transport classes that implement the TransportHandlers contract and make this contract public. 

3. I don't know the details but polling seems inherently HTTP-ish. I would hope we don't have to poll in AMQP and MQTT and instead receive updates as messages over a link/topic. That would invalidate a lot of this FSM design.

provisioning/device/devdoc/transport_state_machine_requirements.md
BertKleewein: Absolutely.  The differences in authentication between TPM/x509 and MQTT/HTTP make it necessary to remove this as a public.

(TPM over HTTP has you pass a different authentication at registration depending on the state, but x509 over MQTT required you to pass authentication during the connect phase.)

I have a fix in a different branch,


---
In reply to: [146401297](https://github.com/Azure/azure-iot-sdk-node/pull/115#discussion_r146401297) [](ancestors = 146401297)

provisioning/device/devdoc/transport_state_machine_requirements.md
BertKleewein: Are you talking about references to all the _doBlahForFsm functions?  In that case, I disagree because that's part of the contract between this class and the transports that extend this class.

Or, are you talking specifically about the "if blah fails, then baz shall fail" part?  In that case, I can probably convince myself that you're right.

---
In reply to: [146396235](https://github.com/Azure/azure-iot-sdk-node/pull/115#discussion_r146396235) [](ancestors = 146396235)

provisioning/device/devdoc/transport_state_machine_requirements.md
BertKleewein: Nope.  status queries come every 2 seconds.  I don't want to open that hole.

---
In reply to: [146396612](https://github.com/Azure/azure-iot-sdk-node/pull/115#discussion_r146396612) [](ancestors = 146396612)

provisioning/e2e/_transport_integration.js
BertKleewein: Absolutely.  I just didn't want to take a dependency on the service SDK yet and this was the shortest path to get some validation going.

---
In reply to: [146405917](https://github.com/Azure/azure-iot-sdk-node/pull/115#discussion_r146405917) [](ancestors = 146405917)

provisioning/device/src/transport_state_machine.ts
BertKleewein: Agree on all counts.  Discussing in email.

---
In reply to: [146407891](https://github.com/Azure/azure-iot-sdk-node/pull/115#discussion_r146407891) [](ancestors = 146407891)

provisioning/device/src/transport_state_machine.ts
BertKleewein: These state machine have a strange way of returning failure.  The state you're in doesn't always match the failure you return.  For example, this transition says "move to the connected state, but also pass the error to it.  (ignore the fact that we might not want to stay connected for the moment).  Since the connected _onEnter function accepts a callback and an err, we can tacitly assume that the connect state will do something with this error.  In our case, it calls the callback because it's at the end of a chain of state change.  In other cases, it might pass it onto another state which will call the callback..

All these handlers are not type checked by TS at all.  I'm already getting spoiled because I miss that and it's not always easy to get the types correct.  

I'm going to make the return path more explicit by adding additional states for success and failure.

---
In reply to: [146401709](https://github.com/Azure/azure-iot-sdk-node/pull/115#discussion_r146401709) [](ancestors = 146401709)

provisioning/device/src/transport_state_machine.ts
BertKleewein: That's my thinking, but I'm still noodling on which state to go to in case of error.  Some errors (like a 404) keep the connection, but others don't.  There may be a giant list of failures-that-cause-disconnect in my future. :(

---
In reply to: [146403801](https://github.com/Azure/azure-iot-sdk-node/pull/115#discussion_r146403801) [](ancestors = 146403801,146401341)

provisioning/device/devdoc/transport_state_machine_requirements.md
BertKleewein: agreed. more explicit error objects (like invalid server response) coming in the future.

---
In reply to: [146398661](https://github.com/Azure/azure-iot-sdk-node/pull/115#discussion_r146398661) [](ancestors = 146398661)

provisioning/device/src/transport_state_machine.ts
BertKleewein: yes, and this only points to the need for a more comprehensive state transition review.

---
In reply to: [146400668](https://github.com/Azure/azure-iot-sdk-node/pull/115#discussion_r146400668) [](ancestors = 146400668)

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,239 @@
...
+          disconnect: (callback) => {
+            this._fsm.transition('disconnecting', callback);
+          },
+          register: (callback, registrationId, authorization, requestBody, forceRegistration) => {
+            this._fsm.transition('sendingregistrationRequest', callback, registrationId, authorization, requestBody, forceRegistration);
+          },
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_027: [ If a registration is in progress, `disconnect` shall cause that registration to fail with an `OperationCancelledError`. ] */
+          responseReceived: (callback) => callback(new errors.OperationCancelledError())
+        },
+        sendingRegistrationReqest: {
```
pierreca: >sendingRegistrationReqest [](start = 8, length = 25)

typo: sendingRegistrationRequest

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,239 @@
...
+            this._fsm.handle('connect', (err) => {
+              if (err) {
+                callback(err);
+              } else {
+                this._fsm.transition('sendingRegistrationReqest', callback, registrationId, authorization, requestBody, forceRegistration);
+              }
+            });
+          },
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_027: [ If a registration is in progress, `disconnect` shall cause that registration to fail with an `OperationCancelledError`. ] */
+          responseReceived: (callback) => callback(new errors.OperationCancelledError()),
```
pierreca: are we going to get there? receive a response after being disconnected?

pierreca: wondering if ew need a "pending operation queue" instead that you can flush and call the callbacks with an error

---
In reply to: [147223085](https://github.com/Azure/azure-iot-sdk-node/pull/115#discussion_r147223085) [](ancestors = 147223085)

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,239 @@
...
+              /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_019: [ If `TransportHandlers.queryOperationStatus` fails, `register` shall fail. ] */
+              this._fsm.transition('connected', callback, err);
+            } else if (responseBody) {
+              debug('received response from service:' + JSON.stringify(responseBody));
+              switch (responseBody.status.toLowerCase()) {
+                case 'assigned': {
+                  /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_014: [ If `TransportHandlers.registrationRequest` succeeds with status==Assigned, it shall emit an 'operationStatus' event and call `callback` with null, the response body, and the protocol-specific result. ] */
+                  /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_020: [ If `TransportHandlers.queryOperationStatus` succeeds with status==Assigned, `register` shall emit an 'operationStatus' event and complete and pass the body of the response and the protocol-spefic result to the `callback`. ] */
+                  this.emit('operationStatus', responseBody);
```
pierreca: >this.emit('operationStatus', responseBody); [](start = 18, length = 43)

emit is synchronous meaning you're going to get stuck in that state until the customer-defined callback finishes. I think you're better off transitioning back into connected (or a "registrationComplete" state, not sure), and emitting from there.

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,239 @@
...
+                  this._fsm.transition('connected', callback, null, responseBody, result);
+                  break;
+                }
+                case 'assigning': {
+                  /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_015: [ If `TransportHandlers.registrationRequest` succeeds with status==Assigning, it shall emit an 'operationStatus' event and begin polling for operation status requests. ] */
+                  /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_021: [ If `TransportHandlers.queryOperationStatus` succeeds with status==Assigning, `register` shall emit an 'operationStatus' event and begin polling for operation status requests. ] */
+                  this.emit('operationStatus', responseBody);
```
pierreca: >this.emit('operationStatus', responseBody); [](start = 18, length = 43)

same here

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,239 @@
...
+                }
+              }
+            } else {  // err == null && responseBody == null
+              let err = this._transport.getErrorResult(result);
+              (err as any).result = result;
+              (err as any).responseBody = responseBody;
+              this._fsm.transition('connected', callback, err, responseBody, result);
+            }
+          },
+          responseReceived: () => { throw new Error(); },
```
pierreca: >new Error() [](start = 42, length = 11)

I imagine we're considering it's a bug to receive another response if we're already analyzing one considering the timer hasn't been restarted yet?
in that case, I'm ok with throwing but i think we should make the error explicit so that customers can tell us "hey, i ran into that"

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,239 @@
...
+          _onEnter: (callback, registrationId, operationId, pollingInterval) => {
+            debug('waiting for ' + pollingInterval + ' ms');
+            this._pollingTimer = setTimeout(() => {
+              this._fsm.transition('polling', callback, registrationId, operationId, pollingInterval);
+            }, pollingInterval);
+            this._pollingTimer.callback = callback;
+          },
+          disconnect: (callback, err, responseBody) => {
+            /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_027: [ If a registration is in progress, `disconnect` shall cause that registration to fail with an `OperationCancelledError`. ] */
+            this._pollingTimer.callback(new errors.OperationCancelledError());
```
pierreca: >callback [](start = 31, length = 8)

same synchronous callback issue here - we're going to stay stuck in this state until the callback is executed, then transition to disconnecting. is this the desirable behavior? I feel like we should be transitioning, and once transitioned, call the necessary callbacks.

pierreca: one of the reason you want to be in the good state before calling anything is in case the callback triggers a retry.

---
In reply to: [147229585](https://github.com/Azure/azure-iot-sdk-node/pull/115#discussion_r147229585) [](ancestors = 147229585)

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,239 @@
...
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_008: [ If `connect` is called while the transport is executing an operation status query, it shall do nothing and call `callback` immediately. ] */
+          connect: (callback) => callback(),
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_024: [ If `register` is called while a different request is in progress, it shall fail with an `InvalidOperationError`. ] */
+          register: (callback) => callback(new errors.InvalidOperationError('another operation is in progress'))
+        },
+        disconnecting: {
+          _onEnter: (callback, err) => {
+            this._transport.disconnect((disconnectErr) => {
+              if (disconnectErr) {
+                debug('error received from transport during disconnection' + disconnectErr.toString());
```
pierreca: missing ': ' to separate the error from the text.

provisioning/device/package.json
```diff
@@ -22,15 +22,15 @@
     "@types/node": "^7.0.5"
   },
   "scripts": {
-    "lint": "tslint --type-check --project . -c ../../tslint.json",
```
anthonyvercolano: >--type-check [](start = 20, length = 12)

This will conflict with Pierres checkin.

provisioning/e2e/_transport_integration.js
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+
+var fs = require('fs');
+var assert = require('chai').assert;
+var debug = require('debug')('azure-device-provisioning:e2e');
```
pierreca: >:e2e [](start = 55, length = 4)

nit: -e2e (name of the package). the : separator is used within the package to separate files/classes.

provisioning/transport/http/src/http.ts
```diff
@@ -235,20 +111,54 @@ export class Http extends EventEmitter implements DeviceProvisioningTransport.Tr
...
 
-  private _registrationInProgress(): boolean {
-    return (this._registrationCallback != null);
+  getErrorResult(result: any): any {
+    return new Error();
   }
 
 }
 
+export class Http extends Provisioning.TransportStateMachine {
```
pierreca: >class [](start = 7, length = 5)

for later: will have to add docs comments.

provisioning/device/test/_transport_state_machine_test.js
```diff
@@ -0,0 +1,502 @@
...
+    } else {
+      assert(false, 'second status query returned Assigned.  There should not be a third status query');
+    }
+  });
+};
+
+describe('state machine', function () {
+  this.timeout(1000);
+
+  var makeNewMachine = function () {
```
pierreca: >makeNewMachine [](start = 6, length = 14)

why not use beforeEach() and afterEach() hooks to reinitialize everything between tests?

provisioning/device/test/_transport_state_machine_test.js
```diff
@@ -0,0 +1,502 @@
...
+    assert.isFalse(machine._transport.getErrorResult.called);
+  };
+
+  var callRegisterWithDefaultArgs = function (callback) {
+    machine.register(fakeRegistrationId, fakeAuth, fakeRequestBody, false, callback);
+  };
+
+  var machine;
+  beforeEach(function () {
+    machine = makeNewMachine();
```
pierreca: >machine [](start = 4, length = 7)

here you could have a transport object that you pass to makeNewMachine.
then you'd have a public transport object on which you can assert stuff without reaching into machine._transport.

provisioning/device/src/transport_state_machine.ts
pierreca: you can get help from the TS compiler by adding types in the signatures if you want.

---
In reply to: [147129572](https://github.com/Azure/azure-iot-sdk-node/pull/115#discussion_r147129572) [](ancestors = 147129572,146401709)

provisioning/device/devdoc/transport_state_machine_requirements.md
BertKleewein: I'm using errors.InvalidOperationError('another operation is in progress')

---
In reply to: [146397948](https://github.com/Azure/azure-iot-sdk-node/pull/115#discussion_r146397948) [](ancestors = 146397948)

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,253 @@
...
+          disconnect: (callback) => this._fsm.transition('disconnecting', callback),
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_006: [ If `connect` is called while the transport is executing the first registration request, it shall do nothing and call `callback` immediately. ] */
+          connect: (callback) => callback(),
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_024: [ If `register` is called while a different request is in progress, it shall fail with an `InvalidOperationError`. ] */
+          register: (callback) => callback(new errors.InvalidOperationError('another operation is in progress'))
+        },
+        response_received: {
+          _onEnter: (callback, err, registrationId, body, result, pollingInterval) => {
+            if (err) {
+              /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_013: [ If `TransportHandlers.registrationRequest` fails, `register` shall fail. ] */
```
pierreca: why not have this if(err) in _onEnter of the on sending_registration_request so that you transition directly into the right state?

provisioning/device/src/transport_state_machine.ts
BertKleewein: This shouldn't happen any more.  I've removed the reseponseReceived action and handle the cancel case directly in the callback from the transport.

---
In reply to: [147225350](https://github.com/Azure/azure-iot-sdk-node/pull/115#discussion_r147225350) [](ancestors = 147225350)

provisioning/device/src/transport_state_machine.ts
```diff
@@ -0,0 +1,253 @@
...
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_008: [ If `connect` is called while the transport is executing an operation status query, it shall do nothing and call `callback` immediately. ] */
+          connect: (callback) => callback(),
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_024: [ If `register` is called while a different request is in progress, it shall fail with an `InvalidOperationError`. ] */
+          register: (callback) => callback(new errors.InvalidOperationError('another operation is in progress'))
+        },
+        disconnecting: {
+          _onEnter: (callback, err, body, result) => {
+            /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_027: [ If a registration is in progress, `disconnect` shall cause that registration to fail with an `OperationCancelledError`. ] */
+            this._pendingOperations.popAllPendingOperations((callback) => {
+              callback(new errors.OperationCancelledError(''));
```
pierreca: >'' [](start = 58, length = 2)

What would be a good error message here? "Client is being disconnected" ?

PR #117, opened 2017-10-24
Add Keep-Alive default option for HTTPS connections

rafarhat:
# Checklist

# Description of the problem
Https connections are not using keep-alive by default. This means in scenarios where the service sdk is being used by other services / to performance many operations registry crud etc, the client will throw ETIMEDOUT and other client side errors as connections are made for every request.

# Description of the solution
Added keep-alive option for https requests by default.


pierreca:
Closing for a couple of reasons:
- merge conflicts
- our fault not getting to this in a timely manner
- keepalive seems to be a little more complicated of a subject than just turning on that boolean and i want to understand it more (see https://www.npmjs.com/package/agentkeepalive)
- it also relates to another issue open on HTTP for proxy support (ie we need a better way to pass HTTP options down from the client to the transport.

This is not being forgotten, it's just that as it is there's no point in keeping this open - we will get to that feature though.

PR #123, opened 2017-10-26
moving http_errors.ts to common in preparation to unify error mapping…

BertKleewein:
# Description of the problem

I'm tired of having at least 3 different pieces of code that map http status codes to iothub errors.  (one in device/transport/http, one in common/transport/http, and one in the twin code.  I don't want to add another one for provisioning, so I'm moving it to common/core.

This is the first commit, just moving the existing implementation.  I'll resolve the other 2 implementations (which are slightly different than the first) in my next commit.

### DIFF
--- a/common/core/common.d.ts
+++ b/common/core/common.d.ts
@@ -13,6 +13,7 @@ export { Message }
 export { SharedAccessSignature } from './lib/shared_access_signature';
 export { RetryOperation } from './lib/retry_operation';
 export { RetryPolicy, NoRetry, ExponentialBackOffWithJitter } from './lib/retry_policy';
+export { httpTranslateError } from './lib/http_errors';
 
 // Typescript only, used by other modules in azure-iot-sdk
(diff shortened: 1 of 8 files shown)


anthonyvercolano:
>       /*Codes_SRS_NODE_DEVICE_HTTP_ERRORS_16_005: [`translateError` shall return an `IotHubQuotaExceededError` if the HTTP response status code is `403`.]*/

I guess we won't be seeing this in the provisioning service?

---
Refers to: common/core/src/http_errors.ts:38 in 621fe1e. [](commit_id = 621fe1e3f87355a03df10ff7d7233779cff45c59, deletion_comment = False)

BertKleewein:
This is where it gets tricky.  My gut says that we want one handler that does as many status codes as possible.  If there are situation-specific cases where the base mapping is incorrect, then we can have a situation-specific translate error overrides the base for the overwritten mappings.

PR #125, opened 2017-10-27
remove external connect function from state machine

BertKleewein:
# Description of the problem
Having an explicit connect method for the provisioning state machine is unnecessary and causes problems with protocol-specific implementations.  

In particular, some protocols need authorization when connecting (like MQTT), others need the authorization when doing registration (like HTTP).  Since TPM-based registration requires multiple calls to register with different authorization strings, it is much easier to remove the connect method and support always passing the authorization to the register method.


provisioning/device/src/transport_state_machine.ts
```diff
@@ -45,12 +45,12 @@ export class TransportStateMachine extends EventEmitter implements Provisioning.
               callback(err, body, result);
             }
           },
-          connect: (callback) => {
-            this._fsm.transition('connecting', callback);
+          connect: (authorization, callback) => {
+            this._fsm.transition('connecting', callback, authorization);
```
anthonyvercolano: >callback, authorization [](start = 47, length = 23)

Odd parameter order.  callbacks usually last.

anthonyvercolano:
>               if (this._pendingOperations.operationIsStillPending(callback)) {

Not really sure how this works but, it seems that the identifying characteristic is the callback.  That doesn't seem to be discerning enough. 

---
Refers to: provisioning/device/src/transport_state_machine.ts:97 in 4c85dce. [](commit_id = 4c85dce87e0d02d41f7e900b8973b8b279db7859, deletion_comment = False)

provisioning/device/src/transport_state_machine.ts
pierreca: yes. callback last please 

---
In reply to: [147507291](https://github.com/Azure/azure-iot-sdk-node/pull/125#discussion_r147507291) [](ancestors = 147507291)

pierreca:
>             this._fsm.transition('sending_registration_request', callback, registrationId, authorization, requestBody, forceRegistration);

same here. callback last please.

---
Refers to: provisioning/device/src/transport_state_machine.ts:88 in 4c85dce. [](commit_id = 4c85dce87e0d02d41f7e900b8973b8b279db7859, deletion_comment = False)

pierreca:
>           _onEnter: (callback, registrationId, authorization, requestBody, forceRegistration) => {

move as last parameter

---
Refers to: provisioning/device/src/transport_state_machine.ts:92 in 4c85dce. [](commit_id = 4c85dce87e0d02d41f7e900b8973b8b279db7859, deletion_comment = False)

pierreca:
>           _onEnter: (callback, err, registrationId, body, result, pollingInterval) => {

same here

---
Refers to: provisioning/device/src/transport_state_machine.ts:107 in 4c85dce. [](commit_id = 4c85dce87e0d02d41f7e900b8973b8b279db7859, deletion_comment = False)

pierreca:
>               this._fsm.transition('response_error', callback, err);

and everywhere (not gonna add more comments about that after this one)

---
Refers to: provisioning/device/src/transport_state_machine.ts:111 in 4c85dce. [](commit_id = 4c85dce87e0d02d41f7e900b8973b8b279db7859, deletion_comment = False)

pierreca:
>         sending_registration_request: {

we generally use camelcasing instead of underscores

---
Refers to: provisioning/device/src/transport_state_machine.ts:91 in 4c85dce. [](commit_id = 4c85dce87e0d02d41f7e900b8973b8b279db7859, deletion_comment = False)

provisioning/device/src/transport_state_machine.ts
pierreca: actually, I'm realizing callback comes first in most of the state machines, that take an error or a result. so now I'm torn. let's talk.

PR #126, opened 2017-10-27
rename transports

BertKleewein:
# Description of the problem
provisioning transports were named incorrectly according to the visio file.  Easy fix now, much harder fix later.

### DIFF
--- a/provisioning/transport/amqp/package.json
+++ b/provisioning/transport/amqp/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "azure-iot-provisioning-amqp",
+  "name": "azure-device-provisioning-amqp",
   "version": "0.0.1",
   "description": "AMQP transport for Azure Device Provisioning SDK",
   "author": "Microsoft Corporation",
@@ -9,6 +9,7 @@
   "dependencies": {
     "azure-iot-amqp-base": "1.2.1",
     "azure-iot-common": "1.2.1",
+    "azure-device-provisioning-client": "0.0.1",
     "debug": "^3.0.1",
     "machina": "^2.0.0"
   },
--- a/provisioning/transport/http/package.json
+++ b/provisioning/transport/http/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "azure-iot-provisioning-http",
+  "name": "azure-device-provisioning-http",
   "version": "0.0.1",
   "description": "HTTP transport for Azure Device Provisioning SDK",
   "author": "Microsoft Corporation",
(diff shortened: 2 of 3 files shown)


provisioning/transport/amqp/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "azure-iot-provisioning-amqp",
+  "name": "azure-device-provisioning-amqp",
```
anthonyvercolano: >azure-device-provisioning-amqp [](start = 11, length = 30)

I think it should be azure-provisioning-device-client

provisioning/transport/amqp/package.json
```diff
@@ -9,6 +9,7 @@
   "dependencies": {
     "azure-iot-amqp-base": "1.2.1",
     "azure-iot-common": "1.2.1",
+    "azure-device-provisioning-client": "0.0.1",
```
anthonyvercolano: >azure-device-provisioning-client [](start = 5, length = 32)

azure-provisioning-device-client

provisioning/transport/http/package.json
```diff
@@ -9,6 +9,7 @@
   "dependencies": {
     "azure-iot-http-base": "1.2.0",
     "azure-iot-common": "1.2.1",
+    "azure-device-provisioning-client": "0.0.1",
```
anthonyvercolano: See comment for amqp.

provisioning/transport/mqtt/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "azure-iot-provisioning-mqtt",
+  "name": "azure-device-provisioning-mqtt",
```
anthonyvercolano: >ice-provisioning-mqtt", [](start = 20, length = 23)

see comment for http.

provisioning/transport/amqp/package.json
pierreca: really? "Azure provisioning" carries a very different meaning than "azure device provisioning". I'm really not sure about that one...

---
In reply to: [147511225](https://github.com/Azure/azure-iot-sdk-node/pull/126#discussion_r147511225) [](ancestors = 147511225)

provisioning/transport/amqp/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "azure-iot-provisioning-amqp",
+  "name": "azure-device-provisioning-amqp",
   "version": "0.0.1",
```
pierreca: Please do not check in without reviewing all dev-setup/teardown scripts

PR #127, opened 2017-10-29
[Service SDK] Create a state machine for the AMQP transport and add retry logic to the Client object

pierreca:
# Description of the problem
In order to enable retries at the client level, the AMQP transport must "understand" its state and manage its connection and link objects. Once this is done, the client can use the existing `RetryPolicy` object to retry operations on "retryable" errors.

# Description of the solution
- Introduce a state machine in the `Amqp` object
- Add a `setRetryPolicy` method on the `Client` object
- Add `RetryOperation` objects in all "retryable" operations.


service/src/amqp.ts
```diff
@@ -69,7 +80,331 @@ export class Amqp extends EventEmitter implements Client.Transport {
...
+                        finalError = translateError('error while detaching the file upload notification link when disconnecting', detachErr);
+                      }
+                      callback();
+                    });
+                  }
+                } else {
+                  callback();
+                }
+              },
+              (callback) => {
```
anthonyvercolano: >(callback) => { [](start = 14, length = 15)

Do you want to clear the put token renewal timer?

service/test/_amqp_test.js
```diff
@@ -73,59 +67,87 @@ describe('Amqp', function() {
...
-      sinon.stub(amqp._amqp,'connect').callsArgWith(2, testError);
+      fakeAmqpBase.connect = sinon.stub().callsArgWith(2, testError);
+      var amqp = new Amqp(fakeConfig, fakeAmqpBase);
       amqp.connect(function (err) {
         assert.instanceOf(err, Error);
         done();
       });
     });
 
+    /*Tests_SRS_NODE_IOTHUB_SERVICE_AMQP_16_019: [The `connect` method shall call the `connect` method of the base AMQP transport and translate its result to the caller into a transport-agnostic object.]*/
```
anthonyvercolano: >/*Tests_SRS_NODE_IOTHUB_SERVICE_AMQP_16_019: [](start = 4, length = 44)

Not really sure what the test is.

service/test/_amqp_test.js
```diff
@@ -134,69 +156,542 @@ describe('Amqp', function() {
...
+      // blocked in connecting because the connect callback won't be called.
+      amqp.disconnect(function () {
+        assert.isTrue(fakeAmqpBase.disconnect.calledOnce);
+        testCallback();
+      });
+    });
+
+    /*Tests_SRS_NODE_IOTHUB_SERVICE_AMQP_16_018: [All asynchronous instance methods shall call the `done` callback with either no arguments or a first null argument and a second argument that is the result of the operation if the operation succeeded.]*/
+    it('calls the base transport disconnect method if the transport is authenticating', function(testCallback) {
+      fakeAmqpBase.putToken = sinon.stub();
```
anthonyvercolano: >fakeAmqpBase.putToken = sinon.stub(); [](start = 6, length = 37)

It's unclear to me what this has to do with the test.

anthonyvercolano:
>   transportSpecificTests(opts);

Sleep much?  o7

---
Refers to: service/test/_client_test.js:395 in 1cf12b6. [](commit_id = 1cf12b6f5e5f5544247d131267716bef4791e2d9, deletion_comment = False)

service/src/amqp.ts
pierreca: this is done on the _onExit of the authenticated state

---
In reply to: [147610067](https://github.com/Azure/azure-iot-sdk-node/pull/127#discussion_r147610067) [](ancestors = 147610067)

pierreca: I have moved the call to setTimeout in the _onEnter too to make this more symmetrical

---
In reply to: [148421117](https://github.com/Azure/azure-iot-sdk-node/pull/127#discussion_r148421117) [](ancestors = 148421117,147610067)

service/test/_amqp_test.js
pierreca: ugh. thanks.

---
In reply to: [147610681](https://github.com/Azure/azure-iot-sdk-node/pull/127#discussion_r147610681) [](ancestors = 147610681)

service/test/_amqp_test.js
pierreca: I'm replacing the existing stub of putToken (existing stub calls its callback) with a stub() that does nothing. therefore it blocks the state machine in the "authenticating" state.

---
In reply to: [147616289](https://github.com/Azure/azure-iot-sdk-node/pull/127#discussion_r147616289) [](ancestors = 147616289)

common/core/src/retry_operation.ts
```diff
@@ -32,11 +32,11 @@ export class RetryOperation {
    * @param {(opCallback: (err?: Error, result?: any) => void) => void} operation The operation to execute.
    * @param {(err?: Error, result?: any) => void} finalCallback                   The callback to call with the final error or result, after retries if necessary.
    */
-  retry(operation: (opCallback: (err?: Error, result?: any) => void) => void, finalCallback: (err?: Error, result?: any) => void): void {
+  retry(operation: (opCallback: (err?: Error, result?: any, response?: any) => void) => void, finalCallback: (err?: Error, result?: any, response?: any) => void): void {
     const retryOperation = () => {
       this._retryCount++;
       /*Codes_SRS_NODE_COMMON_RETRY_OPERATION_16_001: [The `operation` function should be called at every retry.]*/
-      operation((err, result) => {
+      operation((err, result, response) => {
```
BertKleewein: >result [](start = 22, length = 6)

aargh.  this again.  For DPS, i've been using "result" for the protocol-specific object and "body" or "responseBody" for the body of the response.  I've also been putting body before result because it's more useful for consumers.  (result is more for diagnosing).  You were right earlier.  We do need to standardize on this and I'm hoping it's not too bad.

PR #129, opened 2017-10-30
remove operations queue

BertKleewein:
# Description of the problem
Previous state machine had a queue of pending operations.  There will never be more than one, so I changed it from a queue to a scalar value.


provisioning/device/src/transport_state_machine.ts
```diff
@@ -98,10 +97,10 @@ export class TransportStateMachine extends EventEmitter implements Provisioning.
...
           _onEnter: (callback, registrationId, authorization, requestBody, forceRegistration) => {
             /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_012: [ `register` shall call `TransportHandlers.registrationRequest`. ] */
-            this._pendingOperations.operationStarted(callback);
+            this._currentOperationCallback = callback;
             this._transport.registrationRequest(registrationId, authorization, requestBody, forceRegistration, (err, body, result, pollingInterval) => {
               // Check if the operation is still pending before transitioning.  We might be in a different state now and we don't want to mess that up.
-              if (this._pendingOperations.operationIsStillPending(callback)) {
+              if ( this._currentOperationCallback === callback) {
                 this._fsm.transition('response_received', callback, err, registrationId, body, result, pollingInterval);
               }
```
pierreca: should we log something if `this._currentOperationCallback !== callback` ?

Issue #73, opened 2017-08-20
MQTT "premature close"


sgacond:
# Context
- **OS and version used:** Mac OS X 10.12.5 / Raspbian Jessie Lite
- **Node.js version:** 6.10.3
- **npm version:** 3.10.10
- **list of installed packages:** 
```
├─┬ azure-iot-device@1.1.17
│ ├─┬ azure-iot-common@1.1.11
│ │ └── crypto@0.0.3
│ ├── azure-iot-http-base@1.1.11
│ ├─┬ azure-storage@2.3.0
│ │ ├── browserify-mime@1.2.9
│ │ ├── extend@1.2.1
│ │ ├─┬ json-edm-parser@0.1.2
│ │ │ └── jsonparse@1.2.0
│ │ ├─┬ md5.js@1.3.4
│ │ │ └── hash-base@3.0.4
│ │ ├─┬ readable-stream@2.0.6
│ │ │ ├── core-util-is@1.0.2
│ │ │ ├── isarray@1.0.0
│ │ │ ├── process-nextick-args@1.0.7
│ │ │ ├── string_decoder@0.10.31
│ │ │ └── util-deprecate@1.0.2
│ │ ├─┬ request@2.81.0
│ │ │ ├── aws-sign2@0.6.0
│ │ │ ├── aws4@1.6.0
│ │ │ ├── caseless@0.12.0
│ │ │ ├─┬ combined-stream@1.0.5
│ │ │ │ └── delayed-stream@1.0.0
│ │ │ ├── extend@3.0.1
│ │ │ ├── forever-agent@0.6.1
│ │ │ ├─┬ form-data@2.1.4
│ │ │ │ └── asynckit@0.4.0
│ │ │ ├─┬ har-validator@4.2.1
│ │ │ │ ├─┬ ajv@4.11.8
│ │ │ │ │ ├── co@4.6.0
│ │ │ │ │ └─┬ json-stable-stringify@1.0.1
│ │ │ │ │   └── jsonify@0.0.0
│ │ │ │ └── har-schema@1.0.5
│ │ │ ├─┬ hawk@3.1.3
│ │ │ │ ├── boom@2.10.1
│ │ │ │ ├── cryptiles@2.0.5
│ │ │ │ ├── hoek@2.16.3
│ │ │ │ └── sntp@1.0.9
│ │ │ ├─┬ http-signature@1.1.1
│ │ │ │ ├── assert-plus@0.2.0
[...]

PR #131, opened 2017-11-01
Fix for #73 (premature close error over MQTT)

pierreca:
#73 

# Description of the problem
There are 2 problems that are solved in this PR.
- #73 we weren't capturing socket errors that could be retried and were instead failing
- The `updateSharedAccessSignature`, if called while the socket was disconnected, would hang indefinitely

# Description of the solution
- Catch and convert the socket errors properly
- force disconnection of the socket, meaning it'll work even if the device is disconnected from the network.

### DIFF
--- a/common/transport/mqtt/devdoc/mqtt_translate_errors_requirements.md
+++ b/common/transport/mqtt/devdoc/mqtt_translate_errors_requirements.md
@@ -28,6 +28,6 @@
 
 **SRS_NODE_DEVICE_MQTT_ERRORS_18_011: [** `translateError` shall return an `Error` if none of the other string rules match **]**
 
-**SRS_NODE_DEVICE_MQTT_ERRORS_16_001: [** `translateError` shall return a `NotConnectedError` if the error object as a truthy `code` property (node socket errors) **]**
+**SRS_NODE_DEVICE_MQTT_ERRORS_16_001: [** `translateError` shall return a `NotConnectedError` if the error object has a truthy `code` property (node socket errors) **]**
 
 **SRS_NODE_DEVICE_MQTT_ERRORS_16_002: [** `translateError` shall return a `NotConnectedError` if the error message contains 'premature close' (from `end-of-stream`) **]**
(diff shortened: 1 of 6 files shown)


common/transport/mqtt/devdoc/mqtt_translate_errors_requirements.md
```diff
@@ -27,3 +27,7 @@
 **SRS_NODE_DEVICE_MQTT_ERRORS_18_010: [** `translateError` shall return a `InternalServerError` if the MQTT error message contains the string 'unrecognized packet type' **]**
 
 **SRS_NODE_DEVICE_MQTT_ERRORS_18_011: [** `translateError` shall return an `Error` if none of the other string rules match **]**
+
+**SRS_NODE_DEVICE_MQTT_ERRORS_16_001: [** `translateError` shall return a `NotConnectedError` if the error object as a truthy `code` property (node socket errors) **]**
```
BertKleewein: >as [](start = 114, length = 2)

has

common/transport/mqtt/src/mqtt_base.ts
```diff
@@ -108,7 +108,7 @@ export class MqttBase extends EventEmitter {
             /*Codes_SRS_NODE_COMMON_MQTT_BASE_16_033: [The `updateSharedAccessSignature` method shall disconnect and reconnect the mqtt client with the new `sharedAccessSignature`.]*/
             /*Codes_SRS_NODE_COMMON_MQTT_BASE_16_035: [The `updateSharedAccessSignature` method shall call the `callback` argument with no parameters if the operation succeeds.]*/
             /*Codes_SRS_NODE_COMMON_MQTT_BASE_16_036: [The `updateSharedAccessSignature` method shall call the `callback` argument with an `Error` if the operation fails.]*/
-            this._disconnectClient(false, () => {
+            this._disconnectClient(true, () => {
```
BertKleewein: >true, [](start = 35, length = 5)

There's a very weak connection between this change and the effect it has.  Is there a better way to do this?  "forceClose" always feels like a blunt tool, and I've mostly seen it used when shutting down or dealing with errors.  This seems like a non-error case that needs to be forced, but I'm not sure why.

pierreca: unfortunately we have I don't know of any way to detect if a network connection is up or not. I guess we could try to disconnect properly with a timeout and past that timeout, force the disconnect.. and tbh i'm not sure what the downside of forcing the disconnect is.

pierreca: Turns out there is a downside, which is to screw up the existing subscriptions, potentially messing features such as device methods after an update to the shared access signature has happened. I'm reverting this for now until we find a better solution. this problem is separate from the premature close error anyways...

PR #133, opened 2017-11-02
Parallelize build steps and some of the end-to-end tests

BertKleewein:
# Description of the solution
speed up builds using npm-run-all and some crude parallelism.
>*Why build a tree when an array is just as good and half as expensive*

This PR brings Jenkins builds down to 13 minutes.  I'm not sure we should accept it as is, but it has some interesting ideas.  It has 3 changes:
1. Minor fixes to various package.json files.  these are bugs, mostly with dependencies, that were masked by our reliance on dev_setup.
1. A script (build_parallel.js) that goes through our package.json files, keeps track of inter-module dependencies, and uses npm-build-all to do some *easy* build parallelization (explained below).  This saves us about 3-5 minutes from dev-setup.cmd, dev-teardown.cmd, and build.cmd.  As a bonus, the only input is a list of packages to build.  Everything else is pulled from their package.json files, so no manually editing of build scripts.
1. A quick hack to E2E tests to do some more easy parallelization.  I left the huge loop in e2e_tests.js alone, but I peeled off the 5 groups at the bottom so I could run in parallel as a POC.  This saves us ~6+ minutes.  e2e_tests.js is still the long pole at 10 minutes, but could be refactored to introduce more parallelism without *too* much work.

## build_parallel.js
This script reads package.json files and decides how to run scripts in parallel.  It does this by building it's own package.json (which only exists to runs npm on other package.json files), and intelligently running these new npm scripts in parallel.

build_parallel works like this:  it goes through all of our package.json files.  Any files that have their dependencies satisfied get added to a "build now" list.  Once it's done looping through, it builds everything in the "build now" list in parallel and marks them as done.  It then goes through the list again to create a new "build now" list, only this time more dependencies are satisfied, and more things are buildable.  It does this until everything is built.  

(for what it's worth, it doesn't actually build while it's looping through the list.  It actually just keeps track of what it would build and executes it later using async.js and npm-run-all.js)

So, for the dev-setup case, it builds a package.json that looks like this:
```
{
  "scripts": {
    "setup_azure-iot-common": "cd F:\\repos\\node\\common\\core  && npm ln && npm run build",
    "setup_azure-iot-mqtt-base": "cd F:\\repos\\node\\common\\transport\\mqtt  && npm ln azure-iot-common && npm ln && npm run build",
    "setup_azure-iot-http-base": "cd F:\\repos\\node\\common\\transport\\http  && npm ln azure-iot-common && npm ln && npm run build",
```

In the first phase, build_parallel.js looks through all the package.json files.  The only package it can build is azure-iot-common.  In the second phase, it can build anything that depends on azure-iot-common (e.g. mqtt-base, http-base, amqp-base, etc).  It builds all of these in parallel.  It keeps going until everything is built.

The parallelism that we accomplish looks like this:
```
call npm-run-all -p setup__azure-iot-common
call npm-run-all -p setup__azure-iot-mqtt-base setup__azure-iot-http-base setup__azure-iot-amqp-base setup__azure-device-provisioning-client
call npm-run-all -p setup__azure-iot-provisioning-http setup__azure-iot-device setup__azure-iothub
call npm-run-all -p setup__azure-iot-device-mqtt setup__azure-iot-device-http setup__azure-iot-device-amqp
```


build/dev-setup.cmd
```diff
@@ -8,136 +8,16 @@ set node-root=%~dp0..
...
+cd %node-root%\build\tools
 echo.
-echo -- Creating links for %cd% --
-call npm link azure-iot-http-base
-call npm link azure-iot-common
-call npm link azure-iot-provisioning-device
-call npm link
-call npm run build
+echo -- Setting up links for build tools --
+call npm install
```
pierreca: >install [](start = 9, length = 7)

this will fail during the release - the reason is that the create_device.js script relies on azure-iothub and during the release the version number of the dependency is bumped before it is published.

build/dev-setup.sh
```diff
@@ -5,135 +5,13 @@
...
-echo "\n-- Creating links for `pwd` --"
-npm link azure-iot-common
-npm link azure-iot-device
-npm link azure-iot-device-amqp
-npm link azure-iot-device-http
-npm link azure-iot-device-mqtt
-npm link azure-iothub
+cd $node_root/build/tools
+echo "\n-- Setting up links for build tools --"
 npm install
```
pierreca: >install [](start = 4, length = 7)

same comment as in dev-setup.cmd

build/tools/build_parallel.js
```diff
@@ -0,0 +1,203 @@
...
+
+var usage = function() {
+  console.log('usage: node ' + __filename + ' task');
+  console.log('where task is one of:');
+  var tasks = require('./projects.json').tasks;
+  Object.keys(tasks).forEach(function(taskName) {
+    var task = tasks[taskName];
+    console.log("  " + taskName + " - " + task.description);
+  });
+;}
```
pierreca: };

build/tools/projects.json
```diff
@@ -0,0 +1,74 @@
+{
+  "tasks": {
+    "setup": {
+      "description": "install and build all project dependencies",
+      "command": "cd {fullpath} {perDependencyCommand} && npm ln && npm run build",
+      "perDependencyCommand": " && npm ln {dependency}"
+    },
+    "teardown": {
+      "description": "un-install all project dependencies",
```
pierreca: >un-install [](start = 22, length = 10)

nit: I believe "uninstall" without a hyphen is ok. I'd also add "deregister the package from the cache"

build/tools/projects.json
```diff
@@ -0,0 +1,74 @@
+{
+  "tasks": {
+    "setup": {
+      "description": "install and build all project dependencies",
```
pierreca: >install [](start = 22, length = 7)

install, registers the package with the cache, and build all...

e2etests/test/authentication.js
```diff
@@ -181,4 +182,6 @@ module.exports = function authentication_tests(hubConnectionString) {
       });
     });
   });
-};
\ No newline at end of file
+};
+
+runTests();
```
pierreca: >runTests [](start = 0, length = 8)

I'd suggest just removing the function alltogether and just call describe directly

service/package.json
```diff
@@ -7,6 +7,7 @@
   "main": "iothub.js",
   "typings": "iothub.d.ts",
   "dependencies": {
+    "amqp10": "3.5.3",
```
pierreca: >"amqp10": "3.5.3", [](start = 4, length = 18)

?

e2etests/test/authentication.js
pierreca: (same for all scripts)

---
In reply to: [148617265](https://github.com/Azure/azure-iot-sdk-node/pull/133#discussion_r148617265) [](ancestors = 148617265)

build/tools/build_parallel.js
```diff
@@ -0,0 +1,203 @@
...
+var async = require('async');
+var EventEmitter = require('events');
+
+// See section titled "About MaxListenersExceededWarning" in https://github.com/mysticatea/npm-run-all/blob/HEAD/docs/node-api.md
+EventEmitter.defaultMaxListeners = 30;
+
+var usage = function() {
+  console.log('usage: node ' + __filename + ' task');
+  console.log('where task is one of:');
+  var tasks = require('./projects.json').tasks;
```
pierreca: >var tasks = require('./projects.json').tasks; [](start = 2, length = 45)

that is going to make a call to usage() block while the file is loaded - not that big of a deal but why not require it at the top with all the other require statements?

build/tools/build_parallel.js
```diff
@@ -0,0 +1,203 @@
...
+    var packageJson = require(path.join(project.fullpath, 'package.json'));
+    project.dependencies = packageJson.dependencies;
+    project.name = packageJson.name;
+    projects[project.name] = project;
+  });
+  return projects;
+};
+var projects = readProjectFile();
+
+// helper to see if we run the current task for a given project`
```
pierreca: >` [](start = 63, length = 1)

nit: extraneous backquote

build/tools/build_parallel.js
```diff
@@ -0,0 +1,203 @@
...
+var needToRunTask = function(project) {
+  return (!project.hasOwnProperty(taskName) || !!project[taskName]);
```
pierreca: >return (!project.hasOwnProperty(taskName) || !!project[taskName]); [](start = 2, length = 66)

had to think for a bit and look at the JSON file to understand what this was doing. I guess: run the task by default and if the task has such a property, then coerce that property value to boolean and use that? 

it's very easy for scripts to get very cryptic - I think at least a comment would help. ideally though we'd have a README file going through the design and usage for the tool.

Making our builds faster is one noble goal but the end game is for customers to be able to use this as much as we are.

build/tools/projects.json
```diff
@@ -0,0 +1,74 @@
+{
+  "tasks": {
```
pierreca: >tasks [](start = 3, length = 5)

file is named projects but contains both projects and tasks?

build/tools/build_parallel.js
```diff
@@ -0,0 +1,203 @@
...
+  process.exit(1);
+}
+
+console.log('dirname is ' + __dirname);
+var gitRoot = findRoot(__dirname, function (dir) {
+  return fs.existsSync(path.resolve(dir, '.gitattributes'));
+});
+
+// ingest projects.json
+var readProjectFile = function() {
```
pierreca: >readProjectFile [](start = 4, length = 15)

nit: readProjectsFromConfigFile? since you also have tasks in there and you're not returning those

pierreca: and you're doing a little more than reading the file here - you're actually building an object from each package.json for the listed projects in the project file

---
In reply to: [148622571](https://github.com/Azure/azure-iot-sdk-node/pull/133#discussion_r148622571) [](ancestors = 148622571)

build/tools/build_parallel.js
```diff
@@ -0,0 +1,203 @@
...
+  try {
+    fs.statSync(tempDir).isDirectory();
+  } catch (err) {
+    fs.mkdirSync(tempDir);
+  }
+
+  var packageJson = { 'scripts' : {} };
+
+  Object.keys(projects).forEach(function (projectName) {
+    var project = projects[projectName];
```
pierreca: took me a while to understand what is going on here. please consider 
- documenting/typing the data structures you're using
- revising naming (soo many projectthings in those 2 lines!)
and in this specific case - why not use
```js
for (var project in projects) {
  //...
}
```

build/tools/build_parallel.js
```diff
@@ -0,0 +1,203 @@
...
+            dependencyCommand += task.perDependencyCommand.replace(/{dependency}/g, dependencyName);
+          }
+        });
+        command = command.replace(/{perDependencyCommand}/g, dependencyCommand);
+      }
+      packageJson.scripts[taskName + '_' + project.name] = command;
+    }
+  });
+  packageJsonFilename = path.join(tempDir, 'package.json');
+  fs.writeFileSync(packageJsonFilename, JSON.stringify(packageJson,null,'  '));
```
pierreca: this whole thing is elegant but extremely hard to understand. I'm afraid maintaining it might become a headache for anyone that is not you.
Please consider at least adding comments, and maybe splitting it in functions that are named after each of the steps you're taking, and also try to have less repetitive names for variables.

build/tools/build_parallel.js
```diff
@@ -0,0 +1,203 @@
...
+};
+makePackageJson();
+
+// helper to know if all projects have been built
+var allProjectsCompleted = function() {
+  var allDone = true;
+  Object.keys(projects).forEach(function (projectName) {
+    var project = projects[projectName];
```
pierreca: same here - why not for .. in ..?

PR #134, opened 2017-11-03
Remove device_provision.js dependency in e2e tests in preparation to parallelize them

pierreca:
# Description of the problem
E2E tests right cannot be parallelized because a lot of them depend on a single test device provisioning script that runs before everything else. we'd like to be able to parallelize e2e tests to shorten the build time

# Description of the solution
Remove the device_provision and device_teardown scripts and have each test suite create its own test devices.


e2etests/test/device_service.js
```diff
@@ -13,13 +13,55 @@ var createDeviceClient = require('./testUtils.js').createDeviceClient;
...
+    device_service_tests(deviceTransport, createDeviceMethod);
+  });
+});
+
+[
+  deviceAmqp.Amqp,
+  deviceMqtt.Mqtt
+].forEach(function (deviceTransport) {
+  device_service_tests(deviceTransport, DeviceIdentityHelper.createDeviceWithX509CASignedCert);
+});
```
anthonyvercolano: kinda looks like you're doing this twice.

e2etests/test/device_acknowledge_tests.js
```diff
@@ -6,16 +6,50 @@
...
+[
+  DeviceIdentityHelper.createDeviceWithSas,
+  DeviceIdentityHelper.createDeviceWithSymmetricKey,
+  DeviceIdentityHelper.createDeviceWithX509SelfSignedCert
+].forEach(function (createDeviceMethod) {
+  [
+    deviceHttp.Http,
+    deviceAmqp.Amqp,
+    deviceAmqp.AmqpWs
+  ].forEach(function (deviceTransport) {
```
BertKleewein: >forEach [](start = 4, length = 7)

Creating the device before calling describe could slow the tests down because it forces us to pass --delay to mocha.  I don't know if the delay is significant or not

e2etests/test/sas_token_tests.js
```diff
@@ -12,9 +12,32 @@ var Client = require('azure-iot-device').Client;
...
 var SharedAccessSignature = require('azure-iot-device').SharedAccessSignature;
 var ConnectionString = require('azure-iot-device').ConnectionString;
+var DeviceIdentityHelper = require('./device_identity_helper.js');
+
+var hubConnectionString = process.env.IOTHUB_CONNECTION_STRING;
+var transports  = [
+  require('azure-iot-device-amqp').Amqp,
+  require('azure-iot-device-amqp').AmqpWs,
+  require('azure-iot-device-mqtt').Mqtt,
+  require('azure-iot-device-mqtt').MqttWs,
```
BertKleewein: insignificant, but does this load and parse the MQTT module twice?

e2etests/test/device_service.js
pierreca: it's a weird matrix.
in the first nested loops i do it for 3 authentication methods over 4 protocols, then I do it only over MQTT and AMQP for the CA signed cert authentication

---
In reply to: [149209475](https://github.com/Azure/azure-iot-sdk-node/pull/134#discussion_r149209475) [](ancestors = 149209475)

e2etests/test/d2c_disconnect.js
```diff
@@ -102,181 +105,189 @@ var protocolAndTermination = [
...
+    var deviceClient, ehClient, senderInterval;
+    var provisionedDevice;
 
-      beforeEach(function () {
-        this.timeout(20000);
-        ehClient = eventHubClient.fromConnectionString(hubConnectionString);
-        deviceClient = createDeviceClient(testConfiguration.transport, provisionedDevice);
-        senderInterval = null;
+    before(function (beforeCallback) {
+      DeviceIdentityHelper.createDeviceWithSas(function (err, testDeviceInfo) {
```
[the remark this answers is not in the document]
pierreca: not trimming it down, this is the same as before. 
our auth vs feature test coverage is a little weird for sure and would benefit from more clarity - but if anything at that point i think the existing matrix makes us run too many tests... we'll have to rethink this.

---
In reply to: [149447999](https://github.com/Azure/azure-iot-sdk-node/pull/134#discussion_r149447999) [](ancestors = 149447999)

e2etests/test/sas_token_tests.js
pierreca: no. require() caches the module once it's loaded. 

---
In reply to: [149450212](https://github.com/Azure/azure-iot-sdk-node/pull/134#discussion_r149450212) [](ancestors = 149450212)

PR #135, opened 2017-11-03
ProvisioningDeviceClient object

BertKleewein:
This is an initial implementation of ProvisioningDeviceClient along with an E2E test that shows it working with an X509 cert.

There are a few things that we need to address:
1. I added a static createDeviceClient function to the transport object.  This seems cleaner than creating a transport object then passing it into ProvisioningDeviceClient.create.
1. ProvisioningDeviceClient.register takes an cert, token, or authentication string.  We want to change this to get the credentials from the security/authentication/whatever object.
1. Parameters for create/register need to be revisited.  The doc had create taking a URI, which isn't as useful as an idScope.  In this implementation, I have the URI as a member variable along with other config values.  Do we want to keep these as members, or do we want a different mechanism to set options?


provisioning/device/devdoc/client_state_machine_requirements.md
```diff
@@ -1,7 +1,7 @@
-# transport_state_machine Requirements
+cd ..# client_state_machine Requirements
```
pierreca: >cd .. [](start = 0, length = 5)

woops! typo!

provisioning/device/devdoc/client_state_machine_requirements.md
```diff
@@ -46,12 +38,12 @@ Register round-trips one step of the registration process, not returning until a
 **SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_024: [** If `register` is called while a different request is in progress, it shall fail with an `InvalidOperationError`. **]**
 
 
-### disconnect(callback: (err: Error) => void): void
-Disconnect is used to disconnect the transport
+### endSession(callback: (err: Error) => void): void
```
pierreca: >endSession [](start = 4, length = 10)

during other SDKs reviews we talked about this and I believe we ended up deciding that we should automatically "end the session" when we get to the assigned state.

as to how to provide a way to "end the session" before it's complete, TBD. 

in any case, we should not rely on the customer to call this to clean things up when registration is done.

provisioning/device/src/interfaces.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * default timeout to use when communicating with the service
+   */
+  timeoutInterval: number;
+
+  /**
+   * idScope to use when communicating with the provisioning service
+   */
+  idScope: string;
```
pierreca: >idScope [](start = 2, length = 7)

public API nitpicking: I believe other SDKs use scopeId.

provisioning/e2e/package.json
```diff
@@ -10,13 +10,17 @@
     "azure-iot-provisioning-device-amqp": "0.0.1",
     "azure-iot-provisioning-device-http": "0.0.1",
     "azure-iot-provisioning-device-mqtt": "0.0.1",
-    "debug": "^3.0.1",
-    "machina": "^2.0.0",
+    "azure-iot-provisioning-service": "0.0.1",
+    "azure-iothub": "^1.1.18",
```
pierreca: >1.1.18 [](start = 22, length = 6)

we're now on 1.2.0 - also please make sure you add this to the input.js file in the release repository.

provisioning/e2e/_provisioning_e2e.js
```diff
@@ -0,0 +1,145 @@
...
+var registryConnectionString = process.env.IOTHUB_CONNECTION_STRING;
+
+var provisioningTransports = [ Http ];
+
+describe('IoT Provisioning', function() {
+  var deviceId;
+  var registrationId;
+  this.timeout(30000);
+
+  var provisioningServiceClient = Promise.promisifyAll(ProvisioningServiceClient.fromConnectionString(provisioningConnectionString));
```
pierreca: >promisifyAll [](start = 42, length = 12)

I see the benefit but I'm worried we're adding one layer between the tests and the SDK that might alter the test results ?

provisioning/e2e/_provisioning_e2e.js
```diff
@@ -0,0 +1,145 @@
...
+  beforeEach (function() {
+    var id = uuid.v4();
+    deviceId = 'deleteme_provisioning_node_e2e_' + id;
+    registrationId = 'reg-' + id;
+  });
+
+  afterEach (function(callback) {
+    debug('deleting enrollment');
+    provisioningServiceClient.deleteIndividualEnrollmentAsync(registrationId)
+      .catch(function() {
```
pierreca: >catch [](start = 7, length = 5)

you should be able to use only one catch statement at the end after all the then() operations - this being said, not sure if promises serve our purpose here.

provisioning/e2e/_provisioning_e2e.js
```diff
@@ -0,0 +1,145 @@
...
+      .then(function() {
+        debug('verifying registration status is unassigned');
+        return assertRegistrationStatus(registrationId, 'unassigned');
+      })
+      .then(function() {
+        debug('registering device');
+        var provisioningDeviceClient = Promise.promisifyAll(Transport.createDeviceClient(idScope));
+        return provisioningDeviceClient.registerAsync(registrationId, x509cert, false)
+          .then(function() {
+            return provisioningDeviceClient.endSessionAsync();
```
pierreca: >provisioningDeviceClient.endSessionAsync(); [](start = 19, length = 43)

I'd like it better if we didn't have to force the customer to call this and ended our session cleanly ourselves.

provisioning/device/src/interfaces.ts
```diff
@@ -0,0 +1,72 @@
...
+export interface DeviceConfiguration {
+  iotHubUri: string;
+  deviceId: string;
+}
+
+/**
+ * Interface that all provisioning transports must implement in order to support the provisioning service
+ */
+export interface TransportHandlers {
+  setClientConfig(config: ClientConfiguration): void;
```
pierreca: >setClientConfig [](start = 2, length = 15)

I'm not clear as to why we need this?

provisioning/device/src/client.ts
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * idScope to use when communicating with the provisioning service
+   */
+  idScope: string;
+
+  private _stateMachine: ClientStateMachine;
+
+  private constructor(transport: Provisioning.TransportHandlers, idScope: string) {
+    this.idScope = idScope;
+    transport.setClientConfig(this);
```
pierreca: >setClientConfig [](start = 14, length = 15)

hmm. so the transport constructor takes a config object and we let the user create the transport but then we overwrite it?

provisioning/device/src/client.ts
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+import * as Provisioning from './interfaces';
+import { ClientStateMachine } from './client_state_machine';
+import * as dbg from 'debug';
+const debug = dbg('azure-device-provisioning:device-client');
+
+export class ProvisioningDeviceClient implements Provisioning.ClientConfiguration {
```
pierreca: >ClientConfiguration [](start = 62, length = 19)

single responsibility principle question here: it's both the client and the config object? feels like those two should be separate.

provisioning/device/src/client.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * User-Agent string passed to the service as part of communication
+   */
+  userAgent: string = 'azure-iot-node-provisioning'; // TODO
```
anthonyvercolano: >string = 'azure-iot-node-provisioning'; // TODO [](start = 13, length = 47)

const packageJson = require('../package.json');

eg: for http: RestApiClient(config, packageJson.name + '/' + packageJson.version)

anthonyvercolano:
>               if (this._currentOperationCallback === callback) {

Consider if this state machine is invoked with no callback.

---
Refers to: provisioning/device/src/client_state_machine.ts:152 in d935c82. [](commit_id = d935c82bacb60a5a31a1c04f9d824c737c5d6594, deletion_comment = False)

anthonyvercolano:
>         waitingToPoll: {

Question the utility of this state.

---
Refers to: provisioning/device/src/client_state_machine.ts:132 in d935c82. [](commit_id = d935c82bacb60a5a31a1c04f9d824c737c5d6594, deletion_comment = False)

provisioning/device/devdoc/client_state_machine_requirements.md
BertKleewein: OK.  We may still want to be explicit at this level though.

---
In reply to: [148916208](https://github.com/Azure/azure-iot-sdk-node/pull/135#discussion_r148916208) [](ancestors = 148916208)

provisioning/e2e/_provisioning_e2e.js
BertKleewein: yeah, I go back and forth.  I was using this as an excuse to learn promises.  

---
In reply to: [148917838](https://github.com/Azure/azure-iot-sdk-node/pull/135#discussion_r148917838) [](ancestors = 148917838)

BertKleewein: for now, I'm going to leave it.  I'll change it later if we find a reason.

---
In reply to: [149802945](https://github.com/Azure/azure-iot-sdk-node/pull/135#discussion_r149802945) [](ancestors = 149802945,148917838)

provisioning/e2e/_provisioning_e2e.js
BertKleewein: Like I said, I go back and forth on the usefulness of promises.

However, this catch exists to ignore the error on deleteIndividualEnrollment then continue.  In other words, if deleteIndividualEnrollment fails, this catch will execute, then the flow will continue with a call to registry.delete.

reference: [http://odetocode.com/blogs/scott/archive/2015/10/01/javascript-promises-and-error-handling.aspx](http://odetocode.com/blogs/scott/archive/2015/10/01/javascript-promises-and-error-handling.aspx)

'Just like normal execution can resume in procedural code after a catch statement, normal execution can resume with promises after a handled error.'

---
In reply to: [148917887](https://github.com/Azure/azure-iot-sdk-node/pull/135#discussion_r148917887) [](ancestors = 148917887)

provisioning/e2e/_provisioning_e2e.js
BertKleewein: OK.  will remove this on the next iteration.

---
In reply to: [148917962](https://github.com/Azure/azure-iot-sdk-node/pull/135#discussion_r148917962) [](ancestors = 148917962)

BertKleewein:
>         waitingToPoll: {

I can agree with this.  I'll remove this in a later commit.

---
In reply to: [342306515](https://github.com/Azure/azure-iot-sdk-node/pull/135#issuecomment-342306515) [](ancestors = 342306515)

---
Refers to: provisioning/device/src/transport_state_machine.ts:132 in ef60661. [](commit_id = ef60661284986df0511b46a52d43728636eeb0ee, deletion_comment = False)

provisioning/device/src/interfaces.ts
BertKleewein: nope.  csharp uses idScope, which matches the portal UI

---
In reply to: [148917686](https://github.com/Azure/azure-iot-sdk-node/pull/135#discussion_r148917686) [](ancestors = 148917686)

provisioning/device/src/interfaces.ts
BertKleewein: we don't.  I was trying a different flow and it doesn't work well.  removing.

---
In reply to: [148918078](https://github.com/Azure/azure-iot-sdk-node/pull/135#discussion_r148918078) [](ancestors = 148918078)

BertKleewein: next commit.  needs some refactoring.

---
In reply to: [149808630](https://github.com/Azure/azure-iot-sdk-node/pull/135#discussion_r149808630) [](ancestors = 149808630,148918078)

provisioning/device/src/client.ts
BertKleewein: agree.  refactoring in later commit.

---
In reply to: [148918192](https://github.com/Azure/azure-iot-sdk-node/pull/135#discussion_r148918192) [](ancestors = 148918192)

provisioning/device/src/client.ts
BertKleewein: yes, ugly.  refactoring.

---
In reply to: [148918155](https://github.com/Azure/azure-iot-sdk-node/pull/135#discussion_r148918155) [](ancestors = 148918155)

PR #137, opened 2017-11-06
Adjust to new names and add better documentations and e2e

anthonyvercolano:
# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/provisioning/e2e/_service_create_delete.js
+++ b/provisioning/e2e/_service_create_delete.js
@@ -1,0 +1,95 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+
+var assert = require('chai').assert;
+var debug = require('debug')('azure-device-provisioning-e2e');
+var uuid = require('uuid');
+
+
+var provisioningServiceClient = require('azure-iot-provisioning-service').ProvisioningServiceClient;
+
+var serviceClient = provisioningServiceClient.fromConnectionString(process.env.IOT_PROVISIONING_SERVICE_CONNECTION_STRING);
+
+var enrollment = {
+  registrationId: 'e2e-node-deleteme-psc-' + uuid.v4(),
+  attestation: {
+    type: 'tpm',
+    tpm: {
+      endorsementKey: 'a',
+      storageRootKey: 'b'
+    }
+  }
+};
+
+var enrollmentGroup = {
+  enrollmentGroupId: 'e2e-node-deleteme-psc-' + uuid.v4(),
+  attestation: {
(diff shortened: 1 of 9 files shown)


provisioning/service/src/interfaces.ts
```diff
@@ -18,138 +30,336 @@ export type AttestationTypes = 'none' | 'tpm' | 'x509';
...
+ * This object is not provided by the application.  It is populated by the service from
+ * a provided certificate.
  */
 export interface X509CertificateInfo {
+  /**
+   * For clientCertificates the subjectName is used as the deviceId on the IoT Hub.
+   */
   subjectName: string;
+  /**
+   * A hash of the certificate whose size is 160 bits.
```
BertKleewein: >A hash of the certificate whose size is 160 bits [](start = 5, length = 48)

awkward.  It sounds like the certificate is 160 bits.  Suggest "a 160 bit hash of the certificate"?  Same for sha256

anthonyvercolano: Done.

---
In reply to: [149230292](https://github.com/Azure/azure-iot-sdk-node/pull/137#discussion_r149230292) [](ancestors = 149230292)

provisioning/e2e/_service_create_delete.js
```diff
@@ -0,0 +1,95 @@
...
+'use strict';
+
+var assert = require('chai').assert;
+var debug = require('debug')('azure-device-provisioning-e2e');
+var uuid = require('uuid');
+
+
+var provisioningServiceClient = require('azure-iot-provisioning-service').ProvisioningServiceClient;
+
+var serviceClient = provisioningServiceClient.fromConnectionString(process.env.IOTHUB_PROVISIONING_SERVICE_CONNECTION_STRING);
```
pierreca: >IOTHUB [](start = 79, length = 6)

nit: IOT not IOTHUB

provisioning/service/src/interfaces.ts
```diff
@@ -18,138 +30,336 @@ export type AttestationTypes = 'none' | 'tpm' | 'x509';
...
 
 /**
  * X509 certificate info.
+ *
+ * This object is not provided by the application.  It is populated by the service from
+ * a provided certificate.
  */
 export interface X509CertificateInfo {
+  /**
+   * For clientCertificates the subjectName is used as the deviceId on the IoT Hub.
```
pierreca: >clientCertificates [](start = 9, length = 18)

camelcasing suggests a property of something?

pierreca: >subjectName [](start = 32, length = 11)

same here - also I think we should refer to this as common name ? or CN property of the subject?

Issue #132, opened 2017-11-02
Debian: MQTT Not detecting disconnects + Loss of messages


dsjh:
# Context
- **OS and version used:** Debian Jessie (8) | 3.16.0-4-amd64 #1 SMP Debian 3.16.43-2+deb8u2 (2017-06-26) x86_64 GNU/Linux 
- **Node.js version:** 4.32
- **npm version:** 2.14.12
- **list of installed packages:** azure-iot-device-mqtt@1.2.1 node_modules/azure-iot-device-mqtt
├── azure-iot-common@1.2.1
├── debug@3.1.0 (ms@2.0.0)
├── azure-iot-mqtt-base@1.2.1 (mqtt@2.13.1)
└── machina@2.0.0 (lodash@3.10.1)
azure-iot-device@1.2.0 node_modules/azure-iot-device
├── debug@3.1.0 (ms@2.0.0)
├── traverse@0.6.6
├── azure-iot-common@1.2.1
├── azure-iot-http-base@1.2.0 (uuid@2.0.3, debug@2.6.9, azure-iot-common@1.2.0)
├── azure-storage@2.6.0 (extend@1.2.1, underscore@1.8.3, xmlbuilder@0.4.3, browserify-mime@1.2.9, validator@3.35.0, uuid@3.1.0, readable-stream@2.0.6, md5.js@1.3.4, json-edm-parser@0.1.2, xml2js@0.2.7)
├── machina@2.0.0 (lodash@3.10.1)
└── lodash@4.17.4
azure-iot-device-amqp@1.2.1 node_modules/azure-iot-device-amqp
├── uuid@3.1.0
├── azure-iot-common@1.2.1
├── debug@3.1.0 (ms@2.0.0)
├── machina@2.0.0 (lodash@3.10.1)
├── async@2.5.0 (lodash@4.17.4)
└── azure-iot-amqp-base@1.2.1 (amqp10-transport-ws@0.0.5, bluebird@3.5.1, amqp10@3.5.3)
- **cloned repo: N/A

# Description of the issue:
We are running the simple device code (here: https://github.com/Azure/azure-iot-sdk-node/blob/master/device/samples/simple_sample_device.js). When the internet connection drops the C2D messages are not getting queued (they are getting black-holed) and D2C messages are also getting lost (with no errors).

[...]

PR #139, opened 2017-11-08
Fix for #132: MqttBase getting stuck while disconnecting if the network is down

pierreca:
# Description of the problems

1. When the network connection is down, the call the mqttClient.end() never calls its callback.
(see https://github.com/mqttjs/MQTT.js/issues/710)
2. When closing and opening the client repeatedly there is an event handler leak. 

# Description of the solution

1. Jump directly to disconnected when we receive a close event since this means the socket is already closed and there is no need to call client.end().
2. Remove the event handler for an event that was removed when the state machines were introduced.

**edit**: added another commit for a separate event handler leak issue.

### DIFF
--- a/common/transport/mqtt/test/_mqtt_base_test.js
+++ b/common/transport/mqtt/test/_mqtt_base_test.js
@@ -569,7 +569,7 @@ describe('MqttBase', function () {
       fakeMqtt.emit('connect');
     });
 
-    it('emits a NotConnectedError when the mqtt client emits an close', function (testCallback) {
+    it('emits a NotConnectedError when the mqtt client emits an close while connected', function (testCallback) {
       var fakeError = new Error('fake');
       var fakeMqtt = new FakeMqtt();
       var mqttBase = new MqttBase('test', fakeMqtt);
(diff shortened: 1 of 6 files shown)


common/transport/mqtt/test/_mqtt_base_test.js
```diff
@@ -569,7 +569,7 @@ describe('MqttBase', function () {
       fakeMqtt.emit('connect');
     });
 
-    it('emits a NotConnectedError when the mqtt client emits an close', function (testCallback) {
+    it('emits a NotConnectedError when the mqtt client emits an close while connected', function (testCallback) {
       var fakeError = new Error('fake');
```
anthonyvercolano: > var fakeError = new Error('fake'); [](start = 5, length = 35)

This serves what purpose?

common/transport/mqtt/test/_mqtt_base_test.js
```diff
@@ -582,5 +582,25 @@ describe('MqttBase', function () {
       });
       fakeMqtt.emit('connect');
     });
+
+    it('swallows the close event if the client is already disconnecting', function (testCallback) {
+      var fakeError = new Error('fake');
```
anthonyvercolano: >var fakeError = new Error('fake'); [](start = 6, length = 34)

Serves what purpose?

PR #146, opened 2017-11-11
Add devDependencies scanning to build_parallel.js

pierreca:
# Description of the problem
build_parallel.js only scans dependencies - and therefore misses to establish necessary links listed in packageJson.devDependencies.

# Description of the solution
Now scanning both dependencies and devDependencies.
Also moved service package up in config.json order to make sure it's installed/built before the packages depending on it.

### DIFF
--- a/build/build_parallel/config.json
+++ b/build/build_parallel/config.json
@@ -57,6 +57,9 @@
       "directory": "device/core"
     },
     {
+      "directory": "service"
+    },
+    {
       "directory": "device/transport/mqtt"
     },
     {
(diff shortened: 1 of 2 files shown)


build/build_parallel/config.json
```diff
@@ -57,6 +57,9 @@
       "directory": "device/core"
     },
     {
+      "directory": "service"
```
BertKleewein: >"directory": "service" [](start = 6, length = 22)

I'm curious to know why this ordering is significant.  AFAIK, the algorithm should be order-independent.

PR #154, opened 2017-11-15
flow and api updates

BertKleewein:
stubbed out a bunch of objects to show the user experience with the api.
Now the user does the following to register:
1. Create the transport object 
1. Create the securityClient object
1. Creates the ProvisioningDeviceClient and calls the registration method.

I'm expecting to change the following:
1. EndSession will be removed from the top level API.  The session will end when the registration is complete.
1. New interfaces will be added for TPM.  Every time you see x509, assume there will be a corresponding change for TPM.
1. Interfaces will be added for X509 and TPM security clients.  I haven't created any interfaces because there's only really 1 implementation so far.
1. The registerX509 method will become a register method.  The client will determine if it's x509 or TPM based on the security client.

### DIFF
--- a/provisioning/device/index.d.ts
+++ b/provisioning/device/index.d.ts
@@ -1,7 +1,8 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-export { Authentication, TransportHandlers, ResponseCallback, ClientConfiguration, DeviceConfiguration } from './lib/interfaces';
-export { ClientStateMachine } from './lib/client_state_machine';
+export { ProvisioningAuthentication, PollingTransportHandlers, ProvisioningTransportOptions, ProvisioningDeviceConfiguration, ProvisioningTransportHandlersBase, ProvisioningTransportHandlersX509 } from './lib/interfaces';
+export { PollingStateMachine } from './lib/polling_state_machine';
 export { ProvisioningDeviceClient } from './lib/client';
+export { ProvisioningDeviceConstants } from './lib/constants';
 
--- a/provisioning/device/src/client.ts
+++ b/provisioning/device/src/client.ts
@@ -2,74 +2,53 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 'use strict';
-import * as Provisioning from './interfaces';
-import { ClientStateMachine } from './client_state_machine';
+import { ProvisioningTransportHandlersX509, ProvisioningTransportHandlersBase, ProvisioningDeviceConfiguration } from './interfaces';
+import { errors } from 'azure-iot-common';
 import * as dbg from 'debug';
 const debug = dbg('azure-device-provisioning:device-client');
 
-// tslint:disable-next-line:no-var-requires
-const packageJson = require('../package.json');
+export class ProvisioningDeviceClient {
 
-export class ProvisioningDeviceClient implements Provisioning.ClientConfiguration {// tslint:disable-next-line:no-var-requires
+  private _transport: ProvisioningTransportHandlersBase;
+  private _securityClient: any;
 
-
-  /**
-   * User-Agent string passed to the service as part of communication
-   */
-  userAgent: string = packageJson.name + '/' + packageJson.version;
-
-  /**
-   * Default interval for polling, to use in case service doesn't provide it to us.
-   */
-  pollingInterval: number = 2000;
-
-  /**
-   * Default host for the provisioning service
-   */
-  provisioningHost: string = 'global.azure-devices-provisioning.net';
-
-  /**
-   * apiVersion to use while communicating with service.
-   */
-  apiVersion: string = '2017-08-31-preview';
-
-  /**
-   * default timeout to use when communicating with the service
-   */
-  timeoutInterval: number = 4000;
-
-  /**
-   * idScope to use when communicating with the provisioning service
-   */
-  idScope: string;
-
-  private _stateMachine: ClientStateMachine;
-
-  private constructor(transport: Provisioning.TransportHandlers, idScope: string) {
-    this.idScope = idScope;
-    transport.setClientConfig(this);
-    this._stateMachine = new ClientStateMachine(transport);
+  private constructor(transport: ProvisioningTransportHandlersBase, securityClient: any) {
+    this._transport = transport;
+    this._securityClient = securityClient;
   }
 
-  register(registrationId: string, authentication: Provisioning.Authentication, forceRegistration: boolean, callback: (err: Error, deviceConfig?: Provisioning.DeviceConfiguration) => void): void {
-    this._stateMachine.register(registrationId, authentication, {'registrationId' : registrationId}, forceRegistration, (err, response) => {
+  registerX509(registrationId: string, forceRegistration: boolean, callback: (err: Error, deviceConfig?: ProvisioningDeviceConfiguration) => void): void {
+    if (!((this._transport as any).registerX509)) {
+      throw new errors.InvalidOperationError('Transport does not support X509 authentication');
+    }
+    if (!((this._securityClient as any).getCertificate)) {
+      throw new errors.InvalidOperationError('Security Client does not support X509 authentication');
+    }
+
+    this._securityClient.getCertificate((err, cert)  => {
       if (err) {
-        debug('_stateMachine.register returned error');
-        debug(err.toString);
+        debug('security client returned error on cert acquisition');
         callback(err);
       } else {
-        let deviceConfig: Provisioning.DeviceConfiguration = {
-          iotHubUri: response.registrationStatus.assignedHub,
-          deviceId: response.registrationStatus.deviceId
-        };
-        callback(null, deviceConfig);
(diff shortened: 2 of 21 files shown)


provisioning/device/index.d.ts
```diff
@@ -1,7 +1,8 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-export { Authentication, TransportHandlers, ResponseCallback, ClientConfiguration, DeviceConfiguration } from './lib/interfaces';
-export { ClientStateMachine } from './lib/client_state_machine';
+export { ProvisioningAuthentication, PollingTransportHandlers, ProvisioningTransportOptions, ProvisioningDeviceConfiguration, ProvisioningTransportHandlersBase, ProvisioningTransportHandlersX509 } from './lib/interfaces';
```
pierreca: >ProvisioningAuthentication [](start = 9, length = 26)

nit: that's a very long line that is probably worth cutting

pierreca: also do we use all of these in other packages?

---
In reply to: [151319163](https://github.com/Azure/azure-iot-sdk-node/pull/154#discussion_r151319163) [](ancestors = 151319163)

security/x509/test/_x509_test.js
```diff
@@ -6,16 +6,20 @@
...
 
   describe('getCertificate', function() {
-    it ('throws', function() {
-      assert.throws(function() {
-        obj.getCertificate();
-      });
+    it ('returns the cert', function(callback) {
+        obj.getCertificate(function(err, cert) {
+          assert(!err);
+          assert.equal(cert, fakeCert);
```
pierreca: >equal [](start = 17, length = 5)

strictEqual (=== vs ==)

provisioning/transport/http/src/http_transport.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  endSession(callback: (err?: Error) => void): void {
+    this._stateMachine.endSession(callback);
+  }
+
+  setTransportOptions(options: ProvisioningTransportOptions): void {
+    this._pollingHandlers.setTransportOptions(options);
+  }
+
+  registerX509(registrationId: string, authorization: X509, forceRegistration: boolean, callback: (err?: Error, assignedHub?: string, deviceId?: string, responseBody?: any, result?: any) => void): void {
```
pierreca: >assignedHub?: string, deviceId?: string [](start = 112, length = 39)

we're going to find these 2 together pretty much all the time, i'd be in favor of creating a RegistrationResult object.

provisioning/e2e/_provisioning_e2e.js
```diff
@@ -80,7 +82,7 @@ describe('IoT Provisioning', function() {
       })
       .then(function() {
         debug('deleting device');
-        registry.delete(deviceId);
+        registry.deleteAsync(deviceId);
```
pierreca: >registry [](start = 8, length = 8)

arent' you supposed to return the promise if you want it to be awaited?

provisioning/device/src/interfaces.ts
```diff
@@ -6,61 +6,50 @@
...
  * type defining the types of authentication we support
  */
-export type Authentication = string | X509 | SharedAccessSignature;
-
-export interface ClientConfiguration {
-  /**
-   * User-Agent string passed to the service as part of communication
-   */
-  userAgent: string;
+export type ProvisioningAuthentication = string | X509 | SharedAccessSignature;
```
pierreca: I think that'd be TPM. it ends up using SAS tokens but the flow is typical of TPM (and yes, it's asymmetrical with x509 but I think that's what other SDKs landed on)

provisioning/device/src/client.ts
```diff
@@ -2,74 +2,53 @@
...
-import * as Provisioning from './interfaces';
-import { ClientStateMachine } from './client_state_machine';
+import { ProvisioningTransportHandlersX509, ProvisioningTransportHandlersBase, ProvisioningDeviceConfiguration } from './interfaces';
+import { errors } from 'azure-iot-common';
 import * as dbg from 'debug';
 const debug = dbg('azure-device-provisioning:device-client');
 
-// tslint:disable-next-line:no-var-requires
-const packageJson = require('../package.json');
+export class ProvisioningDeviceClient {
```
pierreca: >ProvisioningDeviceClient [](start = 13, length = 24)

if you look at the PR i have for TPM I believe this is equivalent to the TpmProvisioning class I have.

initially we started with the idea that the device client would be agnostic of TPM vs x509 and protocol: we'd create it with a security client, and a transport. based on the security client we get, we should determine what type of ProvisioningClient we're using (TpmProvisioning or X509Provisioning: these would not be public objects). then each of those would have their own "flow" state machine and would call into the transport and security client as needed. Of course there's an additional complexity/separation that I don't have on the TPM side that is, you have a polling state machine. That polling state machine should be hidden within the transport layers themselves - the polling mechanism is a transport-specific thing, even if all transports share it (for now), it really is a transport-layer responsibility.

This looks to me like the X509Provisioning class, not the ProvisioningDeviceClient

provisioning/device/src/client.ts
```diff
@@ -2,74 +2,53 @@
...
 
-  register(registrationId: string, authentication: Provisioning.Authentication, forceRegistration: boolean, callback: (err: Error, deviceConfig?: Provisioning.DeviceConfiguration) => void): void {
-    this._stateMachine.register(registrationId, authentication, {'registrationId' : registrationId}, forceRegistration, (err, response) => {
+  registerX509(registrationId: string, forceRegistration: boolean, callback: (err: Error, deviceConfig?: ProvisioningDeviceConfiguration) => void): void {
+    if (!((this._transport as any).registerX509)) {
+      throw new errors.InvalidOperationError('Transport does not support X509 authentication');
+    }
+    if (!((this._securityClient as any).getCertificate)) {
+      throw new errors.InvalidOperationError('Security Client does not support X509 authentication');
+    }
```
pierreca: if this becomes the X509ProvisioningClass you don't need that. we can just use compiler type safety.

provisioning/device/src/client.ts
```diff
@@ -2,74 +2,53 @@
...
-          deviceId: response.registrationStatus.deviceId
-        };
-        callback(null, deviceConfig);
+        (this._transport as ProvisioningTransportHandlersX509).registerX509(registrationId, cert, forceRegistration, (err, assignedHub, deviceId) => {
+          if (err) {
+            debug('_stateMachine.register returned error');
+            debug(err.toString);
+            callback(err);
+          } else {
+            let deviceConfig: ProvisioningDeviceConfiguration = {
```
pierreca: >ProvisioningDeviceConfiguration [](start = 30, length = 31)

why not let registerX509 callback use that class instead of 2 parameters?

provisioning/device/src/polling_state_machine.ts
```diff
@@ -6,17 +6,17 @@
 import { EventEmitter } from 'events';
 import { errors } from 'azure-iot-common';
 import * as machina from 'machina';
-import * as Provisioning from './interfaces';
+import { ProvisioningAuthentication, PollingTransportHandlers } from './interfaces';
 import * as dbg from 'debug';
 const debug = dbg('azure-device-provisioning:transport-fsm');
 
-export class  ClientStateMachine extends EventEmitter {
+export class  PollingStateMachine extends EventEmitter {
```
pierreca: >PollingStateMachine [](start = 14, length = 19)

the authorization object we carry around make this very x509 specific. is there a way to make this polling state machine agnostic from x509 or TPM? we are going to have the same need with TPM except the authorization will be a SAS token.

provisioning/transport/http/src/http_transport.ts
BertKleewein: Good idea.  The ProvisioningDeviceConfiguration object in interfaces.ts has this.  I'll rename that and use it whenever we have these 2 together

---
In reply to: [151319471](https://github.com/Azure/azure-iot-sdk-node/pull/154#discussion_r151319471) [](ancestors = 151319471)

provisioning/device/index.d.ts
BertKleewein: Yes we do.  Most of these have to do with passing information between the device client and the transport.

---
In reply to: [151319210](https://github.com/Azure/azure-iot-sdk-node/pull/154#discussion_r151319210) [](ancestors = 151319210,151319163)

provisioning/device/src/polling_state_machine.ts
BertKleewein: It should be agnostic already AFAIK.  This is what we're passing around (taken from restApiclient): 
export type ProvisioningAuthentication = string | X509 | SharedAccessSignature;

---
In reply to: [151320986](https://github.com/Azure/azure-iot-sdk-node/pull/154#discussion_r151320986) [](ancestors = 151320986)

provisioning/e2e/_provisioning_e2e.js
BertKleewein: YES!  you are my hero.  

---
In reply to: [151319694](https://github.com/Azure/azure-iot-sdk-node/pull/154#discussion_r151319694) [](ancestors = 151319694)

PR #155, opened 2017-11-16
DPS SDK: TPM Provisioning state machine (cancel method not tested)

pierreca:
# Description of the problem
TPM registration flow requires a state machine that is independent of the transport used

# Description of the solution
This is the state machine for the TPM registration flow.
I've added a `cancel` method to the API but haven't tested it because we need to agree whether we want this in our objects or not. I think we do... the C# SDK has a cancellation token. if we decide to go with it I'll populate the tests and coverage will go back up.

### DIFF
--- a/provisioning/device/src/tpm_registration.ts
+++ b/provisioning/device/src/tpm_registration.ts
@@ -1,0 +1,239 @@
+import { EventEmitter } from 'events';
+import * as machina from 'machina';
+import * as dbg from 'debug';
+const debug = dbg('azure-iot-provisioning-device:TpmRegistration');
+import { RegistrationClient, TpmProvisioningTransport, TpmSecurityClient, TpmRegistrationInfo } from './interfaces';
+
+export class TpmRegistration extends EventEmitter implements RegistrationClient {
+  private _fsm: machina.Fsm;
+  private _transport: TpmProvisioningTransport;
+  private _securityClient: TpmSecurityClient;
+  private _idScope: string;
+
+  constructor(idScope: string, transport: TpmProvisioningTransport, securityClient: TpmSecurityClient) {
+    super();
+    this._idScope = idScope;
+    this._transport = transport;
+    this._securityClient = securityClient;
+
+    this._fsm = new machina.Fsm({
+      namespace: 'tpm-registration',
+      initialState: 'notStarted',
+      states: {
+        notStarted: {
+          _onEnter: (err, callback) => {
+            if (callback) {
+              callback(err);
+            } else if (err) {
+              this.emit('error', err);
+            }
+          },
+          register: (callback) => this._fsm.transition('authenticating', callback),
+          cancel: (callback) => callback()
+        },
+        authenticating: {
+          _onEnter: (registerCallback) => {
+            let registrationInfo: TpmRegistrationInfo = {
+              endorsementKey: undefined,
+              storageRootKey: undefined,
+              registrationId: undefined
+            };
+            /*Codes_SRS_NODE_DPS_TPM_REGISTRATION_16_001: [The `register` method shall get the endorsement key by calling `getEndorsementKey` on the `TpmSecurityClient` object passed to the constructor.]*/
+            this._securityClient.getEndorsementKey((err, ek) => {
+              if (err) {
+                debug('failed to get endorsement key from TPM security client: ' + err.toString());
+                /*Codes_SRS_NODE_DPS_TPM_REGISTRATION_16_010: [If any of the calls the the `TpmSecurityClient` or the `TpmProvisioningTransport` fails, the `register` method shall call its callback with the error resulting from the failure.]*/
+                this._fsm.transition('notStarted', err, registerCallback);
+              } else {
+                registrationInfo.endorsementKey = ek;
+                registrationInfo.registrationId = this._createRegistrationIdFromEndorsementKey(ek);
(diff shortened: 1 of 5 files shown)


provisioning/device/src/tpm_registration.ts
```diff
@@ -0,0 +1,239 @@
...
+const debug = dbg('azure-iot-provisioning-device:TpmRegistration');
+import { RegistrationClient, TpmProvisioningTransport, TpmSecurityClient, TpmRegistrationInfo } from './interfaces';
+
+export class TpmRegistration extends EventEmitter implements RegistrationClient {
+  private _fsm: machina.Fsm;
+  private _transport: TpmProvisioningTransport;
+  private _securityClient: TpmSecurityClient;
+  private _idScope: string;
+
+  constructor(idScope: string, transport: TpmProvisioningTransport, securityClient: TpmSecurityClient) {
```
BertKleewein: >idScope: string [](start = 14, length = 15)

For X509, I made the scope a parameter of the transport constructor.  I did this because the transport is the only component that needs this.  e.g. the X509Registration object doesn't need to know about it.  We should synchronize here.  I could certainly move it to the client constructor.

provisioning/device/src/tpm_registration.ts
```diff
@@ -0,0 +1,239 @@
...
+            if (callback) {
+              callback(err);
+            } else if (err) {
+              this.emit('error', err);
+            }
+          },
+          register: (callback) => this._fsm.transition('authenticating', callback),
+          cancel: (callback) => callback()
+        },
+        authenticating: {
```
BertKleewein: >authenticating [](start = 8, length = 14)

Should we add explicit error handling if the user calls register while a registration is in progress?  Or does the FSM provide a meaningful enough error if there's no hander available in the current state?

provisioning/device/src/tpm_registration.ts
```diff
@@ -0,0 +1,239 @@
...
+              }
+              this._securityClient.cancel((err) => {
+                if (err) {
+                  debug('failed to stop provisioning transport: ' + err.toString());
+                }
+                this._fsm.transition('notStarted', err, callback);
+              });
+            });
+          },
+          '*': () => this._fsm.deferUntilTransition()
```
BertKleewein: >'*': () => this._fsm.deferUntilTransition() [](start = 10, length = 43)

This is used in some states but not others.  Is this by design?  This relates to my question about expected behavior if the user calls register if a registration is in progress.

provisioning/device/src/tpm_registration.ts
```diff
@@ -0,0 +1,239 @@
...
+            - `symmetricKey`: the symmetric key property of the `TpmChallenge` object returned by the previous call to `TpmProvisioningTransport.getAuthenticationChallenge`
+            - a callback that will handle an optional error if the operation fails.
+            ]*/
+            this._securityClient.activateIdentityKey(registrationResult.symmetricKey, (err) => {
+              if (err) {
+                debug('failed to stop provisioning transport: ' + err.toString());
+                /*Codes_SRS_NODE_DPS_TPM_REGISTRATION_16_010: [If any of the calls the the `TpmSecurityClient` or the `TpmProvisioningTransport` fails, the `register` method shall call its callback with the error resulting from the failure.]*/
+                this._fsm.transition('notStarted', err, registerCallback);
+              } else {
+                this._fsm.transition('completed', registrationResult, registerCallback);
```
BertKleewein: >this._fsm.transition('completed', registrationResult, registerCallback); [](start = 16, length = 72)

where do we close the TLS connection?  I know we don't expose the endSession operation to the caller, but what level is responsible for closing the connection?  (I'm thinking the transport could do this internally, but we should make sure we agree)

provisioning/device/src/tpm_registration.ts
pierreca: the '*' action will defer it until the next transition - then the next transition takes care of it.

---
In reply to: [152068231](https://github.com/Azure/azure-iot-sdk-node/pull/155#discussion_r152068231) [](ancestors = 152068231)

provisioning/device/src/tpm_registration.ts
pierreca: it should be by design. If there is no "catch-all" clause like this one, the state machine doesn't error out, it just doesn't do anything.

---
In reply to: [152068942](https://github.com/Azure/azure-iot-sdk-node/pull/155#discussion_r152068942) [](ancestors = 152068942)

provisioning/device/src/tpm_registration.ts
pierreca: I think the transport should take care of it since the connection is just an artifact of the transport handling the registration request. 

---
In reply to: [152069440](https://github.com/Azure/azure-iot-sdk-node/pull/155#discussion_r152069440) [](ancestors = 152069440)

provisioning/device/src/tpm_registration.ts
pierreca: that's a good question... The only place we use the idScope is to generate the SAS. I'm not sure yet if the format of the SAS is generic or transport-specific. Let's see how that pans out with an actual transport implementation.

---
In reply to: [151768228](https://github.com/Azure/azure-iot-sdk-node/pull/155#discussion_r151768228) [](ancestors = 151768228)

PR #158, opened 2017-11-20
Add E2E diagnostics support for AMQP and MQTT messaging

erich-wang:
end to end diagnostic design spec for device SDKs is available at [here](
https://microsoft.sharepoint.com/:w:/r/teams/Azure_IoT/_layouts/15/Doc.aspx?sourcedoc=%7B2B696832-DC1D-4CA1-AF2F-C15647D34146%7D&file=Azure%20IoT%20Hub%20Device%20Client%20SDKs%20-%20E2E%20Diagnostic%20P0%20scenario.docx)

### DIFF
--- a/common/core/src/message.ts
+++ b/common/core/src/message.ts
@@ -47,7 +47,7 @@ export class Message {
   /**
    * Diagnostic data attached to the message
    */
-  diagnosticPropertyData: any;
+  diagnostics: any;
 
   /**
    * @private
@@ -72,7 +72,7 @@ export class Message {
     this.lockToken = '';
     this.correlationId = '';
     this.userId = '';
-    this.diagnosticPropertyData = null;
+    this.diagnostics = null;
   }
 
   /**
--- a/common/core/src/message.ts
+++ b/common/core/src/message.ts
@@ -44,6 +44,10 @@ export class Message {
    * Type of feedback requested (in case of cloud-to-device command)
    */
   ack: string;
+  /**
+   * Diagnostic data attached to the message
+   */
+  diagnosticPropertyData: any;
 
   /**
    * @private
@@ -68,6 +72,7 @@ export class Message {
     this.lockToken = '';
     this.correlationId = '';
     this.userId = '';
+    this.diagnosticPropertyData = null;
   }
 
   /**
--- a/common/transport/amqp/src/amqp_message.ts
+++ b/common/transport/amqp/src/amqp_message.ts
@@ -54,6 +54,7 @@ export class AmqpMessage {
     let amqpMessage = new AmqpMessage();
     /*Codes_SRS_NODE_IOTHUB_AMQPMSG_05_002: [The created AmqpMessage object shall have a property of type Object named properties.]*/
     amqpMessage.properties = {};
+    amqpMessage.messageAnnotations = {};
 
     /*Codes_SRS_NODE_IOTHUB_AMQPMSG_05_003: [If the message argument has a to property, the properties property of the AmqpMessage object shall have a property named to with the same value.]*/
     if (message.to) {
@@ -86,6 +87,12 @@ export class AmqpMessage {
     if (message.ack) {
       ensureApplicationPropertiesCreated();
       amqpMessage.applicationProperties['iothub-ack'] = message.ack;
+    }
+
+    if (message.diagnosticPropertyData) {
+      // Codes_SRS_NODE_IOTHUB_AMQPMSG_26_001: [If message.diagnosticPropertyData is truthy, the function shall add diagnostic information as AMQP message annotation.]
+        amqpMessage.messageAnnotations['Diagnostic-Id'] = message.diagnosticPropertyData.getDiagnosticId();
+        amqpMessage.messageAnnotations['Correlation-Context'] = message.diagnosticPropertyData.getCorrelationContext();
     }
 
     /*Codes_SRS_NODE_IOTHUB_AMQPMSG_13_001: [ If message.properties is truthy, then all the properties in it shall be copied to the applicationProperties property of the AmqpMessage object. ]*/
--- a/common/transport/amqp/src/amqp_message.ts
+++ b/common/transport/amqp/src/amqp_message.ts
@@ -89,10 +89,11 @@ export class AmqpMessage {
       amqpMessage.applicationProperties['iothub-ack'] = message.ack;
     }
 
-    if (message.diagnosticPropertyData) {
-      // Codes_SRS_NODE_IOTHUB_AMQPMSG_26_001: [If message.diagnosticPropertyData is truthy, the function shall add diagnostic information as AMQP message annotation.]
-        amqpMessage.messageAnnotations['Diagnostic-Id'] = message.diagnosticPropertyData.getDiagnosticId();
-        amqpMessage.messageAnnotations['Correlation-Context'] = message.diagnosticPropertyData.getCorrelationContext();
+    if (message.diagnostics) {
+      // Codes_SRS_NODE_IOTHUB_AMQPMSG_26_001: [If message.diagnostics is truthy, the function shall add diagnostic id as AMQP message annotation.]
+      amqpMessage.messageAnnotations['Diagnostic-Id'] = message.diagnostics.id;
+      // Codes_SRS_NODE_IOTHUB_AMQPMSG_26_002: [If message.diagnostics is truthy, the function shall add diagnostic context as AMQP message annotation.]
+      amqpMessage.messageAnnotations['Correlation-Context'] = message.diagnostics.getEncodedCorrelationContext();
     }
 
     /*Codes_SRS_NODE_IOTHUB_AMQPMSG_13_001: [ If message.properties is truthy, then all the properties in it shall be copied to the applicationProperties property of the AmqpMessage object. ]*/
--- a/device/core/src/client.ts
+++ b/device/core/src/client.ts
@@ -95,9 +95,6 @@ export class Client extends EventEmitter {
 
     this.blobUploadClient = blobUploadClient;
 
-    /* Codes_SRS_NODE_DEVICE_CLIENT_26_001: [The constructor shall initialize device client diagnostic] */
-    this._diagnosticClient = new DiagnosticClient();
-
     this._transport = transport;
     this._transport.on('message', (msg) => {
       this.emit('message', msg);
@@ -264,7 +261,10 @@ export class Client extends EventEmitter {
    * @param {Function}                  sendEventCallback  The callback to be invoked when `sendEvent` completes execution.
    */
   sendEvent(message: Message, sendEventCallback?: (err?: Error, result?: results.MessageEnqueued) => void): void {
-    this._diagnosticClient.addDiagnosticInfoIfNecessary(message);
+    /* Codes_SRS_NODE_DEVICE_CLIENT_26_002: [The sendEvent method shall add diagnostic information if necessary.] */
+    if (this._diagnosticClient) {
+      this._diagnosticClient.addDiagnosticInfoIfNecessary(message);
+    }
     const retryOp = new RetryOperation(this._retryPolicy, this._maxOperationTimeout);
     retryOp.retry((opCallback) => {
       /*Codes_SRS_NODE_DEVICE_CLIENT_05_007: [The sendEvent method shall send the event indicated by the message argument via the transport associated with the Client instance.]*/
@@ -286,8 +286,11 @@ export class Client extends EventEmitter {
    *                                                `sendEventBatch` completes execution.
    */
   sendEventBatch(messages: Message[], sendEventBatchCallback?: (err?: Error, result?: results.MessageEnqueued) => void): void {
-    for (let message of messages) {
-      this._diagnosticClient.addDiagnosticInfoIfNecessary(message);
+    /* Codes_SRS_NODE_DEVICE_CLIENT_26_003: [The sendEventBatch method shall add diagnostic information to all messages if necessary.] */
+    if (this._diagnosticClient) {
+      for (let message of messages) {
+        this._diagnosticClient.addDiagnosticInfoIfNecessary(message);
+      }
     }
     const retryOp = new RetryOperation(this._retryPolicy, this._maxOperationTimeout);
     retryOp.retry((opCallback) => {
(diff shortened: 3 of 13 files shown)


pierreca:
I'm not seeing a Node section in the spec - is it going to be added later?
Also, since we're doing all our work in the open, linking to an internal document is not great - I would much rather see the description of the feature in the PR description itself.

common/core/src/message.ts
```diff
@@ -44,6 +44,10 @@ export class Message {
    * Type of feedback requested (in case of cloud-to-device command)
    */
   ack: string;
+  /**
+   * Diagnostic data attached to the message
+   */
+  diagnosticPropertyData: any;
```
pierreca: >diagnosticPropertyData [](start = 2, length = 22)

I would rather simply name this "diagnostics" since "propertyData" doesn't add much value to the name. What would add value though is a proper type. I'm guessing something like:

interface Diagnostics {
  id: string;
  correlationContext: string; (or even better, a key/value pair object that properly overrides toString() for serialization)
}

Also, unlike Java, we don't really use getters and setters. So having properties directly instead of methods (getDiagnosticId/getCorrelationContext) would be better.

common/transport/amqp/src/amqp_message.ts
```diff
@@ -88,6 +89,12 @@ export class AmqpMessage {
       amqpMessage.applicationProperties['iothub-ack'] = message.ack;
     }
 
+    if (message.diagnosticPropertyData) {
+      // Codes_SRS_NODE_IOTHUB_AMQPMSG_26_001: [If message.diagnosticPropertyData is truthy, the function shall add diagnostic information as AMQP message annotation.]
```
pierreca: >Codes_SRS_NODE_IOTHUB_AMQPMSG_26_001 [](start = 9, length = 36)

Please add the corresponding requirement in common/transport/amqp/devdoc/amqp_message.md

pierreca: >add diagnostic information [](start = 112, length = 26)

this is a little vague - i'd much rather see 2 requirements, one for each property, with expected format for each property.

device/core/src/client.ts
```diff
@@ -483,6 +492,22 @@ export class Client extends EventEmitter {
...
   }
 
+  /*
+   * @method           module:azure-iot-device.Client#setDiagnosticSamplingPercentage
+   * @description      The `setDiagnosticSamplingPercentage` method will set percentage value for diagnostic sampling
+   *
+   * @param {number} value            The value of diagnostic sampling
+   *
+   */
+  setDiagnosticSamplingPercentage(value: number): void {
```
pierreca: >setDiagnosticSamplingPercentage [](start = 2, length = 31)

we already have a setOptions call in the client - why create  a separate API?

device/core/src/client_diagnostic.ts
```diff
@@ -0,0 +1,205 @@
...
+      // 65 is char code of 'A'
+      return String.fromCharCode(65 + value - 10);
+    } else {
+      // 97 is char code of 'a'
+      return String.fromCharCode(97 + value - 36);
+    }
+  }
+
+  private generateEightRandomCharacters(): string {
+    // Codes_SRS_DEVICECLIENTDIAGNOSTIC_01_005: [This function shall generate 8 random chars, each is from 0-9a-z.]
```
pierreca: >Codes_SRS_DEVICECLIENTDIAGNOSTIC_01_005 [](start = 7, length = 39)

all these belong in a requirement document that we should review together prior implementation.

device/core/src/client_diagnostic.ts
```diff
@@ -0,0 +1,205 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+
+import { Message, errors } from 'azure-iot-common';
+
+/**
+ * Data structure to store diagnostic property data
+ */
+export class DiagnosticPropertyData {
```
pierreca: >DiagnosticPropertyData [](start = 13, length = 22)

"PropertyData" doesn't add value to the name

device/core/src/client_diagnostic.ts
```diff
@@ -0,0 +1,205 @@
...
+  // And symbol in diagnostic context.
+  static DIAGNOSTIC_CONTEXT_SYMBOL_AND: string = '&';
+
+  private diagnosticId: string;
+  private diagnosticCreationTimeUtc: string;
+
+  /**
+   * @constructor
+   */
+  constructor(diagnosticId: string, diagnosticCreationTimeUtc: string) {
```
pierreca: >diagnosticId [](start = 14, length = 12)

"diagnostic" prefix can be guessed from the class name and is unnecessary too

device/core/src/client_diagnostic.ts
```diff
@@ -0,0 +1,205 @@
...
+    /* Codes_SRS_DIAGNOSTICPROPERTYDATA_01_001: [If the diagnosticId or diagnosticCreationTimeUtc is null, the constructor shall throw an IllegalArgumentException.] */
+    if (!diagnosticId || !diagnosticCreationTimeUtc) {
+      throw new errors.ArgumentError('The diagnosticId or diagnosticCreationTimeUtc cannot be null or empty.');
+    }
```
pierreca: compiler checks that for us. since this is an internal API, no need for the extra tests.

device/core/src/client_diagnostic.ts
```diff
@@ -0,0 +1,205 @@
...
+   * @param {string}    diagnosticId   The value of diagnostic id.
+   * @throws {ArgumentError}    If value is invalid
+   */
+  setDiagnosticId(diagnosticId: string): void {
+    /* Codes_SRS_DIAGNOSTICPROPERTYDATA_01_003: [A valid diagnosticId shall not be null or empty.] */
+    if (!diagnosticId) {
+      throw new errors.ArgumentError('The diagnosticId cannot be null or empty.');
+    }
+    this.diagnosticId = diagnosticId;
+  }
```
pierreca: unnecessary (make the property public and edit it directly - it's an internal API)

device/core/src/client.ts
```diff
@@ -483,6 +492,22 @@ export class Client extends EventEmitter {
...
+    if (value !== null && typeof value !== 'undefined') {
+      // Codes_SRS_NODE_DEVICE_CLIENT_26_002: ["SetDiagnosticSamplingPercentage" would set percentage].
+      this._diagnosticClient.setDiagSamplingPercentage(value);
+    }else {
```
pierreca: nit: missing space

device/core/src/client_diagnostic.ts
```diff
@@ -0,0 +1,205 @@
...
+   * @method            module:azure-iot-device.DiagnosticClient.setDiagSamplingPercentage
+   * @description       set the diagnostic sampling percentage.
+   *
+   * @param {number}   diagSamplingPercentage   The value of sampling percentage.
+   * @throws {ArgumentError}     If value < 0 or value > 100.
+   */
+  setDiagSamplingPercentage(diagSamplingPercentage: number): void {
+    if (typeof diagSamplingPercentage !== 'number') {
+      throw new errors.ArgumentError('Sampling percentage only accept number');
+    }
```
pierreca: unnecessary - compiler takes care of that for you

device/core/src/client_diagnostic.ts
```diff
@@ -0,0 +1,205 @@
...
+    if (diagSamplingPercentage % 1 !== 0) {
+      throw new errors.ArgumentError('Sampling percentage should be integer');
+    }
```
pierreca: do we care if it's not? I wouldn't think so.

device/core/src/client_diagnostic.ts
```diff
@@ -0,0 +1,205 @@
...
+    // Codes_SRS_DEVICECLIENTDIAGNOSTIC_01_007: [This function shall return false if sampling percentage is set to 0.]
+    if (this.diagSamplingPercentage > 0 && this.diagSamplingPercentage <= 100) {
+      if (this.currentMessageNumber === 100) {
+        this.currentMessageNumber = 0;
+      }
+      this.currentMessageNumber++;
+      // Codes_SRS_DEVICECLIENTDIAGNOSTIC_01_008: [This function shall return value due to the sampling percentage setting.]
+      result = (Math.floor((this.currentMessageNumber - 2) * this.diagSamplingPercentage / 100.0) < Math.floor((this.currentMessageNumber - 1) * this.diagSamplingPercentage / 100.0));
+    }
+    return result;
```
pierreca: why not simply generate a random number and return true if <= diagSamplingPercentage?

device/core/src/client_diagnostic.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  private generateEightRandomCharacters(): string {
+    // Codes_SRS_DEVICECLIENTDIAGNOSTIC_01_005: [This function shall generate 8 random chars, each is from 0-9a-z.]
+    let result: string = '';
+    for (let i = 0; i < 8; i++) {
+      let randomNum: number = Math.floor(Math.random() * DiagnosticClient.DIAGNOSTIC_ID_CHARACTER_BASE);
+      result += this.getDiagnosticIdChar(randomNum);
+    }
+    return result.toString();
+  }
```
pierreca: or we could just use guids and save us all that trouble.

device/core/src/client_diagnostic.ts
michaeljqzq: fixed

device/core/src/client_diagnostic.ts
michaeljqzq: fixed

device/core/src/client_diagnostic.ts
michaeljqzq: removed null check

device/core/src/client_diagnostic.ts
michaeljqzq: removed null check

device/core/src/client_diagnostic.ts
michaeljqzq: removed

device/core/src/client_diagnostic.ts
michaeljqzq: Since our work around in phase 1 is to use device twin properties to deliver the setting from cloud to device, and we only allow integer from 0 - 100(float number is not necessary to support here). We would better check if the value is legal here since the input value is fetched from device twin and user can set any value to it.

device/core/src/client_diagnostic.ts
michaeljqzq: We discussed this with IoT Hub team and made the decision.
You can refer to our design doc here: https://teams.microsoft.com/_#/docx/viewer/teams/https%3A~2F~2Fmicrosoft.sharepoint.com~2Fteams~2FIoTToolingTeam~2FShared%20Documents~2FEnd-to-End%20Diagnostics~2FAzure%20IoT%20Hub%20Device%20Client%20SDKs%20-%20E2E%20Diagnostic%20P0%20scenario.docx?threadId=19%3Acecb83aeeeea41bb818083fc32b545bd%40thread.skype&baseUrl=https%3A~2F~2Fmicrosoft.sharepoint.com~2Fteams~2FIoTToolingTeam&fileId=03F735E4-E751-4917-A3E4-EF04CFB2AAD6&ctx=files&viewerAction=view

device/core/src/client_diagnostic.ts
michaeljqzq: This is by our design. We would like to provide a stable rate of diagnostic sampling. If we use a random number, we can not ensure stability. Here's design doc about this: https://teams.microsoft.com/_#/docx/viewer/teams/https%3A~2F~2Fmicrosoft.sharepoint.com~2Fteams~2FIoTToolingTeam~2FShared%20Documents~2FEnd-to-End%20Diagnostics~2FAzure%20IoT%20Hub%20Device%20Client%20SDKs%20-%20E2E%20Diagnostic%20P0%20scenario.docx?threadId=19%3Acecb83aeeeea41bb818083fc32b545bd%40thread.skype&baseUrl=https%3A~2F~2Fmicrosoft.sharepoint.com~2Fteams~2FIoTToolingTeam&fileId=03F735E4-E751-4917-A3E4-EF04CFB2AAD6&ctx=files&viewerAction=view

common/core/src/message.ts
michaeljqzq: Got it. We will align with this.

device/core/src/client.ts
michaeljqzq: fixed

device/core/src/client.ts
michaeljqzq: setOptions set the options in transport level. setDiagnosticSamplingPercentage set the options in diagnosticClient. Since our logic to handle if attach diagnostic info to message is in diagnosticClient, and transport is not related to diagnosticClient. So we can not reuse setOptions here.

common/core/src/message.ts
michaeljqzq: About override the toString of correlation context, since correlationContext here is a plain object, it is not suitable to override toString on prototype. And create a class for correlationContext isn't necessary. So still use a function getEncodedCorrelationContext to get encoded string. Please correct me if there's a better way.

device/core/src/client.ts
michaeljqzq: And the public function to set percentage is just a work around here. Finally we will only update this setting by device twin.

common/transport/amqp/src/amqp_message.ts
michaeljqzq: added

michaeljqzq: added

pierreca:
Given the current workload of the SDK team and the impact of those changes, these will be reviewed in January. cc @tameraw

common/transport/amqp/src/amqp_message.ts
pierreca: please also specify the name of the property (Diagnostic-Id/Correlation-Context) and the type of the property

---
In reply to: [155454102](https://github.com/Azure/azure-iot-sdk-node/pull/158#discussion_r155454102) [](ancestors = 155454102)

common/core/src/message.ts
pierreca: still need a proper type (ie not "any")

---
In reply to: [155442584](https://github.com/Azure/azure-iot-sdk-node/pull/158#discussion_r155442584) [](ancestors = 155442584)

device/core/src/client_diagnostic.ts
```diff
@@ -0,0 +1,206 @@
...
+
+import { Message, errors } from 'azure-iot-common';
+import { Twin } from './twin';
+import * as dbg from 'debug';
+const debug = dbg('azure-iot-device:Diagnostics');
+
+/**
+ * Data structure to store diagnostic property data
+ */
+export class Diagnostics {
```
pierreca: >Diagnostics [](start = 13, length = 11)

if the class is Diagnostics the file should be named diagnotics.ts (drop the client_ prefix)

device/core/src/client_diagnostic.ts
```diff
@@ -0,0 +1,206 @@
...
+      encodedCorrelationContext += (Diagnostics.DIAGNOSTIC_CONTEXT_PROPERTY[key] + Diagnostics.DIAGNOSTIC_CONTEXT_SYMBOL_EQUAL + this.correlationContext[key]);
+    }
+    return encodedCorrelationContext;
+  }
+}
+
+/**
+ * Handling device client diagnostic information.
+ */
+export class DiagnosticClient {
```
pierreca: >DiagnosticClient [](start = 13, length = 16)

please separate classes in different files - feel free to group the files in a subfolder if you want.

device/core/src/client_diagnostic.ts
pierreca: I do not have access to this

---
In reply to: [155441477](https://github.com/Azure/azure-iot-sdk-node/pull/158#discussion_r155441477) [](ancestors = 155441477)

device/core/src/client_diagnostic.ts
```diff
@@ -0,0 +1,206 @@
...
+
+  private generateEightRandomCharacters(): string {
+    // This function shall generate 8 random chars, each is from 0-9a-z.
+    let result: string = '';
+    for (let i = 0; i < 8; i++) {
+      let randomNum: number = Math.floor(Math.random() * DiagnosticClient.DIAGNOSTIC_ID_CHARACTER_BASE);
+      result += this.getDiagnosticIdChar(randomNum);
+    }
+    return result.toString();
+  }
```
pierreca: suggested reading: [https://stackoverflow.com/questions/9407892/how-to-generate-random-sha1-hash-to-use-as-id-in-node-js](https://stackoverflow.com/questions/9407892/how-to-generate-random-sha1-hash-to-use-as-id-in-node-js)

device/core/src/client_diagnostic.ts
pierreca: have you verified your assertion that using Math.random() doesn't give you a good enough distribution? how bad is the error? I'm going to reiterate: it's a pain to maintain this type of stuff and it's going to be prone to breaking. if we get good enough performance with simpler code, simpler should be the way to go.

---
In reply to: [155441804](https://github.com/Azure/azure-iot-sdk-node/pull/158#discussion_r155441804) [](ancestors = 155441804)

device/core/src/client_diagnostic.ts
```diff
@@ -0,0 +1,206 @@
...
+   * @description       set the diagnostic sampling percentage.
+   *
+   * @param {number}   diagSamplingPercentage   The value of sampling percentage.
+   * @throws {ArgumentError}     If value < 0 or value > 100.
+   * @throws {ArgumentError}     If value is not integer.
+   */
+  setDiagSamplingPercentage(diagSamplingPercentage: number): void {
+    /* Codes_SRS_NODE_DEVICE_DIAGNOSTICCLIENT_01_004: [The setDiagSamplingPercentage function shall throw an exception when input parameter is not an integer.] */
+    if (diagSamplingPercentage % 1 !== 0) {
+      throw new errors.ArgumentError('Sampling percentage should be integer');
```
pierreca: >throw [](start = 6, length = 5)

this will crash the whole device client. I understand if we were to ignore incorrect values, but crashing the client based on input coming from the service isn't acceptable especially if there is no validation on the service side

device/core/src/client_diagnostic.ts
pierreca: understood but in case of wrong value, don't throw. that would crash the whole device client. (see my other comment)

---
In reply to: [155440982](https://github.com/Azure/azure-iot-sdk-node/pull/158#discussion_r155440982) [](ancestors = 155440982)

PR #159, opened 2017-11-20
Bertk api

BertKleewein:
# address code review update
1. update names and definitions to match Pierre's PR
1. refactor class to move X509 into a separate object.
1. minor test fixes
1. make dev-setup fail on error


provisioning/e2e/package.json
```diff
@@ -6,7 +6,7 @@
   "license": "MIT",
   "dependencies": {
     "azure-iot-common": "1.2.2",
-    "azure-iot-provisioning-service": "0.0.1",
+    "azure-iot-provisioning-service": "0.0.1-preview",
```
anthonyvercolano: >-preview" [](start = 44, length = 9)

Yeah, that addition kind of messed up dev-setup.

provisioning/e2e/_provisioning_e2e.js
```diff
@@ -53,7 +53,7 @@ describe('IoT Provisioning', function() {
         x509: {
           clientCertificates: {
             primary: {
-              certificate: x509.cert
+              certificate: new Buffer(x509.cert).toString('base64')
```
anthonyvercolano: >new Buffer(x509.cert).toString('base64') [](start = 27, length = 40)

Really?  I'm pretty sure that this would be a pem or a cert and didn't need to be bin64'd again.

provisioning/e2e/_service_create_delete.js
```diff
@@ -29,7 +29,7 @@ var enrollmentGroup = {
     x509: {
       signingCertificates: {
         primary: {
-          certificate: new Buffer(process.env.IOTHUB_CA_ROOT_CERT, 'base64').toString('ascii')
+          certificate: process.env.IOTHUB_CA_ROOT_CERT
```
anthonyvercolano: >process.env.IOTHUB_CA_ROOT_CERT [](start = 23, length = 31)

Nope.

provisioning/device/src/client.ts
```diff
@@ -1,64 +1,27 @@
...
   /**
    * Factory method to use to create a DeviceClient object.
    *
    * @param transport Transport instance to use
-   * @param idScope Scope ID assigned to the current provisioning service instance
+   * @param securityClient: Security client object which can provide access to the necessary secrets and/or encryption functions
    */
-  static create(transport: ProvisioningTransportHandlersBase, securityClient: any): ProvisioningDeviceClient {
-    return new ProvisioningDeviceClient(transport, securityClient);
+  static create(transport: any, securityClient: any): RegistrationClient {
```
pierreca: >any [](start = 27, length = 3)

X509ProvisioningTransport | TpmProvisioningTransport

pierreca: >any [](start = 48, length = 3)

X509SecurityClient | TpmSecurityClient

provisioning/device/src/interfaces.ts
```diff
@@ -28,26 +31,36 @@ export interface ProvisioningTransportOptions {
...
 /**
- * Interface that all provisioning transports must implement in order to support the provisioning service
+ * @private
  */
-
-export interface ProvisioningTransportHandlersBase {
+export interface X509ProvisioningTransport {
   setTransportOptions(options: ProvisioningTransportOptions): void;
   endSession(callback: (err?: Error) => void): void;
+  registerX509(registrationId: string, auth: X509, forceRegistration: boolean, callback: (err?: Error, registrationResult?: X509RegistrationResult, body?: any, result?: any) => void): void;
```
pierreca: >registerX509 [](start = 2, length = 12)

register? no need for the X509 suffix anymore

provisioning/device/src/interfaces.ts
```diff
@@ -56,29 +69,50 @@ export interface PollingTransportHandlers {
   getErrorResult(result: any): any;
 }
 
+/**
+ * Public API used to access the ProvisioningDeviceClient object
+ */
 export interface RegistrationClient {
```
pierreca: I think you could name this ProvisioningDeviceClient too.

provisioning/device/src/interfaces.ts
```diff
@@ -56,29 +69,50 @@ export interface PollingTransportHandlers {
...
 export interface TpmRegistrationInfo {
   registrationId: string;
   endorsementKey: string;
   storageRootKey: string;
 }
 
+/**
+ * Device configuration returned when registration using TPM is complete
+ */
 export interface TpmRegistrationResult {
```
pierreca: >TpmRegistrationResult [](start = 17, length = 21)

wondering if we should have a base class, (RegistrationResult, used by x509) and a TpmRegistrationResult class inheriting from it, with an additional symmetricKey field.

provisioning/device/src/interfaces.ts
```diff
@@ -56,29 +69,50 @@ export interface PollingTransportHandlers {
...
+  assignedHub: string;
   symmetricKey: string;
 }
 
+/**
+ * @private
+ */
 export interface TpmProvisioningTransport {
   getAuthenticationChallenge(registrationInfo: TpmRegistrationInfo, callback: (err: Error, tpmChallenge?: TpmChallenge) => void): void;
   register(registrationInfo: TpmRegistrationInfo, sasToken: string, callback: (err: Error, result?: TpmRegistrationResult) => void): void;
```
pierreca: >register [](start = 2, length = 8)

note for later: this is missing a forceRegistration parameter.

provisioning/transport/http/src/http_transport.ts
```diff
@@ -31,12 +31,16 @@ export class Http  extends EventEmitter implements ProvisioningTransportHandlers
...
+     this._stateMachine.register(registrationId, auth, {'registrationId' : registrationId}, forceRegistration, (err, responseBody, result) => {
       if (err) {
-        callback(err, null, null, responseBody, result );
+        callback(err, null, responseBody, result );
       } else {
-        callback(err, responseBody.registrationStatus.assignedHub, responseBody.registrationStatus.deviceId, responseBody, result);
+        let registrationResult: X509RegistrationResult = {
+          deviceId: responseBody.registrationStatus.deviceId,
+          assignedHub: responseBody.registrationStatus.assignedHub
+        };
```
pierreca: we shouldn't have to cast this - our interface should be there only to document what the service is sending back and we should be able to pass the registrationResult object to the callback directly.

provisioning/device/src/interfaces.ts
BertKleewein: I'd like to be explicit here because a single transport can support x509 and TPM both 

---
In reply to: [153351796](https://github.com/Azure/azure-iot-sdk-node/pull/159#discussion_r153351796) [](ancestors = 153351796)

provisioning/device/src/interfaces.ts
BertKleewein: I'm wondering if this should be inside the TpmRegistrationInfo structure.  I'm also wondering if I should create a parallel X509RegistrationInfo structure.

---
In reply to: [153352352](https://github.com/Azure/azure-iot-sdk-node/pull/159#discussion_r153352352) [](ancestors = 153352352)

pierreca: Rule of thumb for me is that it makes sense to "box" invidual arguments together in a class if multiple functions take the same list of arguments, or if we need to chain function calls to fully fill a structure (although the latter has a code smell i can't explain). if you're seeing this on the X509 side go for it.

PR #161, opened 2017-11-22
E2E for group enrollment

BertKleewein:
1. Removed use of promises from provisioning E2E tests
1. refactored tests
1. added test for X509 group enrollment


BertKleewein:
Joke's on you. I made a copy paste error. The group enrollment code also does an individual enrollment, so the new code doesn't test anything :). I've been working on a fix.

________________________________

Issue #162, opened 2017-11-24
azure-iothub nextAsTwin() "Expected 2 arguments" continuationToken - is not optional


Qba91:
# Context
- **OS and version used:** Windows 10
- **Node.js version:** v8.9.1
- **npm version:** 5.5.1
- **list of installed packages:** 
+-- @types/node@8.0.53
+-- azure-iothub@1.2.1
| +-- async@2.6.0
| | `-- lodash@4.17.4 deduped
| +-- azure-iot-amqp-base@1.2.3
| | +-- amqp10@3.5.3
| | | +-- bl@1.2.1
| | | | `-- readable-stream@2.3.3
| | | |   +-- core-util-is@1.0.2
| | | |   +-- inherits@2.0.3
| | | |   +-- isarray@1.0.0
| | | |   +-- process-nextick-args@1.0.7
| | | |   +-- safe-buffer@5.1.1
| | | |   +-- string_decoder@1.0.3
| | | |   | `-- safe-buffer@5.1.1 deduped
| | | |   `-- util-deprecate@1.0.2
| | | +-- bluebird@3.5.1 deduped
| | | +-- buffer-builder@0.2.0
| | | +-- debug@2.6.9 deduped
| | | +-- lodash@4.17.4 deduped
| | | +-- node-int64@0.4.0
| | | `-- stately.js@1.3.0
| | +-- amqp10-transport-ws@0.0.5
| | | `-- nodejs-websocket@1.7.1
| | +-- async@2.6.0 deduped
| | +-- azure-iot-common@1.2.2 deduped
| | +-- bluebird@3.5.1
| | +-- debug@3.1.0
| | | `-- ms@2.0.0 deduped
| | +-- machina@2.0.0 deduped
| | `-- uuid@3.1.0
| +-- azure-iot-common@1.2.2
| +-- azure-iot-http-base@1.2.2
| | +-- azure-iot-common@1.2.2 deduped
| | +-- debug@2.6.9 deduped
| | `-- uuid@2.0.3
| +-- debug@2.6.9
| | `-- ms@2.0.0
| +-- lodash@4.17.4
| `-- machina@2.0.0
|   `-- lodash@3.10.1
+-- readline-sync@1.4.7
`-- typescript@2.6.1
- **cloned repo:** https://github.com/Azure/azure-iot-sdk-node.git

# Description of the issue:
done is not an optional parameter - not '?'. Function can not be called with one argument.

# Code sample exhibiting the issue:
[...]

PR #165, opened 2017-11-30
Fixes for device enrollment update test and #162

anthonyvercolano:
Enrollment updates were broken.  We need to pass an etag on the update.

Additional fix for an incorrect parameter optional specification for the done function of nextAsTwin

# Description of the solution

The fix for the update was to copy the property (if it existed) etag from the enrollment object into the header "If-Match".  This is a reasonable fix in that unlike delete there is ALWAYS an enrollment object given as the first argument to the update.  Added updating for the enrollment group object to the enrollment group sample.  Also added an e2e test for updating the enrollment group and the individual enrollment.

A github issue (#162) was raised in that the done function was not specified in the typescript definition of nextAsTwin for query.  Added the ? to the parameter  definition.

### DIFF
--- a/provisioning/e2e/_service_create_delete.js
+++ b/provisioning/e2e/_service_create_delete.js
@@ -5,8 +5,6 @@
 
 var assert = require('chai').assert;
 var uuid = require('uuid');
-var pem = require('pem');
-
 
 var provisioningServiceClient = require('azure-iot-provisioning-service').ProvisioningServiceClient;
 
@@ -19,7 +17,8 @@ var enrollment = {
     tpm: {
       endorsementKey: "AToAAQALAAMAsgAgg3GXZ0SEs/gakMyNRqXXJP1S124GUgtk8qHaGzMUaaoABgCAAEMAEAgAAAAAAAEAtKEADl/sNRgmYAjP6gXmbccRaJoTnVixisUaek0OwAzFGN70xt9ZOYp6fhIwfcft3fdVKOrKpXYcTe72CGNkGJGlQz5ti9n2pQ0uJhcX8aefh4Onm7lVlUCQAVp1K0r6zI8vkEXWsBIvwvxk0eMJbFaq146kbTkJHIGczb89RkFH2TX+CgXeZOG9oXQzUNwktmTUacspamune5Wywc/ce8HsDFYchyUHogFhrZ/LPnzyTDXO8sSC5z5dvsUBtUME3iRYDyKgZOfBtmRMqQewD+4iH+ZEJjtsyJiWR8hFhyKROnOuqXfNFwjd5IcNU4wtlKO0cLyXmTOfQK6Da1pr5Q=="
     }
-  }
+  },
+  provisioningStatus: "enabled"
 };
 
 var enrollmentGroup = {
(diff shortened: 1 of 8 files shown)


provisioning/e2e/_service_create_delete.js
```diff
@@ -92,4 +94,29 @@ testDescription.forEach(function(testConfiguration) {
...
+    var enrollmentToDelete = {};
+    after(function(done) {
+      testConfiguration.deleteFunction(enrollmentToDelete, function(err) {
+        assert.isNull(err,'Non null response from the delete AFTER create.');
+        done();
+      });
+    });
+    it(testConfiguration.testDescription, function(callback) {
+      enrollmentToDelete = {};
+      testConfiguration.createFunction(testConfiguration.enrollmentObject, function(err, returnedEnrollment) {
```
pierreca: >createFunction [](start = 24, length = 14)

shouldn't creation be in a before() clause?

provisioning/e2e/_service_create_delete.js
```diff
@@ -92,4 +94,29 @@ testDescription.forEach(function(testConfiguration) {
...
+        assert.isNull(err,'Should be no error from the create');
+        returnedEnrollment.provisioningState = 'disabled';
+        testConfiguration.updateFunction(returnedEnrollment, function(err, updatedEnrollment) {
```
pierreca: >updateFunction [](start = 26, length = 14)

missing verification that the updatedEnrollment is now disabled.

provisioning/service/src/provisioningserviceclient.ts
```diff
@@ -221,6 +221,7 @@ export class ProvisioningServiceClient {
   }
 
   private _createOrUpdate(endpointPrefix: string, enrollment: any, callback?: (err: Error, enrollmentResponse?: any, response?: any) => void): void {
+
     /*Codes_SRS_NODE_PROVISIONING_SERVICE_CLIENT_06_009: [The `createOrUpdateIndividualEnrollment` method shall throw `ReferenceError` if the `IndividualEnrollment` argument is falsy.]*/
```
pierreca: nit: unnecessary blank line

provisioning/e2e/_service_create_delete.js
```diff
@@ -29,32 +28,35 @@ var enrollmentGroup = {
...
         }
       }
     }
-  }
+  },
+  provisioningStatus: "enabled"
 };
 
-var testDescription =   [
+var testSpecification =   [
```
pierreca: >    [](start = 23, length = 3)

nit: not a fan of the formatting here - specifically: 3 spaces between the = and [ and no indentation for the objects within the array?

provisioning/e2e/_service_create_delete.js
anthonyvercolano: Fixed.

---
In reply to: [154166074](https://github.com/Azure/azure-iot-sdk-node/pull/165#discussion_r154166074) [](ancestors = 154166074)

provisioning/e2e/_service_create_delete.js
anthonyvercolano: Fixed.

---
In reply to: [154166762](https://github.com/Azure/azure-iot-sdk-node/pull/165#discussion_r154166762) [](ancestors = 154166762)

provisioning/e2e/_service_create_delete.js
anthonyvercolano: It's a bit of a pain to propagate the value through to the test.

---
In reply to: [154165590](https://github.com/Azure/azure-iot-sdk-node/pull/165#discussion_r154165590) [](ancestors = 154165590)

anthonyvercolano: Fixed it.

---
In reply to: [154196642](https://github.com/Azure/azure-iot-sdk-node/pull/165#discussion_r154196642) [](ancestors = 154196642,154165590)

provisioning/service/src/provisioningserviceclient.ts
anthonyvercolano: Fixed.

---
In reply to: [154166412](https://github.com/Azure/azure-iot-sdk-node/pull/165#discussion_r154166412) [](ancestors = 154166412)

PR #166, opened 2017-11-30
Fix group enrollment E2E

BertKleewein:
Group enrollment E2E was broken.  To fix this, I needed to add our root cert to the Jenkins DPS instances and provide a proof of possession.  This also entailed updating the E2E tests to use this root cert.  There were also a few syntactic changes that I needed to make because of corresponding service changes.

### DIFF
--- a/provisioning/device/package.json
+++ b/provisioning/device/package.json
@@ -30,7 +30,7 @@
     "alltest": "istanbul cover node_modules/mocha/bin/_mocha -- --reporter spec test/_*_test*.js",
     "ci": "npm -s run lint && npm -s run build && npm -s run alltest-min && npm -s run check-cover",
     "test": "npm -s run lint && npm -s run build && npm -s run unittest",
-    "check-cover": "istanbul check-coverage --statements 77 --branches 65 --lines 79 --functions 69"
+    "check-cover": "istanbul check-coverage --statements 76 --branches 65 --lines 77 --functions 69"
   },
   "engines": {
     "node": ">= 0.10"
--- a/provisioning/device/src/polling_state_machine.ts
+++ b/provisioning/device/src/polling_state_machine.ts
@@ -89,6 +89,13 @@ export class  PollingStateMachine extends EventEmitter {
                   this._fsm.transition('waitingToPoll', callback, registrationId, body.operationId, pollingInterval);
                   break;
                 }
+                case 'failed': {
+                  let err = new Error('registration failed');
+                  (err as any).result = result;
+                  (err as any).body = body;
+                  this._fsm.transition('responseError', callback, err, body, result);
+                  break;
+                }
                 default: {
                   /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_016: [ If `PollingTransportHandlers.registrationRequest` succeeds returns with an unknown status, `register` shall fail with a `SyntaxError` and pass the response body and the protocol-specific result to the `callback`. ] */
                   /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_022: [ If `PollingTransportHandlers.queryOperationStatus` succeeds with an unknown status, `register` shall fail with a `SyntaxError` and pass the response body and the protocol-specific result to the `callback`. ] */
--- a/provisioning/e2e/_provisioning_e2e.js
+++ b/provisioning/e2e/_provisioning_e2e.js
@@ -79,9 +79,11 @@ var X509Individual = function() {
           }
         }
       },
-      initialTwinState: {
-        desiredProperties: {
-          testProp: self._testProp
+      initialTwin: {
+        properties: {
+          desired: {
+            testProp: self._testProp
+          }
         }
       }
     };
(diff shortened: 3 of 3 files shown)


provisioning/device/package.json
```diff
@@ -30,7 +30,7 @@
     "alltest": "istanbul cover node_modules/mocha/bin/_mocha -- --reporter spec test/_*_test*.js",
     "ci": "npm -s run lint && npm -s run build && npm -s run alltest-min && npm -s run check-cover",
     "test": "npm -s run lint && npm -s run build && npm -s run unittest",
-    "check-cover": "istanbul check-coverage --statements 77 --branches 65 --lines 79 --functions 69"
+    "check-cover": "istanbul check-coverage --statements 76 --branches 65 --lines 77 --functions 69"
```
anthonyvercolano: >-functions 6 [](start = 86, length = 12)

Consider adding scripts that call out the individual tests as was done for the hub e2e tests.  Could make debugging a bit easier.

provisioning/e2e/_provisioning_e2e.js
```diff
@@ -130,99 +133,81 @@ var X509Group = function() {
...
-      '[req_distinguished_name]',
-      'commonName = IoTHub Test CA',
-      'commonName_max = 64',
-      '[v3_req]',
-      'basicConstraints = critical,CA:TRUE'
-    ].join('\n');
-    pem.createCertificate({ config: caConfig, commonName: 'IoTHub Test CA' }, function (err, caCert) {
+    debug('creating device cert');
+    self._factoryCert = {
+        cert: new Buffer(process.env.IOTHUB_CA_ROOT_CERT,"base64").toString(),
```
anthonyvercolano: >) [](start = 76, length = 1)

'ascii' ?

provisioning/e2e/_provisioning_e2e.js
```diff
@@ -130,99 +133,81 @@ var X509Group = function() {
...
         };
-        createX509Certificate(certOptions, function(err, cert) {
-          if (err) {
-            callback(err);
-          } else {
-            self._cert = cert;
-            callback();
-          }
-        });
+        fs.writeFileSync('device-cert.pem',self._cert.cert);
```
anthonyvercolano: > fs.writeFileSync('device-cert.pem',self._cert.cert) [](start = 7, length = 52)

You're doing this why?

provisioning/device/src/polling_state_machine.ts
```diff
@@ -89,6 +89,13 @@ export class  PollingStateMachine extends EventEmitter {
                   this._fsm.transition('waitingToPoll', callback, registrationId, body.operationId, pollingInterval);
                   break;
                 }
+                case 'failed': {
+                  let err = new Error('registration failed');
```
pierreca: >Error [](start = 32, length = 5)

This might deserve a new error type? or use an existing type?

provisioning/e2e/_provisioning_e2e.js
```diff
@@ -130,99 +133,81 @@ var X509Group = function() {
...
-      'distinguished_name = req_distinguished_name',
-      '[req_distinguished_name]',
-      'commonName = IoTHub Test CA',
-      'commonName_max = 64',
-      '[v3_req]',
-      'basicConstraints = critical,CA:TRUE'
-    ].join('\n');
-    pem.createCertificate({ config: caConfig, commonName: 'IoTHub Test CA' }, function (err, caCert) {
+    debug('creating device cert');
+    self._factoryCert = {
```
pierreca: >_factoryCert [](start = 9, length = 12)

is that using the CA root cert for something else than testing the CA cert functionality?
I'd rather have separate certs for various purposes. easier to rotate/maintain?

pierreca: also, Doesn't that add to the pile of reasons why an external contributor wouldn't be able to run our tests?

---
In reply to: [154234983](https://github.com/Azure/azure-iot-sdk-node/pull/166#discussion_r154234983) [](ancestors = 154234983)

provisioning/e2e/_provisioning_e2e.js
```diff
@@ -231,54 +216,38 @@ var X509Group = function() {
     var transport = new Transport(idScope);
     var provisioningDeviceClient = ProvisioningDeviceClient.create(transport, securityClient);
     provisioningDeviceClient.register(self._registrationId, false, function (err, result) {
+      assert(!!result.deviceId);
```
pierreca: >assert(!!result.deviceId); [](start = 6, length = 26)

assert.isOk() ? not a fan of type coercion here.

provisioning/e2e/_provisioning_e2e.js
BertKleewein: huh.  utf-8 is default.  It should be the same thing, but I'll be explicit.

---
In reply to: [154233237](https://github.com/Azure/azure-iot-sdk-node/pull/166#discussion_r154233237) [](ancestors = 154233237)

provisioning/e2e/_provisioning_e2e.js
BertKleewein: good idea to use a different variable.  I'll switch that around when I actually turn this on (very soon).  It does add overhead to external contributors.  The certGen.sh script that makes this is in the c repo, so it shouldn't be too hard to document.  (Of course, automated POP would be better, but that's a different story)

---
In reply to: [154235122](https://github.com/Azure/azure-iot-sdk-node/pull/166#discussion_r154235122) [](ancestors = 154235122,154234983)

provisioning/e2e/_provisioning_e2e.js
BertKleewein: because I sometimes forget to remove code i used to debug the certs

---
In reply to: [154234327](https://github.com/Azure/azure-iot-sdk-node/pull/166#discussion_r154234327) [](ancestors = 154234327)

PR #169, opened 2017-12-05
finish last cleanup tasks for X509 HTTP registration

BertKleewein:
add tests for device client, add DPS-specific errors, change registration to take registrationId, idScope, and provisioningHost

This finishes all of the remaining work for x509 over HTTP.
1. device client and x509 registration clients brought under test.
1. device client factory modified to handle TPM cases.
1. DPS registration error added
1. DPS-specific translateError function added
1. register call changed to accept object (registrationId, idScope, provisioningHost) instead of just registrationId.

### DIFF
--- a/build/build_parallel/build_parallel.js
+++ b/build/build_parallel/build_parallel.js
@@ -203,7 +203,7 @@ var makeTaskList = promisify(function (callback) {
 var makeTaskFunction = function (subtasks) {
   var options = {
     parallel: true,
-    continueOnError: !!config.task.continueOnError,
+    continueOnError: true,
     printLabel: true,
     stdout: process.stdout,
     stderr: process.stderr
@@ -230,7 +230,11 @@ var makeTaskFunction = function (subtasks) {
         for (let i = 0; i < err.results.length; i++) {
           console.log(err.results[i].name + ' : ' + (err.results[i].code === 0 ? 'succeeded' : 'failed'));
         }
-        done(new Error());
+        if (config.task.continueOnError) {
+          done();
+        } else {
+          done(new Error());
+        }
       });
   };
 
--- a/common/core/src/errors.ts
+++ b/common/core/src/errors.ts
@@ -417,7 +417,7 @@ export class TwinDetachedError extends Error {
 }
 
 /**
- * Error thrown when ant operation (local or remote) is cancelled
+ * Error thrown when any operation (local or remote) is cancelled
  *
  * @augments {Error}
  */
@@ -430,3 +430,16 @@ export class OperationCancelledError extends Error {
   }
 }
 
+/**
+ * Error thrown when a DPS registration operation fails
+ *
+ * @augments {Error}
+ */
+export class DpsRegistrationFailedError extends Error {
+  constructor(message?: string) {
+    super(message);
+    this.name = 'DpsRegistrationFailedError';
+    this.message = message;
+    Error.captureStackTrace(this, this.constructor);
+  }
+}
--- a/common/core/src/errors.ts
+++ b/common/core/src/errors.ts
@@ -435,10 +435,10 @@ export class OperationCancelledError extends Error {
  *
  * @augments {Error}
  */
-export class DpsRegistrationFailedError extends Error {
+export class DeviceRegistrationFailedError extends Error {
   constructor(message?: string) {
     super(message);
-    this.name = 'DpsRegistrationFailedError';
+    this.name = 'ProvisioningRegistrationFailedError';
     this.message = message;
     Error.captureStackTrace(this, this.constructor);
   }
--- a/provisioning/device/devdoc/dps_error_requirements.md
+++ b/provisioning/device/devdoc/dps_error_requirements.md
@@ -1,0 +1,31 @@
+# azure-iot-provisioning-device.translateError Requirements
+
+## Overview
+`translateError` is a method that translates DPS errors into Azure IoT Hub errors, effectively abstracting the error that is returned to the SDK user of from the transport layer.
+
+## Requirements
+
+** SRS_NODE_DPS_ERRORS_18_001: [** `translateError` shall accept 4 arguments:
+ - A custom error message to give context to the user.
+ - the status code that initiated the error
+ - the response body
+ - the response object itself **]**
+
+
+** SRS_NODE_DPS_ERRORS_18_008: [** Any error object returned by `translateError` shall inherit from the generic `Error` Javascript object and have 3 properties:
+ - `response` shall contain the `IncomingMessage` object returned by the HTTP layer.
+ - `reponseBody` shall contain the content of the HTTP response.
+ - `message` shall contain a human-readable error message **]**
+
+** SRS_NODE_DPS_ERRORS_18_002: [** `translateError` shall return an `ArgumentError` if the status code is `400`. **]**
+
+** SRS_NODE_DPS_ERRORS_18_003: [** `translateError` shall return an `UnauthorizedError` if the status code is `401`. **]**
+
+** SRS_NODE_DPS_ERRORS_18_004: [** `translateError` shall return an `DeviceNotFoundError` if the status code is `404`. **]**
+
+** SRS_NODE_DPS_ERRORS_18_005: [** `translateError` shall return an `IotHubQuotaExceededError` if the status code is `429`. **]**
+
+** SRS_NODE_DPS_ERRORS_18_006: [** `translateError` shall return an `InternalServerError` if the status code is `500`. **]**
+
+** SRS_NODE_DPS_ERRORS_18_007: [** If the status code is unknown, `translateError` should return a generic Javascript `Error` object. **]**
+
--- a/provisioning/device/devdoc/provisioning_client_requirements.md
+++ b/provisioning/device/devdoc/provisioning_client_requirements.md
@@ -17,13 +17,13 @@ The `ProvisioningDeviceClient` class is the factory object used by clients to cr
 
 The `create` method is used to retrieve a client object that can be used to register with the Device Provisioning Service
 
-**SRS_PROVISIONING_CLIENT_18_001: [** If `securityClient` supports x509 security and the `transport` supports x509 authentication, then `crate` shall
+**SRS_PROVISIONING_CLIENT_18_001: [** If `securityClient` implements `X509SecurityClient` and the `transport` implements `X509ProvisioningTransport`, then `create` shall
 return an `X509Registration` object. **]**
 
-**SRS_PROVISIONING_CLIENT_18_002: [** If `securityClient` supports x509 security and the `transport` does not support x509 authentication, then `crate` shall throw a `ArgumentError` exepction. **]**
+**SRS_PROVISIONING_CLIENT_18_002: [** If `securityClient` implements `X509SecurityClient` and the `transport` does not implement `X509ProvisioningTransport`, then `create` shall throw a `ArgumentError` exepction. **]**
 
-**SRS_PROVISIONING_CLIENT_18_003: [** If `securityClient` supports TPM security and the `transport` supports TPM authentication, then `create` shall return a `TpmRegistration` object. **]**
+**SRS_PROVISIONING_CLIENT_18_003: [** If `securityClient` implements `TPMSecurityClient` and the `transport` implements `TPMProvisioningTransport`, then `create` shall return a `TpmRegistration` object. **]**
 
-**SRS_PROVISIONING_CLIENT_18_004: [** If `securityClient` supports TPM security and the `transport` does not support TPM authentication, then `crate` shall throw a `ArgumentError` exepction. **]**
+**SRS_PROVISIONING_CLIENT_18_004: [** If `securityClient` implements `TPMSecurityClient` and the `transport` dos not implement `TPMProvisioningTransport`, then `create` shall throw a `ArgumentError` exepction. **]**
 
-**SRS_PROVISIONING_CLIENT_18_005: [** If `securityClient` does not support X509 or TPM security, then `create` shall show an `ArgumentError` exception. **]**
+**SRS_PROVISIONING_CLIENT_18_005: [** If `securityClient` dos not implement `X509ProvisioningTransport` or `TPMProvisioningTransport`, then `create` shall show an `ArgumentError` exception. **]**
--- a/provisioning/device/devdoc/provisioning_client_requirements.md
+++ b/provisioning/device/devdoc/provisioning_client_requirements.md
@@ -1,0 +1,29 @@
+# provisioning client requirements
+
+## Overview
+
+The `ProvisioningDeviceClient` class is the factory object used by clients to create a registration clients.
+
+## Example usage
+
+```js
+// TODO once all the pieces are there in the SDK
+
+```
+
+## Public API
+
+### create(transport, securityClient)
+
+The `create` method is used to retrieve a client object that can be used to register with the Device Provisioning Service
+
+**SRS_PROVISIONING_CLIENT_18_001: [** If `securityClient` supports x509 security and the `transport` supports x509 authentication, then `crate` shall
+return an `X509Registration` object. **]**
+
+**SRS_PROVISIONING_CLIENT_18_002: [** If `securityClient` supports x509 security and the `transport` does not support x509 authentication, then `crate` shall throw a `ArgumentError` exepction. **]**
+
+**SRS_PROVISIONING_CLIENT_18_003: [** If `securityClient` supports TPM security and the `transport` supports TPM authentication, then `create` shall return a `TpmRegistration` object. **]**
+
+**SRS_PROVISIONING_CLIENT_18_004: [** If `securityClient` supports TPM security and the `transport` does not support TPM authentication, then `crate` shall throw a `ArgumentError` exepction. **]**
+
+**SRS_PROVISIONING_CLIENT_18_005: [** If `securityClient` does not support X509 or TPM security, then `create` shall show an `ArgumentError` exception. **]**
--- a/provisioning/device/devdoc/x509_registration_requirements.md
+++ b/provisioning/device/devdoc/x509_registration_requirements.md
@@ -17,7 +17,7 @@ The `X509Registration` class is used to control the registration flow of an X509
 
 The `register` method completes the authentication and registration flow for the user.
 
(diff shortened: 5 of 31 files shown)


build/build_parallel/build_parallel.js
```diff
@@ -203,7 +203,7 @@ var makeTaskList = promisify(function (callback) {
 var makeTaskFunction = function (subtasks) {
   var options = {
     parallel: true,
-    continueOnError: !!config.task.continueOnError,
+    continueOnError: true,
```
pierreca: >true [](start = 21, length = 4)

commenting here just to double check that this is an intended change (since you seem to care about the value below)

common/core/src/errors.ts
```diff
@@ -430,3 +430,16 @@ export class OperationCancelledError extends Error {
   }
 }
 
+/**
+ * Error thrown when a DPS registration operation fails
+ *
+ * @augments {Error}
+ */
+export class DpsRegistrationFailedError extends Error {
```
pierreca: >DpsRegistrationFailedError [](start = 13, length = 26)

We shouldn't be using the DPS acronym. DeviceProvisioningError? DeviceRegistrationFailedError?

provisioning/device/index.d.ts
```diff
@@ -2,8 +2,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 export { ProvisioningAuthentication, PollingTransportHandlers, ProvisioningTransportOptions } from './lib/interfaces';
-export { X509ProvisioningTransport, RegistrationResult } from './lib/interfaces';
+export { X509ProvisioningTransport, RegistrationRequest, RegistrationResult } from './lib/interfaces';
 export { PollingStateMachine } from './lib/polling_state_machine';
 export { ProvisioningDeviceClient } from './lib/client';
 export { ProvisioningDeviceConstants } from './lib/constants';
+export { translateError } from './lib/dps_errors';
```
pierreca: >dps_errors [](start = 38, length = 10)

provisioning_errors?

provisioning/device/devdoc/dps_error_requirements.md
```diff
@@ -0,0 +1,31 @@
+# azure-iot-provisioning-device.translateError Requirements
+
+## Overview
+`translateError` is a method that translates DPS errors into Azure IoT Hub errors, effectively abstracting the error that is returned to the SDK user of from the transport layer.
```
pierreca: >DPS [](start = 45, length = 3)

please make sure not to use the DPS acronym

provisioning/device/devdoc/dps_error_requirements.md
```diff
@@ -0,0 +1,31 @@
...
+** SRS_NODE_DPS_ERRORS_18_001: [** `translateError` shall accept 4 arguments:
+ - A custom error message to give context to the user.
+ - the status code that initiated the error
+ - the response body
+ - the response object itself **]**
+
+
+** SRS_NODE_DPS_ERRORS_18_008: [** Any error object returned by `translateError` shall inherit from the generic `Error` Javascript object and have 3 properties:
+ - `response` shall contain the `IncomingMessage` object returned by the HTTP layer.
+ - `reponseBody` shall contain the content of the HTTP response.
```
pierreca: those may be empty depending on the error

provisioning/device/devdoc/dps_error_requirements.md
```diff
@@ -0,0 +1,31 @@
...
+ - `message` shall contain a human-readable error message **]**
+
+** SRS_NODE_DPS_ERRORS_18_002: [** `translateError` shall return an `ArgumentError` if the status code is `400`. **]**
+
+** SRS_NODE_DPS_ERRORS_18_003: [** `translateError` shall return an `UnauthorizedError` if the status code is `401`. **]**
+
+** SRS_NODE_DPS_ERRORS_18_004: [** `translateError` shall return an `DeviceNotFoundError` if the status code is `404`. **]**
+
+** SRS_NODE_DPS_ERRORS_18_005: [** `translateError` shall return an `IotHubQuotaExceededError` if the status code is `429`. **]**
```
pierreca: >IotHubQuotaExceededError [](start = 69, length = 24)

mhhh is that the same quota as the IoT Hub quota? might be misleading? do we need a ProvisioningQuotaExceededError or RegistrationQuotaExceededError?

provisioning/device/devdoc/provisioning_client_requirements.md
```diff
@@ -0,0 +1,29 @@
...
+
+```
+
+## Public API
+
+### create(transport, securityClient)
+
+The `create` method is used to retrieve a client object that can be used to register with the Device Provisioning Service
+
+**SRS_PROVISIONING_CLIENT_18_001: [** If `securityClient` supports x509 security and the `transport` supports x509 authentication, then `crate` shall
```
pierreca: >crate [](start = 137, length = 5)

typo (here and below); crEate

pierreca: >supports x509 security and the `transport` supports x509 authentication [](start = 58, length = 71)

Is there a way to clarify how we assert support?

provisioning/device/devdoc/x509_registration_requirements.md
```diff
@@ -0,0 +1,28 @@
...
+
+```
+
+## Public API
+
+### register()
+
+The `register` method completes the authentication and registration flow for the user.
+
+**SRS_NODE_DPS_X509_REGISTRATION_18_001: [** `register` shall acquite the X509 certificate from the `X509SecurityClient` object. **]**
```
pierreca: >acquite [](start = 62, length = 7)

typo: acquiRe

pierreca: using which method? i think it's a good practice to explicitely tell what API we're supposed to use (like the next requirement does)

---
In reply to: [154875649](https://github.com/Azure/azure-iot-sdk-node/pull/169#discussion_r154875649) [](ancestors = 154875649)

provisioning/device/devdoc/x509_registration_requirements.md
```diff
@@ -0,0 +1,28 @@
...
+
+**SRS_NODE_DPS_X509_REGISTRATION_18_002: [** `register` shall call `registerX509` on the transport object and call it's callback with the result of the transport operation. **]**
+
+### cancel()
+
+The `cancel` method cancels a registration flow that is in progress.
+
+**SRS_NODE_DPS_X509_REGISTRATION_18_003: [** `cancel` shall call `endSession` on the transport object. **]**
```
pierreca: >endSession [](start = 66, length = 10)

nit/naming: I'd be in favor of calling this "cancel" across all objects. I'm not sure the concept of session is valid here (it may be but we're not calling this a session anywhere anyway)

provisioning/device/src/client.ts
```diff
@@ -14,14 +15,30 @@ export class ProvisioningDeviceClient {
    * @param securityClient: Security client object which can provide access to the necessary secrets and/or encryption functions
    */
   static create(transport: X509ProvisioningTransport | TpmProvisioningTransport, securityClient: X509SecurityClient | TpmSecurityClient): RegistrationClient {
-    if ((securityClient as X509SecurityClient).getCertificate && (securityClient as X509SecurityClient).getCertificateChain) {
-      if (!((transport as X509ProvisioningTransport).registerX509)) {
-        throw new errors.InvalidOperationError('Transport does not support X509 authentication');
-      } else {
+    let isX509Security: boolean = ((securityClient as X509SecurityClient).getCertificate !== undefined);
```
pierreca: >let [](start = 4, length = 3)

const?

provisioning/device/src/client.ts
```diff
@@ -14,14 +15,30 @@ export class ProvisioningDeviceClient {
...
+        /* Codes_SRS_PROVISIONING_CLIENT_18_002: [ If `securityClient` supports x509 security and the `transport` does not support x509 authentication, then `crate` shall throw a `ArgumentError` exepction. ] */
+        throw new errors.ArgumentError('Transport does not support X509 authentication');
+      }
+    } else if (isTpmSecurity) {
+      if (isTpmTransport) {
+        /* Codes_SRS_PROVISIONING_CLIENT_18_003: [ If `securityClient` supports TPM security and the `transport` supports TPM authentication, then `create` shall return a `TpmRegistration` object. ] */
+        return new TpmRegistration(transport as TpmProvisioningTransport, securityClient as TpmSecurityClient);
+      } else {
+        /* Codes_SRS_PROVISIONING_CLIENT_18_004: [ If `securityClient` supports TPM security and the `transport` does not support TPM authentication, then `crate` shall throw a `ArgumentError` exepction. ] */
+        throw new errors.ArgumentError('Transport does not support X509 authentication');
```
pierreca: >X509 [](start = 67, length = 4)

TPM

provisioning/device/src/dps_errors.ts
```diff
@@ -0,0 +1,62 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+
+import { errors } from 'azure-iot-common';
+
+/**
+ * @private
+ */
+export class DpsError extends Error {
```
pierreca: >DpsError [](start = 13, length = 8)

ProvisioningError?
Also, response and responseBody are very HTTP-ish in the SDK - I'd be curious to see how that works out for other transports.

FWIW we may just get away with a single `transportObj` property like we do in other parts of the IoT Hub SDK. and maybe even make it private so that users don't take a dependency on it (or at least make it clear it's a debug thing that may change)

provisioning/device/src/interfaces.ts
```diff
@@ -20,14 +20,34 @@ export interface ProvisioningTransportOptions {
...
+   */
+  provisioningHost: string;
+
+  /**
+   * ID scope for the provisioning instance
+   */
+  idScope: string;
+
+  /**
+   * true to foce re-regristation
```
pierreca: >regristation [](start = 21, length = 12)

typo (registration)

provisioning/device/src/interfaces.ts
```diff
@@ -90,20 +109,21 @@ export interface RegistrationClient {
...
    * Cancel the registration process if it is in progress.
    */
   cancel(callback: (err?: Error) => void): void;
 }
 
 /**
- * Device configuration returned when registration using TPM is complete
+ * Information passed between client and transport during Tpm registration
  */
 export interface TpmRegistrationInfo {
```
pierreca: >TpmRegistrationInfo [](start = 17, length = 19)

registrationId is going to be duplicated in this structure - let's remove it from TpmRegistrationInfo alltogether

pierreca:
>           _onEnter: (callback, err, body, result) => {

feels kinda weird to have the callback first now. i know, we've been over this before - i've been trying to put all my callbacks at the end because I thought this is what others wanted. maybe a subject for a future refactoring?  let's just get this convention fixed once and for all everywhere (whatever we end up landing on).

---
Refers to: provisioning/device/src/polling_state_machine.ts:29 in 18eec33. [](commit_id = 18eec336e91a2b44a554649ebf343aaa4b69b3bd, deletion_comment = False)

provisioning/device/src/polling_state_machine.ts
```diff
@@ -32,8 +32,8 @@ export class  PollingStateMachine extends EventEmitter {
               callback(err, body, result);
             }
           },
-          register: (callback, registrationId, authorization, requestBody, forceRegistration) => {
-            this._fsm.transition('sendingRegistrationRequest', callback, registrationId, authorization, requestBody, forceRegistration);
+          register: (callback, request, authorization, requestBody) => {
```
pierreca: >authorization [](start = 40, length = 13)

I believe it's actually authentication. we're using this to prove our identity not determine our privileges imho.

provisioning/device/src/polling_state_machine.ts
```diff
@@ -48,18 +48,18 @@ export class  PollingStateMachine extends EventEmitter {
...
+            this._fsm.transition('sendingRegistrationRequest', callback, request, authorization, requestBody);
           },
         },
         sendingRegistrationRequest: {
-          _onEnter: (callback, registrationId, authorization, requestBody, forceRegistration) => {
+          _onEnter: (callback, request, authorization, requestBody) => {
             /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_012: [ `register` shall call `PollingTransportHandlers.registrationRequest`. ] */
             this._currentOperationCallback = callback;
-            this._transport.registrationRequest(registrationId, authorization, requestBody, forceRegistration, (err, body, result, pollingInterval) => {
+            this._transport.registrationRequest(request, authorization, requestBody, (err, body, result, pollingInterval) => {
```
pierreca: >authorization [](start = 57, length = 13)

I don't know that we need to carry the auth parameter around everywhere. feels like a transport class responsibility to know when to use it. here we're just passing it around without really using it. would also help with the form of the API for AMQP and MQTT I believe.

provisioning/device/src/polling_state_machine.ts
```diff
@@ -192,20 +185,15 @@ export class  PollingStateMachine extends EventEmitter {
     });
   }
 
-  register(registrationId: string, authorization: ProvisioningAuthentication, requestBody: any, forceRegistration: boolean, callback: (err?: Error, responseBody?: any, result?: any) => void): void {
-    debug('register called for registrationId "' + registrationId + '"');
-    this._fsm.handle('register', callback, registrationId, authorization, requestBody, forceRegistration);
+  register(request: RegistrationRequest, authorization: ProvisioningAuthentication, requestBody: any, callback: (err?: Error, responseBody?: any, result?: any) => void): void {
```
pierreca: >authorization [](start = 41, length = 13)

yeah seeing it here with the ProvisioningAuthentication type i think it'd be good to make a pass over the code and make sure we use authentication everywhere instead of authorization.

provisioning/device/src/tpm_registration.ts
```diff
@@ -214,7 +213,7 @@ export class TpmRegistration extends EventEmitter implements RegistrationClient
     - `expiryTimeUtc` being the number of seconds since Epoch + a delay during which the initial sas token should be valid (1 hour by default).
     ]*/
     const expiryTimeUtc = Date.now() / 1000 + 3600; // 1 hour from now.
-    const audience = encodeURIComponent(this._idScope + '/registrations/' + registrationId);
+    const audience = encodeURIComponent(registrationInfo.request.idScope + '/registrations/' + registrationInfo.registrationId);
```
pierreca: >registrationId [](start = 112, length = 14)

.request.registrationId - let's not duplicate this property

provisioning/device/test/_x509_registration_test.js
```diff
@@ -0,0 +1,96 @@
...
+
+var fakeResponse = {
+  assignedHub: 'fakeHub',
+  deviceId: 'fakeDeviceId'
+};
+
+describe('X509Registration', function () {
+  describe('#register', function () {
+
+    /* Tests_SRS_NODE_DPS_X509_REGISTRATION_18_001: [ `register` shall acquite the X509 certificate from the `X509SecurityClient` object. ] */
```
pierreca: >acquite [](start = 71, length = 7)

typo (acquire)

provisioning/device/src/tpm_registration.ts
```diff
@@ -189,8 +188,8 @@ export class TpmRegistration extends EventEmitter implements RegistrationClient
     });
   }
 
-  register(registrationId: string, forceRegistration: boolean, callback: (err?: Error, result?: any) => void): void {
-    this._fsm.handle('register', callback);
+  register(request: RegistrationRequest, callback: (err?: Error, result?: any) => void): void {
```
pierreca: >RegistrationRequest [](start = 20, length = 19)

as much as I like the idea of packaging these together internally, I'm not sure other SDKs have such an object in their public API. we need to make sure we present the same set of parameters in the same API calls as much as possible. can you double-check what the others do and make sure we are aligned as far as public APIs are concerned?

pierreca:
> export class Http  extends EventEmitter implements X509ProvisioningTransport {

after having thought more about this issue with AMQP I think i'd suggest 2 changes:
- make this object responsible for the RestApiClient object lifecycle
- move the HttpPollingHandlers back in here and pass "this" to the PollingStateMachine

we can talk offline about why - the gist of it is that with other protocols this object will manage the connection and potentially the links/topic subscriptions. The handlers may need access to those to function. so the Amqp and AmqpPollingHandlers must be part of the same object/context - and it'd be great if HTTP was aligned with that.

---
Refers to: provisioning/transport/http/src/http_transport.ts:15 in 18eec33. [](commit_id = 18eec336e91a2b44a554649ebf343aaa4b69b3bd, deletion_comment = False)

build/build_parallel/build_parallel.js
BertKleewein: Yes,  This just makes tracking down errors easier.  If it's building 4 things in parallel and 1 fails, it finishes the other 3.  Then the summary only lists the single failure.  If this gets set to false, then it cancels the other 3 and they're all listed as failed.  (Jenkins does the same thing on failure and it makes tracking down failures really annoying.

---
In reply to: [154873753](https://github.com/Azure/azure-iot-sdk-node/pull/169#discussion_r154873753) [](ancestors = 154873753)

BertKleewein: The change below affects the next set of parallel tasks.  If config.task.continueOnError is set, it will continue with the next set of parallel tasks on failure.

---
In reply to: [155023758](https://github.com/Azure/azure-iot-sdk-node/pull/169#discussion_r155023758) [](ancestors = 155023758,154873753)

provisioning/device/devdoc/dps_error_requirements.md
BertKleewein: maybe, but I think we're splitting hairs here.

---
In reply to: [154874443](https://github.com/Azure/azure-iot-sdk-node/pull/169#discussion_r154874443) [](ancestors = 154874443)

common/core/src/errors.ts
```diff
@@ -430,3 +430,16 @@ export class OperationCancelledError extends Error {
...
 
+/**
+ * Error thrown when a DPS registration operation fails
+ *
+ * @augments {Error}
+ */
+export class DeviceRegistrationFailedError extends Error {
+  constructor(message?: string) {
+    super(message);
+    this.name = 'ProvisioningRegistrationFailedError';
```
pierreca: >Provisioning [](start = 17, length = 12)

should be "DeviceRegistrationFailedError"

PR #174, opened 2017-12-07
[Provisioning - Device SDK] Changes to polling state machine architecture and X509 amqp registration

pierreca:
The dependency tree between the registration flow state machine, transport state machine and polling state machine was wonky. Bert fixed that.

On top of that, I am adding support for x509 registration over AMQP


provisioning/transport/amqp/src/amqp.ts
```diff
@@ -2,14 +2,332 @@
...
+
+    const amqpErrorListener = (err) => this._amqpStateMachine.handle('amqpError', err);
+
+    const responseHandler = (msg) => {
+      debug('got message with correlationId: ' + msg.correlationId);
+      /*Codes_SRS_NODE_PROVISIONING_AMQP_16_007: [The `registrationRequest` method shall call its callback with a `RegistrationResult` object parsed from the body of the response message which `correlationId` matches the `correlationId` of the request message sent on the sender link.]*/
+      /*Codes_SRS_NODE_PROVISIONING_AMQP_16_017: [The `queryOperationStatus` method shall call its callback with a `RegistrationResult` object parsed from the body of the response message which `correlationId` matches the `correlationId` of the request message sent on the sender link.]*/
+      const registrationResult = JSON.parse(msg.data);
+      if (this._operations[msg.correlationId]) {
+        this._operations[msg.correlationId](null, registrationResult, msg, this._config.pollingInterval);
```
BertKleewein: >this._operations[msg.correlationId](null, registrationResult, msg, this._config.pollingInterval); [](start = 8, length = 97)

need to remove this handler from the object after using it.

provisioning/transport/amqp/src/amqp.ts
```diff
@@ -2,14 +2,332 @@
...
+            requestMessage.properties = {};
+            requestMessage.applicationProperties[MessagePropertyNames.OperationType] = DeviceOperations.Register;
+            requestMessage.applicationProperties[MessagePropertyNames.ForceRegistration] = !!request.forceRegistration;
+            requestMessage.properties.correlationId = uuid.v4();
+
+            debug('initial registration request: ' + JSON.stringify(requestMessage));
+            this._operations[requestMessage.properties.correlationId] = callback;
+            this._senderLink.send(requestMessage, (err) => {
+              if (err) {
+                this._operations[requestMessage.properties.correlationId] = undefined;
```
BertKleewein: >this._operations[requestMessage.properties.correlationId] = undefined; [](start = 16, length = 70)

IIRC,. I think you want delete this._operations[id].  Otherwise the member is still in the object, but it's set to undefined.

pierreca: that is correct - delete has a slight performance hit but we might not care. it doesn't make a difference for how we use it though. if you think it is cleaner i'm happy to oblige.

---
In reply to: [155602005](https://github.com/Azure/azure-iot-sdk-node/pull/174#discussion_r155602005) [](ancestors = 155602005)

pierreca:
closed as overriden by #176

PR #176, opened 2017-12-07
combined polling handler changes & amqp transport work

BertKleewein:
This is my polling handler changes combined with Pierre's AMQP work.  This code, as it exists in this branch,  passes the gate, including AMQP E2E


provisioning/device/devdoc/x509_registration_requirements.md
```diff
@@ -19,7 +19,9 @@ The `register` method completes the authentication and registration flow for the
 
 **SRS_NODE_DPS_X509_REGISTRATION_18_001: [** `register` shall call `getCertificate` on the security object to acquire the X509 certificate. **]**
 
-**SRS_NODE_DPS_X509_REGISTRATION_18_002: [** `register` shall call `registerX509` on the transport object and call it's callback with the result of the transport operation. **]**
+***SRS_NODE_DPS_X509_REGISTRATION_18_004: [** `register` shall pass the certificate into the `setAuthentication` method on the transport **]**
+
+**SRS_NODE_DPS_X509_REGISTRATION_18_002: [** `register` shall call `register` on the pollingStateMachine and call `callback` with the result. **]**
```
pierreca: >SRS_NODE_DPS_X509_REGISTRATION_18_002 [](start = 2, length = 37)

I think we're missing a requirement about what happens when registration fails (calls the callback with an error)

provisioning/device/src/interfaces.ts
```diff
@@ -140,10 +134,9 @@ export interface TpmRegistrationResult extends RegistrationResult {
 /**
  * @private
  */
-export interface TpmProvisioningTransport {
+export interface TpmProvisioningTransport extends PollingTransportHandlers {
+  setSasToken(sasToken: string): void;
```
pierreca: >setSasToken [](start = 2, length = 11)

I think this would be the ek/srk pair instead - i'd advocate to remove that until we actually tackle TPM implementation

provisioning/device/src/interfaces.ts
```diff
@@ -67,10 +62,18 @@ export interface RegistrationResult {
 /**
  * @private
  */
-export interface X509ProvisioningTransport {
-  setTransportOptions(options: ProvisioningTransportOptions): void;
+export interface PollingTransportHandlers {
```
pierreca: >PollingTransportHandlers [](start = 17, length = 24)

PollingTransport (remove Handlers)

provisioning/device/src/interfaces.ts
```diff
@@ -67,10 +62,18 @@ export interface RegistrationResult {
...
+  registrationRequest(request: RegistrationRequest, callback: (err?: Error, body?: any, result?: any, pollingInterval?: number) => void): void;
+  queryOperationStatus(request: RegistrationRequest, operationId: string, callback: (err?: Error, body?: any, result?: any, pollingInterval?: number) => void): void;
   cancel(callback: (err?: Error) => void): void;
-  registerX509(request: RegistrationRequest, auth: X509, callback: (err?: Error, registrationResult?: RegistrationResult, body?: any, result?: any) => void): void;
```
pierreca: huh - where are these going? this is still the public interface for the transport, the register function in the PollingStateMachine is still hidden under it right?

pierreca: never mind - RegistrationClient is the public interface. my bad.

---
In reply to: [155605404](https://github.com/Azure/azure-iot-sdk-node/pull/176#discussion_r155605404) [](ancestors = 155605404)

provisioning/device/src/tpm_registration.ts
```diff
@@ -1,20 +1,23 @@
 import { EventEmitter } from 'events';
 import * as machina from 'machina';
 import * as dbg from 'debug';
-const debug = dbg('azure-iot-provisioning-device:TpmRegistration');
+const debug = dbg('azure-device-provisioning:TPM');
```
pierreca: >TPM [](start = 45, length = 3)

Nope. the convention is to use the class name here. I'd rather see this reverted.

provisioning/device/src/tpm_registration.ts
```diff
@@ -122,7 +125,8 @@ export class TpmRegistration extends EventEmitter implements RegistrationClient
             - `sasToken`: the SAS token generated according to `SRS_NODE_DPS_TPM_REGISTRATION_16_006`
             - `registrationInfo`: an object with the following properties `endorsementKey`, `storageRootKey`, `registrationId` and their previously set values.
             - a callback that will handle an optional error and a `result` object containing the IoT hub name, device id and symmetric key for this device.]*/
-            this._transport.register(registrationInfo, sasToken, (err, result) => {
+            this._transport.setSasToken(sasToken);
```
pierreca: >setSasToken [](start = 28, length = 11)

oh I see. well. since we have a getAuthenticationChallenge I'd suggest we name this "sendAuthenticationChallengeResponse" or something along those lines. Even though it looks like a SAS token i'm not sure that it's one (tokens are usually associated with CBS)

provisioning/device/src/x509_registration.ts
```diff
@@ -43,16 +46,15 @@ export class X509Registration implements RegistrationClient {
           provisioningHost: this._provisioningHost,
           idScope: this._idScope
         };
+        /* Codes_SRS_NODE_DPS_X509_REGISTRATION_18_004: [ `register` shall pass the certificate into the `setAuthentication` method on the transport ] */
+        this._transport.setAuthentication(cert);
         /* Codes_SRS_NODE_DPS_X509_REGISTRATION_18_002: [ `register` shall call `registerX509` on the transport object and call it's callback with the result of the transport operation. ] */
-        this._transport.registerX509(request, cert, (err, result, body) => {
+        this._pollingStateMachine.register(request, (err, response) => {
```
pierreca: >response [](start = 58, length = 8)

Historically the convention for callback parameters in the device SDK is the following:
- result is an object that is independent of the transport
- response in the HTTP object we receive from the server (after we make our request -and it's used only the the context of HTTP)
- responseBody if the actual body of the HTTP response

all callback are usually (err: Error, result: SomeType) => void and sometimes (err: Error, result: SomeType, response: IncomingMessage) => void.

I would very much like a pass for consistency in the provisioning SDK to avoid confusion in the future when we go back to these pieces of code to debug them

provisioning/device/test/_x509_registration_test.js
```diff
@@ -26,25 +26,28 @@ describe('X509Registration', function () {
...
-        registerX509: sinon.spy(function(request, auth, callback) { callback(null, fakeResponse); }),
-        getRegistrationId: function() { return fakeRegistrationId; }
+        setAuthentication: sinon.spy(),
       };
       var security = {
         getCertificate: sinon.spy(function(callback) { callback(null, fakeX509Cert); }),
         getRegistrationId: function() { return fakeRegistrationId; }
       };
       var clientObj = new X509Registration(fakeProvisioningHost, fakeIdScope, transport, security);
+      clientObj._pollingStateMachine.register = sinon.spy(function(request, callback) { callback(null, { registrationState: fakeResponse }); });
```
pierreca: >sinon.spy [](start = 48, length = 9)

in other tests across the SDK we use:
sinon.stub().callsFake(function () {})
and if it's just to call a callback with a default result:
sinon.stub().callsArgWith(<callbackArgIndex>, <error>, <result>)

provisioning/device/test/_x509_registration_test.js
```diff
@@ -26,25 +26,28 @@ describe('X509Registration', function () {
...
       clientObj.register(function(err, response) {
         assert.isNotOk(err);
         assert.strictEqual(response, fakeResponse);
+        assert(clientObj._pollingStateMachine.register.calledOnce);
+        assert.strictEqual(clientObj._pollingStateMachine.register.getCall(0).args[0].provisioningHost, fakeProvisioningHost);
```
pierreca: >getCall(0) [](start = 67, length = 10)

register.firstCall.args

provisioning/transport/http/test/_http_test.js
```diff
@@ -45,7 +46,25 @@ describe('Http', function () {
...
+          callback(null, body.registrationState, body);
+        }
+      });
+    };
+
+    pollingStateMachine.setTransportOptions = function(options) {
+      http.setTransportOptions(options);
+    }
+
+    return pollingStateMachine;
```
pierreca: I have a couple of worries for this thing:
1. it creates something called pollingStateMachine and returns it when the function is called makeNewTransport and those are separate concepts (maybe what is meant here is PollingTransport?)

2. it "hides" 2 API calls behind one that looks like an old one.

I'm afraid a couple months from now when we return to this for a bug or something it's going to be very confusing. To be honest it's a little confusing now already. I would much rather see the complete API flow in the tests, even if it's a little repetitive.

BertKleewein: I agree.  I'm going to create a TFS task for this and a few other items from this review

---
In reply to: [155611534](https://github.com/Azure/azure-iot-sdk-node/pull/176#discussion_r155611534) [](ancestors = 155611534)

provisioning/device/src/x509_registration.ts
BertKleewein: Agree on the need for consistency.  Adding it to my TFS task for later.

---
In reply to: [155607744](https://github.com/Azure/azure-iot-sdk-node/pull/176#discussion_r155607744) [](ancestors = 155607744)

PR #179, opened 2017-12-14
Authentication providers for the device client

pierreca:
# Description of the problem
In preparation for the DPS integration with the IoT Hub device SDK we need to introduce the concept of AuthenticationProvider: a set of classes that obey the same interface and are in charge of acquiring and updating the credentials that the client and transports need to perform

This first PR (out of a series) introduces the interface as well as the first 3 authentication providers that cover legacy authentication methods, using shared access key, shared access signature and x509 certificates. 

The future PRs will include changes to the client and transports to take advantage of these new objects as well as the TPM and X509 authentication providers working with their respective security clients coming from the DPS world.

No breaking changes in here, only new APIs that are currently not used by the client or transports.

### DIFF
--- a/common/core/common.d.ts
+++ b/common/core/common.d.ts
@@ -13,7 +13,7 @@ export { Message }
 export { SharedAccessSignature } from './lib/shared_access_signature';
 export { RetryOperation } from './lib/retry_operation';
 export { RetryPolicy, NoRetry, ExponentialBackOffWithJitter } from './lib/retry_policy';
-export { TokenAuthenticationProvider, AuthenticationProvider, AuthenticationType } from './lib/authentication_provider';
+export { AuthenticationProvider, AuthenticationType } from './lib/authentication_provider';
 
 export interface Receiver extends EventEmitter {
     on(type: 'message', func: (msg: Message) => void): this;
@@ -22,5 +22,4 @@ export interface Receiver extends EventEmitter {
     on(type: string, func: Function): this;
 }
 
-export { DeviceCredentials, X509 } from './lib/authorization';
-export { DeviceCredentials as TransportConfig } from './lib/authorization';
+export { TransportConfig, X509 } from './lib/authorization';
--- a/common/core/common.d.ts
+++ b/common/core/common.d.ts
@@ -13,16 +13,7 @@ export { Message }
 export { SharedAccessSignature } from './lib/shared_access_signature';
 export { RetryOperation } from './lib/retry_operation';
 export { RetryPolicy, NoRetry, ExponentialBackOffWithJitter } from './lib/retry_policy';
-
-// Typescript only, used by other modules in azure-iot-sdk
-export interface X509 {
-    // https://nodejs.org/api/tls.html#tls_tls_connect_options_callback
-    cert?: string | string[] | Buffer | Buffer[];
-    key?: string | Buffer;
-    passphrase?: string;
-    certFile?: string;
-    keyFile?: string;
-}
+export { TokenAuthenticationProvider, AuthenticationProvider, AuthenticationType } from './lib/authentication_provider';
 
 export interface Receiver extends EventEmitter {
     on(type: 'message', func: (msg: Message) => void): this;
@@ -31,9 +22,5 @@ export interface Receiver extends EventEmitter {
     on(type: string, func: Function): this;
 }
 
-export interface TransportConfig {
-    host: string;
-    deviceId: string;
-    sharedAccessSignature?: string;
-    x509?: X509;
-}
+export { DeviceCredentials, X509 } from './lib/authorization';
+export { DeviceCredentials as TransportConfig } from './lib/authorization';
--- a/device/core/devdoc/sak_authentication_provider_requirements.md
+++ b/device/core/devdoc/sak_authentication_provider_requirements.md
@@ -15,32 +15,28 @@ sakAuthProvider.getCredentials(function (err, credentials) {
 });
 
 // to monitor for new tokens:
-sakAuthProvider.on('newTokenAvailable', function () {
-  sakAuthProvider.getCredentials(function (err, credentials) {
-    // do something with the credentials
-  });
+sakAuthProvider.on('newTokenAvailable', function (credentials) {
+  // do something with the credentials
 });
 ```
 
 # Public API
 
-# constructor(credentials: DeviceCredentials, tokenValidTimeInSeconds?, tokenRenewalMarginInSeconds?: number)
+# constructor(credentials: TransportConfig, tokenValidTimeInSeconds?, tokenRenewalMarginInSeconds?: number)
 
 **SRS_NODE_SAK_AUTH_PROVIDER_16_001: [** The `constructor` shall create the initial token value using the `credentials` parameter. **]**
 
 **SRS_NODE_SAK_AUTH_PROVIDER_16_002: [** The `constructor` shall start a timer that will automatically renew the token every (`tokenValidTimeInSeconds` - `tokenRenewalMarginInSeconds`) seconds if specified, or 45 minutes by default. **]**
 
-## getDeviceCredentials(callback: (err: Error, credentials: DeviceCredentials) => void): void
+**SRS_NODE_SAK_AUTH_PROVIDER_16_011: [** The `constructor` shall throw an `ArgumentError` if the `tokenRenewalMarginInSeconds` is less than or equal `tokenValidTimeInSeconds`. **]**
 
-**SRS_NODE_SAK_AUTH_PROVIDER_16_003: [** The `getDeviceCredentials` should call its callback with a `null` first parameter and a `DeviceCredentials` object as a second parameter, containing the latest valid token it generated. **]**
+## getDeviceCredentials(callback: (err: Error, credentials: TransportConfig) => void): void
 
-## updateSharedAccessSignature(sharedAccessSignature: string): void
-
-**SRS_NODE_SAK_AUTH_PROVIDER_16_004: [** The `updateSharedAccessSignature` method shall save the `sharedAccessSignature` passed as a parameter and return it when `getDeviceCredentials` is called, until it gets renewed. **]**
+**SRS_NODE_SAK_AUTH_PROVIDER_16_003: [** The `getDeviceCredentials` should call its callback with a `null` first parameter and a `TransportConfig` object as a second parameter, containing the latest valid token it generated. **]**
 
 ## newTokenAvailable event
 
-**SRS_NODE_SAK_AUTH_PROVIDER_16_005: [** Every time a new token is created, the `newTokenAvailable` event shall be fired with no arguments. **]**
+**SRS_NODE_SAK_AUTH_PROVIDER_16_005: [** Every time a new token is created, the `newTokenAvailable` event shall be fired with the updated credentials. **]**
 
 ## fromConnectionString(connectionString: string, tokenValidTimeInSeconds?, tokenRenewalMarginInSeconds?: number): SharedAccessKeyAuthenticationProvider [static]
 
--- a/device/core/devdoc/sak_authentication_provider_requirements.md
+++ b/device/core/devdoc/sak_authentication_provider_requirements.md
@@ -1,0 +1,64 @@
+# SharedAccessKeyAuthenticationProvider Requirements
+
+# Overview
+
+The `SharedAccessKeyAuthenticationProvider` class is used to provide credentials to transports when the user wants the device to authenticate using security tokens, typically generated from shared access key that is passed as part of a connection string.
+
+The `SharedAccessKeyAuthenticationProvider` class implements the `TokenAuthenticationProvider` interface from `azure-iot-common`.
+
+# Example Usage
+```js
+var sakAuthProvider = SharedAccessKeyAuthenticationProvider.fromConnectionString('<connectionstring>');
+// whenever a new token is needed:
+sakAuthProvider.getCredentials(function (err, credentials) {
+  // do something with the credentials
+});
+
+// to monitor for new tokens:
+sakAuthProvider.on('newTokenAvailable', function () {
+  sakAuthProvider.getCredentials(function (err, credentials) {
+    // do something with the credentials
+  });
+});
+```
+
+# Public API
+
+# constructor(credentials: DeviceCredentials, tokenValidTimeInSeconds?, tokenRenewalMarginInSeconds?: number)
+
+**SRS_NODE_SAK_AUTH_PROVIDER_16_001: [** The `constructor` shall create the initial token value using the `credentials` parameter. **]**
+
+**SRS_NODE_SAK_AUTH_PROVIDER_16_002: [** The `constructor` shall start a timer that will automatically renew the token every (`tokenValidTimeInSeconds` - `tokenRenewalMarginInSeconds`) seconds if specified, or 45 minutes by default. **]**
+
+## getDeviceCredentials(callback: (err: Error, credentials: DeviceCredentials) => void): void
+
+**SRS_NODE_SAK_AUTH_PROVIDER_16_003: [** The `getDeviceCredentials` should call its callback with a `null` first parameter and a `DeviceCredentials` object as a second parameter, containing the latest valid token it generated. **]**
+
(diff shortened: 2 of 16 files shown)


common/core/common.d.ts
```diff
@@ -31,9 +22,5 @@ export interface Receiver extends EventEmitter {
...
 
-export interface TransportConfig {
-    host: string;
-    deviceId: string;
-    sharedAccessSignature?: string;
-    x509?: X509;
-}
\ No newline at end of file
+export { DeviceCredentials, X509 } from './lib/authorization';
+export { DeviceCredentials as TransportConfig } from './lib/authorization';
```
BertKleewein: >as TransportConfig [](start = 27, length = 18)

Ouch.   I'm not a big fan of making aliases like this -- it looks sloppy and it adds cognitive overhead.  Are you planning to remove this alias?  I know we have a few different objects floating around called TransportConfig and they're almost the same but not quite.  (IIRC, one has sharedAccessSignature as a string and a different one has it as a SharedAccessSignature object.)  Conceptually, I think DeviceCredenentials should be a member of TransportConfig.   We should consider unifying all the TransportConfig objects and making DeviceCredentials a member, maybe as an MQ task.

device/core/devdoc/sak_authentication_provider_requirements.md
```diff
@@ -0,0 +1,64 @@
...
+```js
+var sakAuthProvider = SharedAccessKeyAuthenticationProvider.fromConnectionString('<connectionstring>');
+// whenever a new token is needed:
+sakAuthProvider.getCredentials(function (err, credentials) {
+  // do something with the credentials
+});
+
+// to monitor for new tokens:
+sakAuthProvider.on('newTokenAvailable', function () {
+  sakAuthProvider.getCredentials(function (err, credentials) {
```
BertKleewein: >getCredentials [](start = 18, length = 14)

Can the credentials be passed via the emit call?

device/core/src/sak_authentication_provider.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  /**
+   * Gets the current device credentials that are valid and were generated from the connectiom string.
+   *
+   * @param callback function that will be called with either an error or a set of device credentials that can be used to authenticate with the IoT hub.
+   */
+  getDeviceCredentials(callback: (err: Error, credentials: DeviceCredentials) => void): void {
+    if (this._shouldRenewToken()) {
+      this._renewToken();
```
BertKleewein: >_renewToken [](start = 11, length = 11)

(unlikely) possibility for stack overflow here.  _renewToken can fire newTokenAvailable, and newTokenAvailable handler needs to call getDeviceCredentials.

device/core/src/sak_authentication_provider.ts
```diff
@@ -0,0 +1,130 @@
...
+    callback(null, this._credentials);
+  }
+
+  /**
+   * Sets the current security token for the device. Note that it will be overwritten by the timer that is running in the background.
+   * If you're looking for an AuthenticationProvider where you maintain the lifecycle of the token yourself, you should look at the `SharedAccessSignatureAuthorizationProvider` class.
+   *
+   * @param sharedAccessSignature shared access signature that should be stored.
+   */
+  updateSharedAccessSignature(sharedAccessSignature: string): void {
```
BertKleewein: >updateSharedAccessSignature [](start = 2, length = 27)

suggest renaming this function to shootYourselfInTheFoot().  Or maybe it should be removed.  If an interface definition requires it to be present, would it be better to throw?

device/core/src/sak_authentication_provider.ts
```diff
@@ -0,0 +1,130 @@
...
+    resourceUri: <IoT hub host>/devices/<deviceId>
+    keyName: the `SharedAccessKeyName` parameter of the connection string or `null`
+    key: the `SharedAccessKey` parameter of the connection string
+    expiry: the expiration time of the token, which is now + the token validity time, formatted as the number of seconds since Epoch (Jan 1st, 1970, 00:00 UTC).
+    ```]*/
+    const resourceUri = encodeUriComponentStrict(this._credentials.host + '/devices/' + this._credentials.deviceId);
+    const sas = SharedAccessSignature.create(resourceUri, this._credentials.sharedAccessKeyName, this._credentials.sharedAccessKey, newExpiry);
+    this._credentials.sharedAccessSignature = sas.toString();
+
+    const nextRenewalTimeout = (this._tokenValidTimeInSeconds - this._tokenRenewalMarginInSeconds) * 1000;
```
BertKleewein: >this._tokenValidTimeInSeconds - this._tokenRenewalMarginInSeconds [](start = 32, length = 65)

Suggest adding a throw  to the constructor if the margin time is greater than the valid time.

device/core/src/sak_authentication_provider.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    const cs = ConnectionString.parse(connectionString);
+    if (!cs.SharedAccessKey) {
+      /*Codes_SRS_NODE_SAK_AUTH_PROVIDER_16_007: [The `fromConnectionString` method shall throw an `errors.ArgumentError` if the `connectionString` does not have a SharedAccessKey parameter.]*/
+      throw new errors.ArgumentError('Connection string is missing a shared access key');
+    }
+
+      /*Codes_SRS_NODE_SAK_AUTH_PROVIDER_16_008: [The `fromConnectionString` method shall extract the credentials from the `connectionString` argument and create a new `SharedAccessKeyAuthenticationProvider` that uses these credentials to generate security tokens.]*/
+      const credentials: DeviceCredentials = {
```
BertKleewein: >const [](start = 6, length = 5)

indent

device/core/src/sak_authentication_provider.ts
```diff
@@ -0,0 +1,130 @@
...
+   * @param tokenValidTimeInSeconds       [optional] The number of seconds for which a token is supposed to be valid.
+   * @param tokenRenewalMarginInSeconds   [optional] The number of seconds before the end of the validity period during which the `SharedAccessKeyAuthenticationProvider` should renew the token.
+   */
+  static fromConnectionString(connectionString: string, tokenValidTimeInSeconds?: number, tokenRenewalMarginInSeconds?: number): SharedAccessKeyAuthenticationProvider {
+    if (!connectionString) {
+      /*Codes_SRS_NODE_SAK_AUTH_PROVIDER_16_006: [The `fromConnectionString` method shall throw a `ReferenceError` if the `connectionString` parameter is falsy.]*/
+      throw new ReferenceError('connectionString cannot be \'' + connectionString + '\'');
+    }
+
+    const cs = ConnectionString.parse(connectionString);
```
BertKleewein: >cs [](start = 10, length = 2)

nit: add type.

device/core/src/sak_authentication_provider.ts
```diff
@@ -0,0 +1,130 @@
...
+    if (!cs.SharedAccessKey) {
```
BertKleewein: >SharedAccessKey [](start = 12, length = 15)

Interesting...is this the only required field here?

device/core/src/sas_authentication_provider.ts
```diff
@@ -0,0 +1,55 @@
...
+    /*Codes_SRS_NODE_SAS_AUTHENTICATION_PROVIDER_16_004: [The `updateSharedAccessSignature` method shall emit a `newTokenAvailable` event with no arguments.]*/
+    this.emit('newTokenAvailable');
+  }
+
+  static fromSharedAccessSignature(sharedAccessSignature: string): SharedAccessSignatureAuthenticationProvider {
+    if (!sharedAccessSignature) {
+      /*Codes_SRS_NODE_SAS_AUTHENTICATION_PROVIDER_16_005: [The `fromSharedAccessSignature` method shall throw a `ReferenceError` if the `sharedAccessSignature` argument is falsy.]*/
+      throw new ReferenceError('sharedAccessSignature cannot be \'' + sharedAccessSignature + '\'');
+    }
+    const sas = SharedAccessSignature.parse(sharedAccessSignature);
```
BertKleewein: >sas [](start = 10, length = 3)

nit: add type

device/core/src/sas_authentication_provider.ts
```diff
@@ -0,0 +1,55 @@
...
+    const decodedUri = decodeURIComponent(sas.sr);
+    const uriSegments = decodedUri.split('/');
```
BertKleewein: >decodedUri [](start = 24, length = 10)

We're not doing any validation on the format of the URI, and we might not be able to.  This isn't the worst thing in the world, but we're relying on a failure somewhere else to catch this, probably with an unrelated error.  Would it be better to fail here with a meaningful error if we could?  At the very least, we could maybe fail based on the number of segments.

device/core/src/x509_authentication_provider.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  getDeviceCredentials(callback: (err: Error, credentials?: DeviceCredentials) => void): void {
+    /*Codes_SRS_NODE_X509_AUTHENTICATION_PROVIDER_16_002: [The `getDeviceCredentials` method shall call its callback with a `null` error object and the stored device credentials as a second argument.]*/
+    callback(null, this._credentials);
+  }
+
+  setX509Options(x509: X509): void {
+    /*Codes_SRS_NODE_X509_AUTHENTICATION_PROVIDER_16_003: [The `setX509Options` method shall store the `X509` object passed as an argument with the existing credentials.]*/
+    this._credentials.x509 = x509;
+  }
```
BertKleewein: Is there a use case for this function?  If not, suggest removing. The factory requires a cert, why would you ever change it?

device/core/test/_sak_authentication_provider_test.js
```diff
@@ -0,0 +1,136 @@
...
+    /*Tests_SRS_NODE_SAK_AUTH_PROVIDER_16_004: [The `updateSharedAccessSignature` method shall save the `sharedAccessSignature` passed as a parameter and return it when `getDeviceCredentials` is called, until it gets renewed.]*/
+    it('updates the shared access signature', function (testCallback) {
+      var newSas = 'newSas';
+      var fakeCredentials = {
+        deviceId: 'fakeDeviceId',
+        host: 'fake.host.name',
+        sharedAccessKey: 'fakeKey'
+      };
+      var sakAuthProvider = new SharedAccessKeyAuthenticationProvider(fakeCredentials);
+      sakAuthProvider.automaticRenewal = false;
```
BertKleewein: is automaticRenewal part of the public contract?  If so, suggest adding requirements.  If not, suggest renaming -- at least add an underscore.

common/core/common.d.ts
pierreca: technically these config objects contain only device credentials, which is why I thought it was a better name. but for backward compatibility purposes, we have to keep TransportConfig. I guess I could keep that name even though it sucks a little, and we can do the renaming/cleaning later?

---
In reply to: [157019626](https://github.com/Azure/azure-iot-sdk-node/pull/179#discussion_r157019626) [](ancestors = 157019626)

device/core/devdoc/sak_authentication_provider_requirements.md
pierreca: we could i've been going back and forth on that idea...

---
In reply to: [157020356](https://github.com/Azure/azure-iot-sdk-node/pull/179#discussion_r157020356) [](ancestors = 157020356)

device/core/src/sak_authentication_provider.ts
pierreca: maybe. technically it could work though, I'm wary of throwing in cases where we can just make it work

---
In reply to: [157021746](https://github.com/Azure/azure-iot-sdk-node/pull/179#discussion_r157021746) [](ancestors = 157021746)

device/core/src/sak_authentication_provider.ts
pierreca: huh. didn't think about that...

---
In reply to: [157021073](https://github.com/Azure/azure-iot-sdk-node/pull/179#discussion_r157021073) [](ancestors = 157021073)

device/core/src/sak_authentication_provider.ts
pierreca: good idea!

---
In reply to: [157022338](https://github.com/Azure/azure-iot-sdk-node/pull/179#discussion_r157022338) [](ancestors = 157022338)

device/core/src/sak_authentication_provider.ts
pierreca: no... i guess is should use the ConnectionString.parse() second parameter (required fields) too.

---
In reply to: [157023183](https://github.com/Azure/azure-iot-sdk-node/pull/179#discussion_r157023183) [](ancestors = 157023183)

device/core/src/x509_authentication_provider.ts
pierreca: the problem comes with the Client code, where it used to be that the user would create a client with an x509 connection string (not including the cert/key) and then call setOptions to add the certificate. since we can't wait until setOptions to create an AuthenticationProvider to give to the transport, we have to have a thing that calls the AuthenticationProvider from the body of setOptions to add the cert/key. it's mostly to be compatible with legacy code.

---
In reply to: [157025104](https://github.com/Azure/azure-iot-sdk-node/pull/179#discussion_r157025104) [](ancestors = 157025104)

common/core/common.d.ts
BertKleewein: I'm fine keep this as is for now.  I just want us to remember that transport configuration could contain more than just credentials.  I think this discussion gets mixed with our setTransportOptions silliness.  At one point, before code moved around, I wanted polling interval and timeout interval as part of the transport config.

---
In reply to: [157030220](https://github.com/Azure/azure-iot-sdk-node/pull/179#discussion_r157030220) [](ancestors = 157030220,157019626)

device/core/src/x509_authentication_provider.ts
BertKleewein: got it.  that make sense.  please mark as by design.

---
In reply to: [157032885](https://github.com/Azure/azure-iot-sdk-node/pull/179#discussion_r157032885) [](ancestors = 157032885,157025104)

device/core/src/sak_authentication_provider.ts
BertKleewein: this is cosmetic, but I don't see the value here.  If using a SAS provider will work better, why let the user do this with an SAK provider, especially if it's not going to work well?

---
In reply to: [157030877](https://github.com/Azure/azure-iot-sdk-node/pull/179#discussion_r157030877) [](ancestors = 157030877,157021746)

device/core/src/sas_authentication_provider.ts
pierreca: that is something we can revisit later.

---
In reply to: [157024476](https://github.com/Azure/azure-iot-sdk-node/pull/179#discussion_r157024476) [](ancestors = 157024476)

device/core/src/sak_authentication_provider.ts
pierreca: removed 

---
In reply to: [157041863](https://github.com/Azure/azure-iot-sdk-node/pull/179#discussion_r157041863) [](ancestors = 157041863,157030877,157021746)

device/core/test/_sak_authentication_provider_test.js
pierreca: removed

---
In reply to: [157025974](https://github.com/Azure/azure-iot-sdk-node/pull/179#discussion_r157025974) [](ancestors = 157025974)

device/core/src/sak_authentication_provider.ts
pierreca: so the problem is that token renewal can be asynchronous (with TPM for example) and so we have to check if the token needs to be renewed, just in case renewal is happening very slowly and at the time of the call, the token is expired. I guess we could circumvent that with a proper state machine and token request queue I guess.. but that's a major redesign so it'll have to wait

---
In reply to: [157031016](https://github.com/Azure/azure-iot-sdk-node/pull/179#discussion_r157031016) [](ancestors = 157031016,157021073)

PR #180, opened 2017-12-14
Device client modifications to support the new AuthenticationProvider object

pierreca:
The PR follows #179 and implements the required changes in the device client package to support the new AuthenticationProvider classes.

While the unit tests work, changes to the transport packages are required to fully enable the feature so it will be merged with transports PR before it's merged into master.

### DIFF
--- a/device/core/devdoc/device_client_requirements.md
+++ b/device/core/devdoc/device_client_requirements.md
@@ -31,9 +31,11 @@ client.sendEvent(new Message('hello world'), print);
 #### fromConnectionString
 **SRS_NODE_DEVICE_CLIENT_05_003: [** The `fromConnectionString` method shall throw ReferenceError if the connStr argument is falsy. **]**
 
-**SRS_NODE_DEVICE_CLIENT_05_005: [** `fromConnectionString` shall derive and transform the needed parts from the connection string in order to create a new instance of Transport. **]**
-
 **SRS_NODE_DEVICE_CLIENT_05_006: [** The `fromConnectionString` method shall return a new instance of the `Client` object, as by a call to `new Client(new Transport(...))`. **]**
+
+**SRS_NODE_DEVICE_CLIENT_16_087: [** The `fromConnectionString` method shall create a new `SharedAccessKeyAuthorizationProvider` object with the connection string passed as argument if it contains a SharedAccessKey parameter and pass this object to the transport constructor. **]**
+
+**SRS_NODE_DEVICE_CLIENT_16_093: [** The `fromConnectionString` method shall create a new `X509AuthorizationProvider` object with the connection string passed as argument if it contains an X509 parameter and pass this object to the transport constructor. **]**
 
 #### fromSharedAccessSignature
 
@@ -41,14 +43,25 @@ client.sendEvent(new Message('hello world'), print);
 
 **SRS_NODE_DEVICE_CLIENT_16_030: [** The `fromSharedAccessSignature` method shall return a new instance of the `Client` object **]**
 
+**SRS_NODE_DEVICE_CLIENT_16_088: [** The `fromSharedAccessSignature` method shall create a new `SharedAccessSignatureAuthorizationProvider` object with the shared access signature passed as argument, and pass this object to the transport constructor. **]**
+
+#### fromAuthenticationProvider
+
+**SRS_NODE_DEVICE_CLIENT_16_089: [** The `fromAuthenticationProvider` method shall throw a `ReferenceError` if the `authenticationProvider` argument is falsy. **]**
+
+**SRS_NODE_DEVICE_CLIENT_16_092: [** The `fromAuthenticationProvider` method shall throw a `ReferenceError` if the `transportCtor` argument is falsy. **]**
+
+**SRS_NODE_DEVICE_CLIENT_16_090: [** The `fromAuthenticationProvider` method shall pass the `authenticationProvider` object passed as argument to the transport constructor. **]**
+
+**SRS_NODE_DEVICE_CLIENT_16_091: [** The `fromAuthenticationProvider` method shall return a `Client` object configured with a new instance of a transport created using the `transportCtor` argument. **]**
+
+
 ### Constructors
 #### Client(transport, connectionString) constructor
 
 **SRS_NODE_DEVICE_CLIENT_05_001: [** The `Client` constructor shall accept a transport object **]**  , e.g. `azure-iot-device-http.Http`.
 
 **SRS_NODE_DEVICE_CLIENT_16_026: [** The `Client` constructor shall accept a connection string as an optional second argument **]**
-
-**SRS_NODE_DEVICE_CLIENT_16_027: [** If a connection string argument is provided and is using SharedAccessKey authentication, the `Client` shall automatically generate and renew SAS tokens. **]**
 
 ### Public Methods
 
@@ -158,8 +171,6 @@ The `sendEventBatch` method sends a list of event messages to the IoT Hub as the
 #### updateSharedAccessSignature(sharedAccessSignature, done)
 
 **SRS_NODE_DEVICE_CLIENT_16_031: [** The `updateSharedAccessSignature` method shall throw a `ReferenceError` if the sharedAccessSignature parameter is falsy. **]**
-
-**SRS_NODE_DEVICE_CLIENT_06_002: [** The `updateSharedAccessSignature` method shall throw a `ReferenceError` if the client was created using x509. **]**
 
 **SRS_NODE_DEVICE_CLIENT_16_032: [** The `updateSharedAccessSignature` method shall call the `updateSharedAccessSignature` method of the transport currently in use with the sharedAccessSignature parameter. **]**
 
--- a/device/core/src/blob_upload/blob_upload_client.ts
+++ b/device/core/src/blob_upload/blob_upload_client.ts
@@ -4,7 +4,7 @@
 'use strict';
 
 import { Stream } from 'stream';
-import { X509, SharedAccessSignature } from 'azure-iot-common';
+import { AuthenticationProvider } from 'azure-iot-common';
 
 import { BlobUploadResult } from './blob_upload_result';
 import { BlobUploader as DefaultBlobUploader } from './blob_uploader';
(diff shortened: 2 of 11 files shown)


device/core/devdoc/device_client_requirements.md
```diff
@@ -159,8 +172,6 @@ The `sendEventBatch` method sends a list of event messages to the IoT Hub as the
 
 **SRS_NODE_DEVICE_CLIENT_16_031: [** The `updateSharedAccessSignature` method shall throw a `ReferenceError` if the sharedAccessSignature parameter is falsy. **]**
 
-**SRS_NODE_DEVICE_CLIENT_06_002: [** The `updateSharedAccessSignature` method shall throw a `ReferenceError` if the client was created using x509. **]**
```
BertKleewein: >SRS_NODE_DEVICE_CLIENT_06_002 [](start = 2, length = 29)

not sure why you removed this.

device/core/src/client.ts
```diff
@@ -78,19 +76,13 @@ export class Client extends EventEmitter {
    * @param {string}  connStr           A connection string (optional: when not provided, updateSharedAccessSignature must be called to set the SharedAccessSignature token directly).
    * @param {Object}  blobUploadClient  An object that is capable of uploading a stream to a blob.
    */
-  constructor(transport: Client.Transport, connStr?: string, blobUploadClient?: BlobUploadClient) {
+  constructor(transport: Client.Transport, authentication?: string | AuthenticationProvider, blobUploadClient?: BlobUploadClient) {
```
BertKleewein: >authentication [](start = 43, length = 14)

this code doesn't do anything with the authentication object.  I assume that's coming with the transport changes, right?

BertKleewein: ah, this is so we don't have a breaking change in the constructor signature.. This is a breaking change in the constructor behavior, however.  Do we want to change this so the connection string can be used to create an authentication object and then pass it into the transport?  That would retain the previous behavior.

---
In reply to: [157028308](https://github.com/Azure/azure-iot-sdk-node/pull/180#discussion_r157028308) [](ancestors = 157028308)

device/core/src/blob_upload/file_upload_api.ts
```diff
@@ -3,7 +3,7 @@
 
 'use strict';
 
-import { endpoint, X509, SharedAccessSignature } from 'azure-iot-common';
+import { endpoint, AuthenticationProvider } from 'azure-iot-common';
 import { Http as DefaultHttpTransport } from 'azure-iot-http-base';
 import { UploadParams, FileUpload as FileUploadInterface } from './blob_upload_client';
```
BertKleewein: >FileUpload as FileUploadInterface  [](start = 23, length = 34)

nit: aargh.  aliases.

device/core/src/blob_upload/blob_upload_client.ts
```diff
@@ -27,15 +27,15 @@ export interface UploadParams {
...
-    getBlobSharedAccessSignature(blobName: string, auth: X509 | SharedAccessSignature, done: (err: Error, uploadParams?: UploadParams) => void): void;
-    notifyUploadComplete(correlationId: string, auth: X509 | SharedAccessSignature, uploadResult: BlobUploadResult, done: (err?: Error) => void): void;
+  getBlobSharedAccessSignature(blobName: string, done: (err: Error, uploadParams?: UploadParams) => void): void;
+  notifyUploadComplete(correlationId: string, uploadResult: BlobUploadResult, done: (err?: Error) => void): void;
 }
 
 /**
  * @private
  */
 export interface BlobUploader {
```
BertKleewein: >BlobUploader [](start = 17, length = 12)

nit/gripe: these interfaces are hard to understand.  I think some of the interfaces get exposed to the device client, i think some of them get exposed to the end user (maybe), and I think some of them are used internally to the file/blob upload code, but I'm not completely sure.  Having 2 interfaces, one called BlobUpload and one called  BlobUploader that look almost the same also confuses me.  I know one is internal and one is external, but i'll be damned if I can figure out which.  I think it might benefit readability to add prefixes like "ProviderInterface" for internal interfaces that provide functionality for other code,  "ClientInterface" for things that outside users would see.  Maytbe.  I just think the interface name needs to describe what the interface does and not just the feature that it's involved with.  </soapbox>

BertKleewein: There's also some kind of relationship between FileUpload and BlobUpload, and i think maybe one uses the other, but this is also fuzzy

---
In reply to: [157072624](https://github.com/Azure/azure-iot-sdk-node/pull/180#discussion_r157072624) [](ancestors = 157072624)

BertKleewein: The point is, I don't think there's any breaking changes in this folder, but I honestly don't know because I can't figure out how the pieces fit together. (I mean, I could, but like all developers, I'm lazy and it looks like spaghetti to me.)

---
In reply to: [157072902](https://github.com/Azure/azure-iot-sdk-node/pull/180#discussion_r157072902) [](ancestors = 157072902,157072624)

device/core/test/http_simulated.js
```diff
@@ -20,31 +20,33 @@ function makeError(statusCode) {
...
+          if (config.sharedAccessSignature === cmpSig) {  // bad key
+            done(makeError(401));
+          }
+          else {
+            done(null, new results.MessageEnqueued(new Response(204)));
+          }
         }
       }
-    }
+    }.bind(this));
```
BertKleewein: >}.bind(this)); [](start = 4, length = 14)

nit: I hate bind.  I think I know why you did this, but I don't like it (seems like typescript is breaking a promise here).  Why not use self instead?

device/core/test/_client_test.js
```diff
@@ -96,9 +67,27 @@ describe('Client', function () {
       clock.restore();
       testCallback();
     });
+
+    /*Tests_SRS_NODE_DEVICE_CLIENT_16_087: [The `fromConnectionString` method shall create a new `SharedAccessKeyAuthorizationProvider` object with the connection string passed as argument if it contains a SharedAccessKey parameter and pass this object to the transport constructor. ]*/
+    it('creates a SharedAccessKeyAuthorizationProvider and passes it to the transport', function (testCallback) {
+      var client = Client.fromConnectionString(sharedKeyConnectionString, function (authProvider) {
```
BertKleewein: >function (authProvider) [](start = 74, length = 23)

nit: I'm not sure what I think about this.  It feels dirty.  I know it's perfectly legal to pass a callback as a constructor function, but there's a conceptual difference here that makes me uneasy.  Or maybe not.  I'm not sure.

device/core/test/_client_test.js
```diff
@@ -110,32 +99,50 @@ describe('Client', function () {
...
-      var sharedAccessSignature = '"SharedAccessSignature sr=hubName.azure-devices.net%2Fdevices%2FdeviceId&sig=s1gn4tur3&se=1454204843"';
-      var DummyTransport = function (config) {
-        this.on = function() {};
-        assert.strictEqual(config.host, 'hubName.azure-devices.net');
-        assert.strictEqual(config.deviceId, 'deviceId');
-        assert.strictEqual(config.sharedAccessSignature, sharedAccessSignature);
-      };
-      Client.fromSharedAccessSignature(sharedAccessSignature, DummyTransport);
+  describe('#fromAuthenticationProvider', function () {
+    /*Tests_SRS_NODE_DEVICE_CLIENT_16_089: [The `fromAuthenticationProvider` method shall throw a `ReferenceError` if the `authenticationProvider` argument is falsy.]*/
```
BertKleewein: >falsy [](start = 159, length = 5)

nit: should the requirement be !falsy, or should the requirement be that it must be a function?

device/core/devdoc/device_client_requirements.md
BertKleewein: but I'm sure you had a good reason.  I'm just thinking out loud.

---
In reply to: [157027425](https://github.com/Azure/azure-iot-sdk-node/pull/180#discussion_r157027425) [](ancestors = 157027425)

pierreca: it's another's object responsibility now :)

---
In reply to: [157076486](https://github.com/Azure/azure-iot-sdk-node/pull/180#discussion_r157076486) [](ancestors = 157076486,157027425)

device/core/src/blob_upload/file_upload_api.ts
pierreca: unrelated to the change :)

---
In reply to: [157069967](https://github.com/Azure/azure-iot-sdk-node/pull/180#discussion_r157069967) [](ancestors = 157069967)

device/core/test/http_simulated.js
pierreca: this is javascript code so we don't get the automatic _this thing. we could use self instead of bind for sure.

---
In reply to: [157074839](https://github.com/Azure/azure-iot-sdk-node/pull/180#discussion_r157074839) [](ancestors = 157074839)

device/core/test/_client_test.js
pierreca: not sure what you mean. we usually enforce arguments like that but not type of arguments.

---
In reply to: [157076086](https://github.com/Azure/azure-iot-sdk-node/pull/180#discussion_r157076086) [](ancestors = 157076086)

device/core/test/_client_test.js
pierreca: it's not a callback, it's just a function. just like a constructor. i could define a constructor outside, call it constructor, and then pass it here, but that'd be exactly the same thing

---
In reply to: [157075622](https://github.com/Azure/azure-iot-sdk-node/pull/180#discussion_r157075622) [](ancestors = 157075622)

device/core/src/blob_upload/blob_upload_client.ts
pierreca: point taken. it's all old design though, not new stuff, so let's file this as part of future refactoring

---
In reply to: [157073502](https://github.com/Azure/azure-iot-sdk-node/pull/180#discussion_r157073502) [](ancestors = 157073502,157072902,157072624)

device/core/src/client.ts
pierreca: adding an exception to signal this thing isn't meant to be used by users.

---
In reply to: [157068579](https://github.com/Azure/azure-iot-sdk-node/pull/180#discussion_r157068579) [](ancestors = 157068579,157028308)

PR #182, opened 2017-12-15
Mqtt transport modifications for the new AuthenticationProviders

pierreca:
Add support for AuthenticationProviders in the MQTT transport.

### DIFF
--- a/device/transport/mqtt/src/mqtt.ts
+++ b/device/transport/mqtt/src/mqtt.ts
@@ -6,14 +6,15 @@ import * as querystring from 'querystring';
 import * as URL from 'url';
 import * as machina from 'machina';
 
-import { results, errors, Message, X509 } from 'azure-iot-common';
-import { DeviceMethodResponse, Client } from 'azure-iot-device';
+import { results, errors, Message, X509, AuthenticationProvider, TransportConfig } from 'azure-iot-common';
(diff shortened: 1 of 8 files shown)


device/transport/mqtt/src/mqtt.ts
```diff
@@ -6,14 +6,15 @@ import * as querystring from 'querystring';
...
-import { DeviceMethodResponse, Client } from 'azure-iot-device';
+import { results, errors, Message, X509, AuthenticationProvider, TransportConfig } from 'azure-iot-common';
+import { DeviceMethodResponse, Client, X509AuthenticationProvider, SharedAccessSignatureAuthenticationProvider } from 'azure-iot-device';
 import { EventEmitter } from 'events';
 import * as util from 'util';
 import * as dbg from 'debug';
 const debug = dbg('azure-iot-device-mqtt:Mqtt');
 import { MqttBase, translateError } from 'azure-iot-mqtt-base';
 import { MqttTwinReceiver } from './mqtt_twin_receiver';
+import { AuthenticationType } from '../../../../common/core/lib/authentication_provider';
```
BertKleewein: >AuthenticationType [](start = 9, length = 18)

ouch!  please tell me that this doesn't generate a js require() call with this path.  we should fix this later.  (or now if it suits you).

pierreca: good catch. fixed.

PR #192, opened 2018-01-05
separate disconnect and cancel at the transport level

BertKleewein:
This is preparing to implement a proper timeout mechanism that behaves the same way regardless of transport.  This will also lay some groundwork for retry.

This PR has the following:
1. Timeout handling was removed from HTTP and MQTT transports.  HTTP was relying on lower-layer objects to do timeout, so I removed this.  MQTT had specific timeout code that needed to be removed.
1. I added an abort function to azure-iot-http-base.  This may not be strictly necessary, and I am re-thinking about this and may remove it.  This was necessary to support the cancel operation.  (other protocols just stop listening for responses on cancel, so HTTP could maybe do the same)
1. In the transports, I made cancel and disconnect into 2 different operations.  Cancel cancels the current operation but leaves the transport connected.  Disconnect disconnects the transport.
1. I updated transports so failures don't automatically disconnect the transport.  Deciding when to disconnect should be done at higher layers.
1. I updated the polling state machine to handle distinct cancel and disconnect operations.
1. In order to handle the cancel functions in the transports, I leverage the this._operations lookup.  In order for this to work, I needed to allocate the correlationId or rid at the very beginning of the operation.  This is necessary so we can properly cancel an operation while the transport is still connecting.  (This also allowed me to move the _deferredCancelInQueue lookahead in mqtt.ts, which makes me happy.)
1. I changed MQTT disconnect-while-connecting behavior to match AMQP.  If you try to disconnect or cancel while connecting, MQTT will now immediately cancel the connection.  (it used to wait for the connection to complete before disconnecting, which doesn't help us if the connection is _never_ going to complete.)
1. I updated the HTTP requirements to more closely match what the transport does.  They were previously written from a POV above the polling state machine.  Now they're written from the POV above the transport.  Tests still need to be updated to match these new requirements.
1. I added requirements and tests for all of the above.

### DIFF
--- a/provisioning/device/package.json
+++ b/provisioning/device/package.json
@@ -24,13 +24,13 @@
   "scripts": {
     "lint": "tslint --project . -c ../../tslint.json",
     "build": "tsc",
-    "unittest-min": "istanbul cover --report none node_modules/mocha/bin/_mocha -- --reporter dot test/_*_test.js",
-    "alltest-min": "istanbul cover --report none node_modules/mocha/bin/_mocha -- --reporter dot test/_*_test*.js",
+    "unittest-min": "istanbul cover node_modules/mocha/bin/_mocha -- --reporter dot test/_*_test.js",
+    "alltest-min": "istanbul cover node_modules/mocha/bin/_mocha -- --reporter dot test/_*_test*.js",
     "unittest": "istanbul cover node_modules/mocha/bin/_mocha -- --reporter spec test/_*_test.js",
     "alltest": "istanbul cover node_modules/mocha/bin/_mocha -- --reporter spec test/_*_test*.js",
     "ci": "npm -s run lint && npm -s run build && npm -s run alltest-min && npm -s run check-cover",
     "test": "npm -s run lint && npm -s run build && npm -s run unittest",
-    "check-cover": "istanbul check-coverage --statements 84 --branches 68 --functions 70 --lines 86"
+    "check-cover": "istanbul check-coverage --statements 85 --branches 64 --functions 72 --lines 86"
   },
   "engines": {
     "node": ">= 0.10"
--- a/provisioning/device/src/constants.ts
+++ b/provisioning/device/src/constants.ts
@@ -28,5 +28,5 @@ export class ProvisioningDeviceConstants {
   /**
    * default timeout to use when communicating with the service
    */
-  static defaultTimeoutInterval: number = 4000;
+  static defaultTimeoutInterval: number = 30000;
 }
--- a/provisioning/device/src/polling_state_machine.ts
+++ b/provisioning/device/src/polling_state_machine.ts
@@ -6,7 +6,8 @@
 import { EventEmitter } from 'events';
 import { errors } from 'azure-iot-common';
 import * as machina from 'machina';
-import { PollingTransport, RegistrationRequest } from './interfaces';
+import { ProvisioningDeviceConstants } from './constants';
+import { PollingTransport, RegistrationRequest, DeviceRegistrationResult } from './interfaces';
 import * as dbg from 'debug';
 const debug = dbg('azure-iot-provisioning-device:PollingStateMachine');
 
@@ -25,41 +26,43 @@ export class  PollingStateMachine extends EventEmitter {
     this._transport = transport;
 
     this._fsm = new machina.Fsm({
-      namespace: 'provisioning-transport',
+      namespace: 'provisioning-client-polling',
       initialState: 'disconnected',
       states: {
         disconnected: {
           _onEnter: (err, result, response, callback) => {
-            this._currentOperationCallback = null;
             if (callback) {
               callback(err, result, response);
             }
           },
-          register: (request, callback) => {
-            this._fsm.transition('sendingRegistrationRequest', request, callback);
-          },
-          cancel: (err, result, response, callback) => {
-            /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_025: [ If `cancel` is called while disconnected, it shall immediately call its `callback`. ] */
-            // nothing to do.
-            callback();
-          }
+          register: (request, callback) => this._fsm.transition('sendingRegistrationRequest', request, callback),
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_025: [ If `cancel` is called while disconnected, it shall immediately call its `callback`. ] */
+          cancel: (cancelledOpErr, callback) => callback(),
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_031: [ If `disconnect` is called while disconnected, it shall immediately call its `callback`. ] */
+          disconnect: (callback) => callback()
         },
         idle: {
-          _onEnter: (err, result, response, callback) => {
-            callback(err, result, response);
-          },
-          cancel: (err, result, response, callback) => {
-            this._fsm.transition('cancelling', err, result, response, callback);
-          },
-          register: (request, callback) => {
-            this._fsm.transition('sendingRegistrationRequest', request, callback);
-          },
+          _onEnter: (err, result, response, callback) => callback(err, result, response),
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_030: [ If `cancel` is called while the transport is connected but idle, it shall immediately call its `callback`. ] */
+          cancel: (cancelledOpErr, callback) => callback(),
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_032: [ If `disconnect` is called while while the transport is connected but idle, it shall call `PollingTransport.disconnect` and call it's `callback` passing the results of the transport operation. ] */
+          disconnect: (callback) => this._fsm.transition('disconnecting', callback),
+          register: (request, callback) => this._fsm.transition('sendingRegistrationRequest', request, callback)
         },
         sendingRegistrationRequest: {
           _onEnter: (request, callback) => {
+            let timeoutTimer = setTimeout(() => {
+              /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_036: [ If `PollingTransport.registrationRequest` does not call its callback within `ProvisioningDeviceConstants.defaultTimeoutInterval` ms, register shall with with a `TimeoutError` error. ] */
+              if (this._currentOperationCallback === callback) {
+                  debug('timeout while sending request');
+                  /* tslint:disable:no-empty */
+                  this._fsm.handle('cancel', new errors.TimeoutError(), () => {} );
+                }
+              }, ProvisioningDeviceConstants.defaultTimeoutInterval);
             /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_012: [ `register` shall call `PollingTransport.registrationRequest`. ] */
             this._currentOperationCallback = callback;
             this._transport.registrationRequest(request, (err, result, response, pollingInterval) => {
+              clearTimeout(timeoutTimer);
               // Check if the operation is still pending before transitioning.  We might be in a different state now and we don't want to mess that up.
               if (this._currentOperationCallback === callback) {
                 this._fsm.transition('responseReceived', err, request, result, response, pollingInterval, callback);
@@ -68,7 +71,13 @@ export class  PollingStateMachine extends EventEmitter {
               }
             });
           },
-          cancel: (err, result, response, callback) => this._fsm.transition('cancelling', err, result, response, callback),
+          cancel: (cancelledOpErr, callback) => this._fsm.transition('cancelling', cancelledOpErr, callback),
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_033: [ If `disconnect` is called while in the middle of a `registrationRequest` operation, the operation shall be cancelled and the transport shall be disconnected. ] */
+          disconnect: (callback) => {
+            this._fsm.handle('cancel', new errors.OperationCancelledError(''), (err) => {
+              this._fsm.handle('disconnect', callback);
+            });
+          },
           /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_024: [ If `register` is called while a different request is in progress, it shall fail with an `InvalidOperationError`. ] */
           register: (request, callback) => callback(new errors.InvalidOperationError('another operation is in progress'))
         },
@@ -128,7 +137,7 @@ export class  PollingStateMachine extends EventEmitter {
         responseError: {
           _onEnter: (err, result, response, callback) => {
             this._currentOperationCallback = null;
-            this._fsm.transition('cancelling', err, result, response, callback);
+            this._fsm.transition('idle', err, result, response, callback);
           },
           '*': () => this._fsm.deferUntilTransition()
         },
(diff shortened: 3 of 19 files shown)


anthonyvercolano:
>   getEndorsementKey(callback: (err: Error, endorsementKey?: string) => void): void;

Should be buffer.

---
Refers to: provisioning/device/src/interfaces.ts:190 in b941c8a. [](commit_id = b941c8aad2191bbfa44b48ec000d886d7767b694, deletion_comment = False)

anthonyvercolano:
>   getStorageRootKey(callback: (err: Error, storageRootKey?: string) => void): void;

Buffer here and for every string parameter for this API set.

---
Refers to: provisioning/device/src/interfaces.ts:191 in b941c8a. [](commit_id = b941c8aad2191bbfa44b48ec000d886d7767b694, deletion_comment = False)

common/transport/http/src/http.ts
```diff
@@ -26,6 +26,9 @@ export type HttpCallback = (err: Error, body?: string, response?: IncomingMessag
  *                  You generally want to use these higher-level objects (such as [azure-iot-device-http.Http]{@link module:azure-iot-device-http.Http}) rather than this one.
  */
 export class Http {
+
+  private _request: (options: RequestOptions | string, callback?: (res: IncomingMessage) => void) => ClientRequest = request;
```
pierreca: >_request [](start = 10, length = 8)

if this is used only for mocking in tests, why not use sinon stubs? plus the type/default value here is really hard to read and understand

common/transport/http/src/rest_api_client.ts
```diff
@@ -37,6 +37,7 @@ export class RestApiClient {
   private _config: RestApiClient.TransportConfig;
   private _http: HttpBase;
   private _userAgent: string;
+  private _currentRequest: ClientRequest;
```
pierreca: >_currentRequest [](start = 10, length = 15)

this assumes there's only one request at a time - this assumption is not clear at all from the outside perspective (looking at the RestApiClient API).

provisioning/device/src/polling_state_machine.ts
```diff
@@ -69,6 +62,12 @@ export class  PollingStateMachine extends EventEmitter {
             });
           },
           cancel: (err, result, response, callback) => this._fsm.transition('cancelling', err, result, response, callback),
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_033: [ If `disconnect` is called while in the middle of a `registrationRequest` operation, the operation shall be cancelled and the transport shall be disconnected. ] */
+          disconnect: (callback) => {
+            this._fsm.handle('cancel', null, null, null, (err) => {
+              this._fsm.transition('disconnecting', callback);
```
pierreca: >this._fsm.transition('disconnecting', callback); [](start = 14, length = 48)

since cancel induces a transition, i'd rather you call this._fsm.handle(disconnect, callback) here rather than directly a transition.

provisioning/device/src/polling_state_machine.ts
```diff
@@ -145,6 +144,12 @@ export class  PollingStateMachine extends EventEmitter {
             this._pollingTimer = null;
             this._fsm.transition('cancelling', err, result, response, callback);
           },
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_034: [ If `disconnect` is called while the state machine is waiting to poll, the current operation shall be cancelled and the transport shall be disconnected. ] */
+          disconnect: (callback) => {
+            this._fsm.handle('cancel', null, null, null, (err) => {
+              this._fsm.transition('disconnecting', callback);
```
pierreca: >transition [](start = 24, length = 10)

same here - handle('disconnect', callback) instead of transition (and in other places too)

provisioning/device/test/_polling_state_machine_test.js
```diff
@@ -14,6 +15,10 @@ var fakeRequest = {
   provisioningHost: '__FAKE_HOST__',
   idScope: '__FAKE_SCOPE__'
 }
+var fakeError = new Error(fakeErrorText);
+
+assert.isNotAnError = function(err) { assert.isNotOk(err, 'expected success, got an error.'); };
+assert.isAnError = function(err) { assert.isOk(err, 'expected an error.  got success'); };
```
pierreca: uhhh modifying a variable in the global scope... not a generally good practice

how about using assert.instanceOf(err, Error)?

provisioning/device/test/_polling_state_machine_test.js
```diff
@@ -263,27 +269,22 @@ describe('state machine', function () {
...
-            assert(!err);
-            assert(machine._transport.cancel.calledOnce);
-            testCallback();
-          });
-        });
+    /* Tests_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_030: [ If `cancel` is called while the transport is connected but idle, it shall immediately call its `callback`. ] */
+    it ('does nothing if called while idle', function (testCallback) {
+      callRegisterWithDefaultArgs(function(err) {
+        assert.isNotAnError(err);
+        testCallback();
```
pierreca: >testCallback [](start = 8, length = 12)

you're not calling cancel anymore?

provisioning/device/test/_polling_state_machine_test.js
```diff
@@ -356,5 +357,91 @@ describe('state machine', function () {
...
+        op.preconfigure();
+
+        callRegisterWithDefaultArgs(function(err) {
+          assert.instanceOf(err, errors.OperationCancelledError);
+          registerCallbackCalled = true;
+        });
+
+        setTimeout(function () {
+          machine.disconnect(function() {
+            assert(registerCallbackCalled);
```
pierreca: >assert [](start = 12, length = 6)

assert.isTrue?

provisioning/device/test/_polling_state_machine_test.js
```diff
@@ -356,5 +357,91 @@ describe('state machine', function () {
...
+        },
+        assert: () => {
+          assert.isTrue(machine._transport.registrationRequest.calledOnce);
+          assert.isTrue(machine._transport.queryOperationStatus.calledOnce);
+        }
+
+      }
+    ].forEach(function(op) {
+      it ('cancels ' + op.name + ' and disconnects', function (testCallback) {
+        var registerCallbackCalled;
```
pierreca: >registerCallbackCalled [](start = 12, length = 22)

initialize to false?

provisioning/transport/amqp/devdoc/amqp_requirements.md
```diff
@@ -84,12 +84,34 @@ iotdps-operation-id: <operationId>;
...
+
+**SRS_NODE_PROVISIONING_AMQP_16_022: [** `disconnect` shall call its callback immediately if the AMQP connection is disconnected. **]**
+
+**SRS_NODE_PROVISIONING_AMQP_16_023: [** `disconnect` shall detach the sender and receiver links and disconnect the AMQP connection. **]**
+
+**SRS_NODE_PROVISIONING_AMQP_16_024: [** `disconnect` shall call its callback with no arguments if all detach/disconnect operations were successful. **]**
+
+**SRS_NODE_PROVISIONING_AMQP_16_025: [** `disconnect` shall call its callback with the error passed from the first unsuccessful detach/disconnect operation if one of those fail. **]**
+
+**SRS_NODE_PROVISIONING_AMQP_18_009: [** `disconnect` shall disonnect the AMQP connection and cancel the operation that intiatiated a connection if called while the connection is in process. **]**
```
pierreca: >intiatiated [](start = 120, length = 11)

typo

provisioning/transport/amqp/devdoc/amqp_requirements.md
```diff
@@ -84,12 +84,34 @@ iotdps-operation-id: <operationId>;
...
+**SRS_NODE_PROVISIONING_AMQP_18_002: [** `disconnect` shall cause a `queryOperationStatus` operation that is in progress to call its callback passing an `OperationCancelledError` object. **]**
+
 ## cancel(callback: (err?: Error) => void): void
 
-**SRS_NODE_PROVISIONING_AMQP_16_022: [** `cancel` shall call its callback immediately if the AMQP connection is disconnected. **]**
+**SRS_NODE_PROVISIONING_AMQP_18_003: [** `cancel` shall call its callback immediately if the AMQP connection is disconnected. **]**
+
+**SRS_NODE_PROVISIONING_AMQP_18_004: [** `cancel` shall call its callback immediately if the AMQP connection is connected but idle. **]**
+
+**SRS_NODE_PROVISIONING_AMQP_18_005: [** `cancel` shall disconnect the AMQP connection and cancel the operation that intiatiated a connection if called while the connection is in process. **]**
```
pierreca: >intiatiated [](start = 117, length = 11)

typo here too

provisioning/device/test/_polling_state_machine_test.js
BertKleewein: Not if the transport is connected but idle.  It used to be that 'cancel' meant disconnect in which case it made sense.  now that 'cancel' means 'cancel the current operation', it doesn't make sense to call it anymore.

---
In reply to: [160014792](https://github.com/Azure/azure-iot-sdk-node/pull/192#discussion_r160014792) [](ancestors = 160014792)

BertKleewein:
>   getEndorsementKey(callback: (err: Error, endorsementKey?: string) => void): void;

I can accept that, but that's out of scope for this review.

---
In reply to: [355677028](https://github.com/Azure/azure-iot-sdk-node/pull/192#issuecomment-355677028) [](ancestors = 355677028)

---
Refers to: provisioning/device/src/interfaces.ts:190 in b941c8a. [](commit_id = b941c8aad2191bbfa44b48ec000d886d7767b694, deletion_comment = False)

BertKleewein:
>   getStorageRootKey(callback: (err: Error, storageRootKey?: string) => void): void;

out of scope.

---
In reply to: [355677349](https://github.com/Azure/azure-iot-sdk-node/pull/192#issuecomment-355677349) [](ancestors = 355677349)

---
Refers to: provisioning/device/src/interfaces.ts:191 in b941c8a. [](commit_id = b941c8aad2191bbfa44b48ec000d886d7767b694, deletion_comment = False)

common/transport/http/src/rest_api_client.ts
```diff
@@ -22,6 +22,10 @@ export interface HttpTransportError extends Error {
   responseBody?: any;
 }
 
+export interface HttpOperation {
```
pierreca: that feels like a really incomplete interface for an HttpOperation or a really generic name for something that is just aborting a request.

common/transport/http/src/rest_api_client.ts
```diff
@@ -152,6 +156,12 @@ export class RestApiClient {
     }
 
     request.end();
+
+    /*Codes_SRS_NODE_IOTHUB_REST_API_CLIENT_18_004: [ `executeApiCall` shall return an `HttpOperation` object. ] */
+    return {
+      /*Codes_SRS_NODE_IOTHUB_REST_API_CLIENT_18_003: [ `HttpOperation.abort` shall call `abort` on the `ClientRequest` object. ] */
+      abort: () => request.abort()
+    };
```
pierreca: I'm really not sold on returning an object here - I do understand why you're doing it: to provide the caller with a way to cancel the request. I'm not sure why we care about cancelling the request instead of letting it die peacefully (making sure we don't get a random error later? or something like that?). 
I wonder what other options we have. return a request id and maintain a list of requests? return the lower-layer http request object directly? 
I'm also concerned by how much context we'll be keeping in memory (because of the closure) if a user decides to hold on to the reference to the returned object.

provisioning/device/package.json
```diff
@@ -24,13 +24,13 @@
...
-    "unittest-min": "istanbul cover --report none node_modules/mocha/bin/_mocha -- --reporter dot test/_*_test.js",
-    "alltest-min": "istanbul cover --report none node_modules/mocha/bin/_mocha -- --reporter dot test/_*_test*.js",
+    "unittest-min": "istanbul cover node_modules/mocha/bin/_mocha -- --reporter dot test/_*_test.js",
+    "alltest-min": "istanbul cover node_modules/mocha/bin/_mocha -- --reporter dot test/_*_test*.js",
     "unittest": "istanbul cover node_modules/mocha/bin/_mocha -- --reporter spec test/_*_test.js",
     "alltest": "istanbul cover node_modules/mocha/bin/_mocha -- --reporter spec test/_*_test*.js",
     "ci": "npm -s run lint && npm -s run build && npm -s run alltest-min && npm -s run check-cover",
     "test": "npm -s run lint && npm -s run build && npm -s run unittest",
-    "check-cover": "istanbul check-coverage --statements 84 --branches 68 --functions 70 --lines 86"
+    "check-cover": "istanbul check-coverage --statements 85 --branches 64 --functions 72 --lines 86"
```
pierreca: >64 [](start = 71, length = 2)

that's a 4 point drop, i'm starting to feel it :P

provisioning/device/src/constants.ts
```diff
@@ -28,5 +28,5 @@ export class ProvisioningDeviceConstants {
   /**
    * default timeout to use when communicating with the service
    */
-  static defaultTimeoutInterval: number = 4000;
+  static defaultTimeoutInterval: number = 30000;
```
pierreca: >30000 [](start = 42, length = 5)

just double checking if this is vetted with other SDKs and the service folks - this is an important value.

provisioning/device/src/polling_state_machine.ts
```diff
@@ -25,41 +26,43 @@ export class  PollingStateMachine extends EventEmitter {
...
-            this._fsm.transition('sendingRegistrationRequest', request, callback);
-          },
-          cancel: (err, result, response, callback) => {
-            /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_025: [ If `cancel` is called while disconnected, it shall immediately call its `callback`. ] */
-            // nothing to do.
-            callback();
-          }
+          register: (request, callback) => this._fsm.transition('sendingRegistrationRequest', request, callback),
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_025: [ If `cancel` is called while disconnected, it shall immediately call its `callback`. ] */
+          cancel: (cancelledOpErr, callback) => callback(),
```
pierreca: >callback() [](start = 48, length = 10)

feels like we'd be swallowing the cancelledOpErr argument here - are we guaranteed that it's null here? should we throw if it isn't?

provisioning/device/src/polling_state_machine.ts
```diff
@@ -25,41 +26,43 @@ export class  PollingStateMachine extends EventEmitter {
...
+          _onEnter: (err, result, response, callback) => callback(err, result, response),
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_030: [ If `cancel` is called while the transport is connected but idle, it shall immediately call its `callback`. ] */
+          cancel: (cancelledOpErr, callback) => callback(),
+          /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_032: [ If `disconnect` is called while while the transport is connected but idle, it shall call `PollingTransport.disconnect` and call it's `callback` passing the results of the transport operation. ] */
+          disconnect: (callback) => this._fsm.transition('disconnecting', callback),
+          register: (request, callback) => this._fsm.transition('sendingRegistrationRequest', request, callback)
         },
         sendingRegistrationRequest: {
           _onEnter: (request, callback) => {
+            let timeoutTimer: any = setTimeout(() => {
```
pierreca: >any [](start = 30, length = 3)

nit: not needed

provisioning/device/src/polling_state_machine.ts
```diff
@@ -25,41 +26,43 @@ export class  PollingStateMachine extends EventEmitter {
...
+            /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_036: [ If `PollingTransport.registrationRequest` does not call its callback within `ProvisioningDeviceConstants.defaultTimeoutInterval` ms, register shall with with a `TimeoutError` error. ] */
+            if (this._currentOperationCallback === callback) {
```
pierreca: indentation

provisioning/device/src/polling_state_machine.ts
```diff
@@ -139,18 +148,34 @@ export class  PollingStateMachine extends EventEmitter {
...
           },
           register: (request, callback) => callback(new errors.InvalidOperationError('another operation is in progress'))
         },
         polling: {
           _onEnter: (request, operationId, pollingInterval, callback) => {
+            /* Codes_SRS_NODE_PROVISIONING_TRANSPORT_STATE_MACHINE_18_037: [ If `PollingTransport.queryOperationStatus` does not call its callback within `ProvisioningDeviceConstants.defaultTimeoutInterval` ms, register shall with with a `TimeoutError` error. ] */
+            let timeoutTimer: any = setTimeout(() => {
+              debug('timeout while query');
+              if (this._currentOperationCallback === callback) {
+                debug('legit');
```
pierreca: >debug('legit') [](start = 16, length = 14)

those logs end up being what we use with customers in github issues. more detailed explanation?

PR #196, opened 2018-01-10
create createCertificate.js

FabianKLS:
Helpful for creating x509 certificates for individual enrollments of device in device provisioning service. Is inline with example given in the device part of the provisioning documentation.

 

# Checklist

# Description of the problem

# Description of the solution


provisioning/service/samples/createCertificate.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
```
pierreca: Please add the standard Microsoft copyright/license header is missing:

pierreca:
little nitpicky for sure but our file names are usually not camel-cased and instead snake_cased.

provisioning/service/samples/createCertificate.js
```diff
@@ -0,0 +1,31 @@
...
+var pem = require('pem');
+var fs = require('fs');
+var certPath = __dirname + "\\cert\\";
```
pierreca: >__dirname + "\\cert\\"; [](start = 15, length = 23)

I would encourage using path.join or path.resolve instead to automatically deal with cross-platform issues.

pierreca:
Thanks a lot for submitting this pull-request - it'll help users for sure. 
can you also update the readme with instructions on how/why use it?

A couple additional things to consider:
- since this is turning into sort of a command line utility, I would encourage you to look at yargs, and as an example what we've done in build/tools/create_device_certs.js - that makes it easy to create a CLI and "help" in case someone is "using it wrong"
- I'm not sure if this is meant to be a sample or a helper tool - it's fine in the samples folder for now but would be glad to hear if we need sort-of a "tools" folder.

provisioning/service/samples/createCertificate.js
FabianKLS: Added

FabianKLS:
I've made the changes your requested. Thanks for the feedback. The tool is a helper tool for the provisioning service. I wasn't aware of the build folder and the tools there. What is build/tools/create_device_certs.js exactly doing? There seems to be overlap.

pierreca:
The `create_device_certs.js` script creates self-signed certificates (without setting the common name to the device id though) and adds the certificate thumbprint to the registry. it is used for end-to-end tests.

would you mind renaming your file to `create_certificates.js` (snake_case instead of camelCase) too? with that I think that'd be all and we could pull that change in.

provisioning/service/samples/createCertificate.js
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) Microsoft. All rights reserved.
```
BertKleewein: >Copyright [](start = 3, length = 9)

Rats!  I didn't see this PR until just now after I submitted a PR that included a similar tool.  See [https://github.com/Azure/azure-iot-sdk-node/pull/198](https://github.com/Azure/azure-iot-sdk-node/pull/198)

pierreca: yeah at that point we probably need to merge these tools.

And then consider merging this into iothub-explorer.

FabianKLS:
I've renamed the file.

provisioning/service/samples/createCertificate.js
BertKleewein: Disagree on putting this into iothub-explorer.  This is just a wrapper around PEM and has no actual connection to the hub code.  Also, these certs are test-only.  We don't want to get into the business of telling people how to make production certs.  They should have their own process for that.  

---
In reply to: [161137256](https://github.com/Azure/azure-iot-sdk-node/pull/196#discussion_r161137256) [](ancestors = 161137256)

pierreca: fair enough - thanks for chiming in :) 

---
In reply to: [161297082](https://github.com/Azure/azure-iot-sdk-node/pull/196#discussion_r161297082) [](ancestors = 161297082,161137256)

pierreca:
@FabianKLS would you mind reviewing #198 and see if it meets the need that you were fulfilling with this PR too? we're going to need other parts of #198 so we're going to pull it anyway, the question is does it make sense to have your script too?

FabianKLS:
I've used the tool and it's working for me. So this request can be closed/removed.
Only 'request' for #198 is the ability to provide a folder where the certificates are stored.  After 198 is published I will see if I can add this. Oke?

pierreca:
Thanks @FabianKLS. #198 has been merged. whenever you want to add the folder feature, just go for it and @BertKleewein will review it. Closing this one.

PR #199, opened 2018-01-17
create new cert for each run of service E2E tests

BertKleewein:
I've hit some cases where the _service_create_delete.js tests would fail because a previous run left a group alive and there can't be 2 groups that use the same cert.  My fix is to create a new cert for each run.


BertKleewein:
Updated with a new revision with a hardcoded delay to deal with clock drift.  

FWIW, openssl.exe does not let you pass a beginDate when creating a cert.  You can pass this when creating a CA, but only if you use "openssl ca", but PEM doesn't support this parameter.  

reference: https://security.stackexchange.com/questions/31607/how-to-set-not-before-value-to-past-when-creating-certificate-request

PR #203, opened 2018-01-19
Sample takes parameters from the commandline now.

FabianKLS:
I've modified the device sample to take parameters from the commandline, instead of editing the file. 
A user doesn't need to edit register_x509.js to run the example. 

Second, the default protocol is http, but it's possible to change this from the commandline via the --protocol switch.

Thirds; I've made global.azure-devices-provisioning.net as the default provisioning host.

 

# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/provisioning/device/samples/register_x509.js
+++ b/provisioning/device/samples/register_x509.js
@@ -1,26 +1,70 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
 'use strict';
-
-var fs = require('fs');
-var Transport = require('azure-iot-provisioning-device-http').Http;
-
-// Feel free to change the preceding using statement to anyone of the following if you would like to try another protocol.
-// var Transport = require('azure-iot-provisioning-device-amqp').Amqp;
-// var Transport = require('azure-iot-provisioning-device-amqp').AmqpWs;
-// var Transport = require('azure-iot-provisioning-device-mqtt').Mqtt;
-// var Transport = require('azure-iot-provisioning-device-mqtt').MqttWs;
-
 var X509Security = require('azure-iot-security-x509').X509Security;
 var ProvisioningDeviceClient = require('azure-iot-provisioning-device').ProvisioningDeviceClient;
 
-var provisioningHost = '[provisioning host]';
-var idScope = '[id scope]';
-var registrationId = '[registration id]';
+var fs = require('fs');
+var argv = require('yargs')
+  .usage('Usage: $0  --idscope <ID SCOPE> --registrationid <REGISTRATION ID> --certificate <COMMONNAME> --protocol <PROTOCOL>')
+  .option('idscope', {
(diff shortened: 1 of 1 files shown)


provisioning/device/samples/register_x509.js
```diff
@@ -1,34 +1,78 @@
...
+  .usage('Usage: $0  --idscope <ID SCOPE> --registrationid <REGISTRATION ID> --certificate <COMMONNAME> --protocol <PROTOCOL>')
+  .option('idscope', {
+    alias: 'i',
+    describe: 'The ID scope is assigned to a Device Provisioning Service',
+    type: 'string',
+    demandOption: true
+  })
+  .option('registrationid', {
+    alias: 'r',
+    describe: 'Used to uniquely identify a device in the Device Provisioning Servic',
```
pierreca: >Servic [](start = 77, length = 6)

typo (missing "e")

provisioning/device/samples/register_x509.js
```diff
@@ -1,34 +1,78 @@
...
+    type: 'string',
+    demandOption: true
+  })
+  .option('certificate', {
+    alias: 'c',
+    describe: 'Commonname of the certificate without extension. Expected filenames: commonName_cert.pem and commonName_key.pem ',
```
pierreca: >commonName_cert.pem and commonName_key.pem [](start = 84, length = 42)

i'm not sure about the value of this convention over having simply 2 parameters.

pierreca:
Thanks for the pull-request.
Aside from a typo or that discussion about arguments, I'd like to understand how you use the samples and why you feel it's better to have everything in arguments, especially because we tend to think of samples as quick, easy to read code that demonstrates how to use an API. In this case, it feels pulling these changes would be detrimental to the goal since there's now 70 lines of glue code for < 10 lines of "useful" demonstration of the register API

My old take at this is that the approach you're offering works fine for people who want to test a feature first, but not necessarily how the API works: is that what you're going for, or are you thinking of something else?

FabianKLS:
I'm -and my peers- are using the sample to quickly demo the possibilities of Iot and Azure. The initial sample didn't had much explanation on how to use. Personally I took me some time to get all the right params for the device scripts (like idscope) I agree that the amount of lines of code is now a bit to much irt the code that really matters.

pierreca:
@FabianKLS Thanks for the feedback!
would it help for us to split samples and demos?

I could see demos going the direction you're suggesting: easy scripts to run with command line parameters and "prove" functionality, while we'd improve samples with more documentation and comments but keeping them simple and to the point.

would that still solve the problem and make the best of both worlds?

FabianKLS:
Hi @pierreca, sorry for the late response. I was a couple of days off. I think your suggestion is great. To have a 'demo' part for quickly creating demo's or prototypes. Happy to help here.

pierreca:
Great :) I'm closing the PR then - and we can start a discussion over email. let's come up with a list of simple feature demos you'd like to see and let's make this happen.

PR #216, opened 2018-01-31
Update SaslTpm object based on code review feedback

BertKleewein:
Changing SaslTpm implementation so the SaslTpm object knows how to construct its own buffers based off of information passed in by the Amqp transport object.

### DIFF
--- a/provisioning/transport/amqp/src/sasl_tpm.ts
+++ b/provisioning/transport/amqp/src/sasl_tpm.ts
@@ -8,8 +8,8 @@ import * as Promise from 'bluebird';
 import * as dbg from 'debug';
 const debug = dbg('azure-iot-provisioning-device-amqp:SaslTpm');
 
-export type ChallengeResponseCallback = (err: Error, response?: Buffer) => void;
-export type GetResponseFromChallenge = (challenge: Buffer, callback: ChallengeResponseCallback) => void;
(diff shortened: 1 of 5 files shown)


provisioning/transport/amqp/src/sasl_tpm.ts
```diff
@@ -8,45 +8,56 @@ import * as Promise from 'bluebird';
...
+    let init: Buffer = new Builder()
+        .appendUInt8(0)
+        .appendString(this._idScope)
+        .appendUInt8(0)
+        .appendString(this._registrationId)
+        .appendUInt8(0)
+        .appendBuffer(this._endorsementKey)
+        .get();
+
     return new Promise((resolve: any) => {
```
pierreca: Since there is no way this fails, you can just write it like so:
```js
return Promise.resolve(<the response object>);
```
and you don't have to create a new promise that way

provisioning/transport/amqp/test/_amqp_test.js
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 'use strict';
+/*jshint esversion: 6 */
```
pierreca: why do we need this?

PR #223, opened 2018-02-06
Add some unit tests for the security client.

anthonyvercolano:
# Description of the problem

More unit tests are needed for the security client.
# Description of the solution
 
This adds some unit tests for the security client.  Still have a way to go.

### DIFF
--- a/provisioning/device/samples/register_tpm.js
+++ b/provisioning/device/samples/register_tpm.js
@@ -3,7 +3,7 @@
 
 'use strict';
 
-var provisioningTransport = require('azure-iot-provisioning-device-http').Http;
+var ProvisioningTransport = require('azure-iot-provisioning-device-http').Http;
 var iotHubTransport = require('azure-iot-device-mqtt').Mqtt;
 var Client = require('azure-iot-device').Client;
 var Message = require('azure-iot-device').Message;
@@ -29,6 +29,7 @@ var idScope = '<the id scope for your dps instance>';
 
 var securityClient = new tpmSecurity.TpmSecurityClient();
 
+var provisioningClient = ProvisioningDeviceClient.create(provisioningHost, idScope, new ProvisioningTransport(), securityClient);
 // Register the device.  Do not force a re-registration.
 provisioningClient.register(function(err, result) {
   if (err) {
--- a/security/tpm/src/tpm.ts
+++ b/security/tpm/src/tpm.ts
@@ -104,11 +104,6 @@ export class TpmSecurityClient  {
                 this._fsm.handle('activateIdentityKey', identityKey, callback);
               }
             });
-          },
-          disconnect: (callback) => {
-            if (callback) {
-              callback();
-            }
           }
         },
         connecting: {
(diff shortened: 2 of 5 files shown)


provisioning/device/samples/register_tpm.js
```diff
@@ -29,6 +29,7 @@ var idScope = '<the id scope for your dps instance>';
 
 var securityClient = new tpmSecurity.TpmSecurityClient();
 
+var provisioningClient = ProvisioningDeviceClient.create(provisioningHost, idScope, new provisioningTransport(), securityClient);
```
pierreca: nit (fix later): constructors should be pascal-cased (ProvisioningTransport)

security/tpm/src/tpm.ts
```diff
@@ -288,8 +285,28 @@ export class TpmSecurityClient  {
...
+    });
+  }
+
+
   private _signData(dataToSign: Buffer, callback: (err: Error, signedData: Buffer) => void): void {
 
+    /*Codes_SRS_NODE_TPM_SECURITY_CLIENT_06_013: [If `signWithIdentity` is invoked without a previous successful invocation of `activateIdentityKey`, the callback will be invoked with `err` of `InvalidOperationError`] */
+    if (!this._idKeyPub) {
+      callback(new errors.InvalidOperationError('activateIdentityKey must be invoked before any signing is attempted.'), null);
+      return;
```
pierreca: nit: the `,null` is unnecessary and you could in theory just `return callback(new errors...)` that way it doesn't look like something happens after the callback (even if it's just a return)

security/tpm/test/_tpm_test.js
```diff
@@ -97,11 +100,30 @@ describe('tpm', function () {
       testFalsyArg('signWithIdentity', 'dataToSign', falsyDataToSign, ReferenceError);
     });
 
-    it('must call activateIdentityKey first', function() {
-      /*Tests_SRS_NODE_TPM_SECURITY_CLIENT_06_013: [If `signWithIdentity` is invoked without a previous successful invocation of `activateIdentityKey`, an InvalidOperationError is thrown.] */
+    it('must call activateIdentityKey first', function(done) {
+      /*Tests_SRS_NODE_TPM_SECURITY_CLIENT_06_013: [If `signWithIdentity` is invoked without a previous successful invocation of `activateIdentityKey`, the callback will be invoked with `err` of `InvalidOperationError`.] */
```
pierreca: so I guess "previous successful invocation" now applies to the lifecycle of the TPM but not the TPM security client state machine itself, right?

provisioning/device/samples/register_tpm.js
anthonyvercolano: Indeed.

security/tpm/src/tpm.ts
anthonyvercolano: Well as it was written, the callback required signedData to be returned.  Added the ?

security/tpm/test/_tpm_test.js
anthonyvercolano: Yes, somewhere in the depths of time, activateIdentityKey needs to be called.

PR #226, opened 2018-02-08
fix(e2etests): properly detach Event Hubs receivers in e2e tests

pierreca:
# Description of the problem
Some tests are unstable because the event hubs client does not automatically detach links before it disconnects.

# Description of the solution
Keep a list of open receivers and detach them before closing the client.

### DIFF
--- a/e2etests/test/d2c_disconnect.js
+++ b/e2etests/test/d2c_disconnect.js
@@ -108,8 +108,7 @@ var protocolAndTermination = [
 protocolAndTermination.forEach( function (testConfiguration) {
   describe(testConfiguration.transport.name + ' using device/eventhub clients - disconnect d2c', function () {
     this.timeout(20000);
-    var deviceClient, ehClient, senderInterval;
-    var provisionedDevice;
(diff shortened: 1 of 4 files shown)


anthonyvercolano:
>       this.timeout(120000);

Needed.

---
Refers to: e2etests/test/device_service.js:201 in 1ae3d34. [](commit_id = 1ae3d345dc474e724d4a8731a67849f0c9673460, deletion_comment = False)

pierreca:
>       this.timeout(120000);

it extends the timeout from the default 20,000 (set in the describe statement) to 120,000 and yes since it's a message of maximum size it takes a bit of time to get there sometimes...

---
In reply to: [364180875](https://github.com/Azure/azure-iot-sdk-node/pull/226#issuecomment-364180875) [](ancestors = 364180875)

---
Refers to: e2etests/test/device_service.js:201 in 1ae3d34. [](commit_id = 1ae3d345dc474e724d4a8731a67849f0c9673460, deletion_comment = False)

PR #227, opened 2018-02-08
Adding connection string support for modules.  SRS for other modules …

BertKleewein:
1. Adding moduleId connection string support
2. Adding SRS for device client and MQTT transport changes for modules.

---
This change is [Reviewable](https://reviewable.io/reviews/azure/azure-iot-sdk-node/227)


common/transport/mqtt/devdoc/mqtt_base_requirements.md
```diff
@@ -103,4 +103,6 @@ The `disconnect` method closes the connection to the server.
 
 **SRS_NODE_COMMON_MQTT_BASE_16_035: [** The `updateSharedAccessSignature` method shall call the `callback` argument with no parameters if the operation succeeds. **]**
 
-**SRS_NODE_COMMON_MQTT_BASE_16_036: [** The `updateSharedAccessSignature` method shall call the `callback` argument with an `Error` if the operation fails. **]**
\ No newline at end of file
+**SRS_NODE_COMMON_MQTT_BASE_16_036: [** The `updateSharedAccessSignature` method shall call the `callback` argument with an `Error` if the operation fails. **]**
+
+
```
pierreca: nit: no changes (only blank lines)

device/core/devdoc/device_client_requirements.md
```diff
@@ -91,19 +91,27 @@ client.sendEvent(new Message('hello world'), print);
 **SRS_NODE_DEVICE_CLIENT_16_058: [** The `close` method shall immediately call the `closeCallback` function if provided and the transport is already disconnected. **]**
 
 #### sendEvent(message, sendEventCallback)
+#### sendEvent(outputName, message, sendEventCallback)
 The `sendEvent` method sends an event message to the IoT Hub as the device indicated in the constructor argument.
+If an ouptut name is provided then the message is routed to the given output.
```
pierreca: >ouptut [](start = 6, length = 6)

nit: typo (output)

device/core/devdoc/device_client_requirements.md
```diff
@@ -91,19 +91,27 @@ client.sendEvent(new Message('hello world'), print);
...
 **SRS_NODE_DEVICE_CLIENT_16_047: [** The `sendEvent` method shall not throw if the `sendEventCallback` is not passed. **]**
 
 #### sendEventBatch(messages, sendEventBatchCallback)
+#### sendEventBatch(outputName, messages, sendEventBatchCallback)
 The `sendEventBatch` method sends a list of event messages to the IoT Hub as the device indicated in the constructor argument.
+If an ouptut name is provided then the message is routed to the given output.
 
 **SRS_NODE_DEVICE_CLIENT_16_082: [** The `sendEventBatch` method shall throw a `NotImplementedError` if the transport doesn't have that feature. **]**
 
+**SRS_NODE_DEVICE_CLIENT_18_011: [** if `outputName` is provided, `sendEventBatch` shall populate all of `messages` `outputName` fields with this name before possing it to the transport. **]**
```
pierreca: >possing [](start = 158, length = 7)

typo (passing)

device/transport/mqtt/devdoc/mqtt_requirements.md
```diff
@@ -73,6 +30,14 @@ The `connect` method initializes a connection to an IoT hub.
 
 **SRS_NODE_DEVICE_MQTT_12_004: [** The `connect` method shall call the `connect` method on `MqttBase`. **]**
 
+**SRS_NODE_DEVICE_MQTT_18_029: [** If a `moduleId` is not specified, the `connect` method shall use a `clientId` of "<deviceId>" when connecting to the MQTT service. **]**
+
+**SRS_NODE_DEVICE_MQTT_18_030: [** If a `moduleId` is not specified, the `connect` method shall use a `username` of "<host>/<deviceId>/DeviceClientType=<userAgent>&apiVersion=<apiVersion>". **]**
+
+**SRS_NODE_DEVICE_MQTT_18_031: [** If a `moduleId` is specified, The `connect` method shall use a `clientId` of "<deviceId>/<moduleId>" when connecting to the MQTT service. **]**
```
pierreca: >specified [](start = 54, length = 9)

nit: specified where? in the TransportConfig passed to the constructor?

device/transport/mqtt/devdoc/mqtt_requirements.md
```diff
@@ -341,3 +312,67 @@ The `getTwinReceiver` method creates a `MqttTwinReceiver` object for the twin re
...
+}
+```
+**]**
+
+### inputMessage event
+
+**SRS_NODE_DEVICE_MQTT_18_033: [** When the `inputMessage` event is subscribed to, if no `moduleId` was specified through the `constructor`, `Mqtt` shall throw an `InvalidOperationError` exception. **]**
+
+**SRS_NODE_DEVICE_MQTT_18_034: [** When the `inputMessage` event is subscribed to, `Mqtt` shall subscribe to the "devices/<deviceId>/modules/<moduleId>/#" topic **]**
+
```
pierreca: I would avoid that - the reason is that event subscription doesn't have a callback so there's no way to feed the error back to the user if the call to subscribe fail. I would rather see a pattern like the enable/disable functions that is being called from the main client (enableModuleInput/disableModuleInput for example)

device/core/devdoc/device_client_requirements.md
```diff
@@ -265,6 +273,20 @@ interface DeviceMethodEventHandler {
 
 **SRS_NODE_DEVICE_CLIENT_16_066: [** The client shall emit an error if connecting the transport fails while subscribing to message events **]**
 
+#### inputMessage
```
pierreca: >inputMessage [](start = 5, length = 12)

I'm not sure how I feel about this:
1. if we have a specific property, why not use the regular message event?
2. if we have a specific property, why not use the regular sendEvent method?

device/core/devdoc/device_client_requirements.md
```diff
@@ -91,19 +91,27 @@ client.sendEvent(new Message('hello world'), print);
 **SRS_NODE_DEVICE_CLIENT_16_058: [** The `close` method shall immediately call the `closeCallback` function if provided and the transport is already disconnected. **]**
 
 #### sendEvent(message, sendEventCallback)
+#### sendEvent(outputName, message, sendEventCallback)
```
pierreca: >sendEvent [](start = 5, length = 9)

if we must have a special sendEvent method (instead of using the message property) I wonder if we should use a different method name since overloading in javascript/typescript can be a pain.
sendEventTo or sendTo for example (i'm not a fan of the Event terminology.. we talk about telemetry, d2c... we rarely use event anywhere I think. would have to double check with other SDKs.

device/core/devdoc/device_client_requirements.md
```diff
@@ -91,19 +91,27 @@ client.sendEvent(new Message('hello world'), print);
...
 The `sendEvent` method sends an event message to the IoT Hub as the device indicated in the constructor argument.
+If an ouptut name is provided then the message is routed to the given output.
 
 **SRS_NODE_DEVICE_CLIENT_05_002: [** The `sendEvent` method shall send the event (indicated by the `message` argument) via the transport associated with the Client instance. **]**
 
+**SRS_NODE_DEVICE_CLIENT_18_010: [** if `outputName` is provided, `sendEvent` shall populate `message.outputName` with this name before possing it to the transport. **]**
```
pierreca: >outputName [](start = 102, length = 10)

never a big fan of modifying objects that come from the user - can't we pass the outputName to the send method on the transport and let the transport annotate its transport-specific message object accordingly (if necessary?)

pierreca:
Changes for the connection string look good, ship it!

Regarding send/receive, i'm not sure why we have this duality of message property + special send method (or special event).

Either we use a property on the message and we use the existing methods, or we use a specific method and in that case we probably shouldn't modify the message object.
IF this is how other module SDKs are designed, I guess there are reasons and I'd love to hear them :)

Issue #206, opened 2018-01-23
Is there an event triggered if twin state is lost


wmmihaa:
# Context
- **OS and version used:** Ubuntu Core 16.04
- **Node.js version:** 7.10
- **npm version:** 5.5.1
- **We are connected using celluar network, and currently using AMQP**

# Description of the issue:
It seems twin state is less resilient to network interrupts, and from time to time we loose the ability to receive state. Is there an event, indicating we've lost state "connection"? This is somewhat related to the #136 issue

pierreca:
That is a correct assessment of the current situation and we have an item high in our backlog to fix this, likely during February.

wmmihaa:
Thanks @pierreca.  
Let me know if there is ANYTHING we can do to help. I can send you syslogs or other metrix/logs. Just let me know.

pierreca:
@wmmihaa I've just checked in a fix for this - twins can now retry just like any other operation. 
I've discovered a couple other minor things in the twin code that I will strive to fix into another pull request or 2 before I release new packages. stay tuned!

wmmihaa:
Great news.
Will you be available at the mvp summit this week? If so it’d be great to meet up and have a beer.

________________________________

pierreca:
@wmmihaa sure, with pleasure. I’m around on campus. Would thursday work? Kinda busy tomorrow and Wednesday night.

pierreca:
Fix has been published!
https://github.com/Azure/azure-iot-sdk-node/releases/tag/2018-3-9

yzhong94:
@wmmihaa , thank you for your contribution to our open-sourced project!  Please help us improve by filling out this 2-minute customer satisfaction survey.

PR #236, opened 2018-02-24
Refactoring the twin feature to enable retries: MQTT Transport

pierreca:
#206 

# Description of the problem
Because of the way the interface between the transport and the device client is built, it's hard if not impossible to cleanly infuse the retry logic in the twin feature - this has been missing for a while.

# Description of the solution
Refactor the interface between the client and the transport.


device/transport/mqtt/devdoc/mqtt_requirements.md
```diff
@@ -300,17 +260,17 @@ The `getTwinReceiver` method creates a `MqttTwinReceiver` object for the twin re
...
+**SRS_NODE_DEVICE_MQTT_16_057: [** `enableTwinDesiredPropertiesUpdates` shall connect the MQTT connection if it is disconnected. **]**
 
-**SRS_NODE_DEVICE_MQTT_16_058: [** `enableTwin` shall calls its callback with an `Error` object if it fails to connect. **]**
+**SRS_NODE_DEVICE_MQTT_16_058: [** `enableTwinDesiredPropertiesUpdates` shall calls its callback with an `Error` object if it fails to connect. **]**
 
-**SRS_NODE_DEVICE_MQTT_16_059: [** `enableTwin` shall subscribe to the MQTT topics for twins. **]**
+**SRS_NODE_DEVICE_MQTT_16_059: [** `enableTwinDesiredPropertiesUpdates` shall subscribe to the MQTT topics for twins. **]**
 
-**SRS_NODE_DEVICE_MQTT_16_060: [** `enableTwin` shall call its callback with no arguments when the `SUBACK` packet is received. **]**
+**SRS_NODE_DEVICE_MQTT_16_060: [** `enableTwinDesiredPropertiesUpdates` shall call its callback with no arguments when the `SUBACK` packet is received. **]**
```
anthonyvercolano: There seems to be an implication in the first requirement that multiple topics are subscribed to.  However, the second requirement seems to imply only one ack.  Puzzled.

device/transport/mqtt/src/mqtt_twin_client.ts
```diff
@@ -0,0 +1,266 @@
...
+    super();
+    /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_18_001: [The `MqttTwinClient` constructor shall accept a `client` object.]*/
+    this._mqtt = client;
+
+    const messageHandler = this._onMqttMessage.bind(this);
+
+    /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_001: [The `MqttTwinClient` constructor shall immediately subscribe to the `message` event of the `client` object.]*/
+    this._mqtt.on('message', messageHandler);
+
+    this._topicFsm = new machina.BehavioralFsm({
```
anthonyvercolano: >BehavioralFsm [](start = 33, length = 13)

Interesting choice.  What does switching to the Behavioral give you?

device/transport/mqtt/src/mqtt_twin_client.ts
```diff
@@ -0,0 +1,266 @@
...
+            if (callback) {
+              callback(err);
+            }
+          },
+          subscribe: function (client: TopicSubscription, callback: (err?: Error) => void): void {
+            this.transition(client, 'subscribing', callback);
+          },
+          unsubscribe: function (client: TopicSubscription, callback: (err?: Error) => void): void {
+            // not entirely sure about that. if subscription are restored because cleanSession is false, it means technically a user may want to unsubscribe
+            // even though subscribe hasn't been called yet.
```
anthonyvercolano: Indeed.  I am pondering this issue in another issue.

device/transport/mqtt/src/mqtt_twin_client.ts
```diff
@@ -0,0 +1,266 @@
...
+            callback();
+          }
+        },
+        subscribing: {
+          _onEnter: function (client: TopicSubscription, callback: (err?: Error) => void): void {
+            client.mqttClient.subscribe(client.topic, { qos: 0 }, (err, result) => {
```
anthonyvercolano: >client [](start = 12, length = 6)

Is this a manifestation of using Behavioral?

device/transport/mqtt/devdoc/mqtt_requirements.md
```diff
@@ -67,6 +67,8 @@ The `Mqtt` and `MqttWs` constructors initialize a new instance of the MQTT trans
 
 **SRS_NODE_DEVICE_MQTT_18_025: [** If the `Mqtt` constructor receives a second parameter, it shall be used as a provider in place of mqtt.js **]**
 
+**SRS_NODE_DEVICE_MQTT_16_081: [** The `Mqtt` constructor shall subscribe to the `MqttTwinClient` `twinDesiredPropertiesUpdates` and reemit them. **]**
```
BertKleewein: >reemit them. [](start = 133, length = 12)

Could this be better phrased?  The verb reemit is not particularly clear.  Maybe the reemit needs to be a separate SRS?

device/transport/mqtt/devdoc/mqtt_requirements.md
```diff
@@ -214,67 +216,25 @@ interface DeviceMethodResponse {
...
+The `getTwin` method is used to retrieve the device twin.
 
-**SRS_NODE_DEVICE_MQTT_16_032: [** If `sendTwinRequest` is called while `MqttBase` is disconnecting, it shall wait until the disconnection is complete and then try to connect again and send the twin request. **]**
+**SRS_NODE_DEVICE_MQTT_16_075: [** `getTwin` shall establish the MQTT connection by calling `connect` on the `MqttBase` object if it is disconnected. **]**
 
-**SRS_NODE_DEVICE_MQTT_16_033: [** The `sendTwinRequest` method shall call its callback with an error translated using `translateError` if `MqttBase` fails to connect. **]**
+**SRS_NODE_DEVICE_MQTT_16_076: [** `getTwin` shall call its callback with an error if it fails to connect the transport **]**
 
-### getTwinReceiver(done)
+**SRS_NODE_DEVICE_MQTT_16_077: [** `getTwin` shall call the `getTwin` method on the `MqttTwinClient` object and pass it its callback. **]**
```
BertKleewein: >on the `MqttTwinClient` object [](start = 77, length = 30)

<comment only> keeping twin code in a separate client object like this is interesting.  I'm curious to see how this works out and wondering if this is a pattern we should use for other  features & transports (e.g. methods)

device/transport/mqtt/devdoc/mqtt_requirements.md
```diff
@@ -300,17 +260,17 @@ The `getTwinReceiver` method creates a `MqttTwinReceiver` object for the twin re
...
+### enableTwinDesiredPropertiesUpdates
 
-**SRS_NODE_DEVICE_MQTT_16_057: [** `enableTwin` shall connect the MQTT connection if it is disconnected. **]**
+**SRS_NODE_DEVICE_MQTT_16_057: [** `enableTwinDesiredPropertiesUpdates` shall connect the MQTT connection if it is disconnected. **]**
 
-**SRS_NODE_DEVICE_MQTT_16_058: [** `enableTwin` shall calls its callback with an `Error` object if it fails to connect. **]**
+**SRS_NODE_DEVICE_MQTT_16_058: [** `enableTwinDesiredPropertiesUpdates` shall calls its callback with an `Error` object if it fails to connect. **]**
 
-**SRS_NODE_DEVICE_MQTT_16_059: [** `enableTwin` shall subscribe to the MQTT topics for twins. **]**
+**SRS_NODE_DEVICE_MQTT_16_059: [** `enableTwinDesiredPropertiesUpdates` shall subscribe to the MQTT topics for twins. **]**
```
BertKleewein: >** `enableTwinDesiredPropertiesUpdates` shall subscribe to the MQTT topics for twins. **]** [](start = 32, length = 91)

This is inconsistent with your requirements above.  The new requirements above are about calling functions in other objects.  This one is about actually subscribing, which implies calling the base transport from here, which seems broken.

device/transport/mqtt/devdoc/mqtt_requirements.md
```diff
@@ -332,12 +292,12 @@ The `getTwinReceiver` method creates a `MqttTwinReceiver` object for the twin re
...
 **SRS_NODE_DEVICE_MQTT_16_046: [** `disableMethods` shall call its callback with an `Error` if an error is received while unsubscribing. **]**
 
-### disableTwin
+### disableTwinDesiredPropertiesUpdates
 
-**SRS_NODE_DEVICE_MQTT_16_062: [** `disableTwin` shall call its callback immediately if the MQTT connection is already disconnected. **]**
+**SRS_NODE_DEVICE_MQTT_16_062: [** `disableTwinDesiredPropertiesUpdates` shall call its callback immediately if the MQTT connection is already disconnected. **]**
 
-**SRS_NODE_DEVICE_MQTT_16_063: [** `disableTwin` shall unsubscribe from the topics for twin messages. **]**
+**SRS_NODE_DEVICE_MQTT_16_063: [** `disableTwinDesiredPropertiesUpdates` shall unsubscribe from the topics for twin messages. **]**
```
BertKleewein: >shall unsubscribe  [](start = 73, length = 18)

same comment.  Does this function call the mqtt twin object, or does it call the base transport?  I hope it's calling the twin object.

device/transport/mqtt/devdoc/mqtt_twin_client_requirements.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Non-responsibilities
+- connecting/disconnecting the Mqtt connection
+- any other feature that is not twin-related
+- retries and timeouts.
+
+## Public API
+```typescript
+class MqttTwinClient extends EventEmitter {
+  constructor(client: MqttBase);
```
BertKleewein: >MqttBase [](start = 22, length = 8)

Architectural question: If this object accepts an MqttBase object, that has a weak implication that the MqttTwinClient object has some sort of independence from the Mqtt object.  I find that interesting, but this is adding another level of nested FSM and that's starting to get confusing to me. 
I don't think I'm explaining this too well, but this object has an FSM, but that FSM makes assumptions about the state of the base transport.  e.g. this FSM assumes that the base is connected, do device/mqtt creates the base object, and is responsible for connecting it, but this object uses it.
Still sorting this out....smells bad though.

device/transport/mqtt/src/mqtt_twin_client.ts
```diff
@@ -0,0 +1,266 @@
...
+  getTwin(callback: (err?: Error, twin?: TwinProperties) => void): void {
+    /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_002: [The `getTwin` method shall subscribe to the `$iothub/twin/res/#` topic if not already subscribed.]*/
+    this._topicFsm.subscribe(this._responseTopic, (err) => {
+      if (err) {
+        /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_009: [If subscribing to the response topic fails, the callback shall be called with the translated version of the error obtained by using the `translateError` method of the `azure-iot-mqtt-base` package.]*/
+        callback(translateError(err));
+      } else {
+        /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_003: [The `getTwin` method shall publish the request message on the `$iothub/twin/get/?rid=<requestId>` topic using the `MqttBase.publish` method.]*/
+        /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_006: [The request message published by the `getTwin` method shall have an empty body.]*/
+        this._sendTwinRequest('GET', '/', ' ', callback);
```
anthonyvercolano: >('GET', '/', ' ', callback) [](start = 29, length = 27)

A while back I had a pull request here to deal with the '/' parameter.  It would be nice to fix this now (maybe).  (It would have removed the use of '/' here.  Perhaps we should discuss this?

device/transport/mqtt/src/mqtt_twin_client.ts
```diff
@@ -0,0 +1,266 @@
...
+    /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_011: [The `updateTwinReportedProperties` method shall subscribe to the `$iothub/twin/res/#` topic if not already subscribed.]*/
+    this._topicFsm.subscribe(this._responseTopic, (err) => {
+      if (err) {
+        /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_012: [If subscribing to the response topic fails, the callback shall be called with the translated version of the error obtained by using the `translateError` method of the `azure-iot-mqtt-base` package.]*/
+        callback(translateError(err));
+      } else {
+        /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_013: [The `updateTwinReportedProperties` method shall publish the request message on the `$iothub/twin/patch/properties/reported/?rid=<requestId>` topic using the `MqttBase.publish` method.]*/
+        /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_016: [The body of the request message published by the `updateTwinReportedProperties` method shall be a JSON string of the reported properties patch.]*/
+        this._sendTwinRequest('PATCH', '/properties/reported/', JSON.stringify(patch), callback);
```
anthonyvercolano: >/ [](start = 60, length = 1)

And here is another change it  would be nice to get through with this '/'.

device/transport/mqtt/devdoc/mqtt_twin_client_requirements.md
```diff
@@ -0,0 +1,95 @@
...
+
+**SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_002: [** The `getTwin` method shall subscribe to the `$iothub/twin/res/#` topic if not already subscribed. **]**
+
+**SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_009: [** If subscribing to the response topic fails, the callback shall be called with the translated version of the error obtained by using the `translateError` method of the `azure-iot-mqtt-base` package. **]**
+
+**SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_003: [** The `getTwin` method shall publish the request message on the `$iothub/twin/get/?rid=<requestId>` topic using the `MqttBase.publish` method. **]**
+
+**SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_004: [** The `getTwin` method shall publish the request message with QoS=0, DUP=0 and Retain=0. **]**
+
+**SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_005: [** The `requestId` property in the topic querystring should be set to a random unique integer that will be used to identify the response later on. **]**
```
BertKleewein: >integer [](start = 130, length = 7)

could also be a guid.  that's what I should have done originally and that's what other SDKs do.

device/transport/mqtt/src/mqtt_twin_client.ts
BertKleewein: There is also lore and history here around this issue.  IIRC, the service team wanted us to _not_ unsubscribe.  If we do unsubscribe here, then the service client class that updates the desired properties will fail of the device is offline.  Keeping the topic subscribed allows the service API call to succeed.  (no kidding)  Unless they changed this behavior while I was looking the other way.

---
In reply to: [170587012](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170587012) [](ancestors = 170587012)

BertKleewein: Wait, my comment actually applies to the "unsubscribing" block below, and not here, but it still stands.

---
In reply to: [170698604](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170698604) [](ancestors = 170698604,170587012)

device/transport/mqtt/src/mqtt_twin_client.ts
```diff
@@ -0,0 +1,266 @@
...
+          subscribe: function (client: TopicSubscription, callback: (err?: Error) => void): void {
+            callback();
+          },
+          unsubscribe: function (client: TopicSubscription, callback: (err?: Error) => void): void {
+            this.transition(client, 'unsubscribing', callback);
+          }
+        },
+        unsubscribing: {
+          _onEnter: function (client: TopicSubscription, callback: (err?: Error) => void): void {
+            client.mqttClient.unsubscribe(client.topic, (err, result) => {
```
BertKleewein: >unsubscribe [](start = 30, length = 11)

See comment above on unsubscribing.  It may also be that this comment applies only to the PATCH topic and not to the RES topic, but it's so goofy that I can't be confident about anything I'm saying.  IIRC, Max is your man.

device/transport/mqtt/src/mqtt_twin_client.ts
BertKleewein: >client.mqttClient.subscribe(client.topic [](start = 12, length = 40)

Ouch.  why alias this._mqtt to client.mqttClient? client.client.subscribe(client.topic) sounds like a 50's bubblegum song.  (client client bo-bient, banana fana fo-fient, fee fei fo-fient, client!)

As the code stands today, it would be clearer if you made _responseTopic and _desiredPropertiesUpdateTopic into strings and just skipped that little client object altogether

anthonyvercolano: Not sure that song ever made it overseas.

---
In reply to: [170700995](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170700995) [](ancestors = 170700995)

device/transport/mqtt/src/mqtt_twin_client.ts
```diff
@@ -0,0 +1,266 @@
...
+            });
+          },
+          '*': function (client: TopicSubscription): void { this.deferUntilTransition(client); }
+        },
+        subscribed: {
+          _onEnter: function (client: TopicSubscription, callback: (err?: Error) => void): void {
+            callback();
+          },
+          subscribe: function (client: TopicSubscription, callback: (err?: Error) => void): void {
+            callback();
```
BertKleewein: >callback [](start = 12, length = 8)

Now I'm utterly confused.  If I'm reading this correctly, this object can only be subscribed to a single topic at a time, but it looks like there's only one and it handles both RES and PATCH topics.  What is this thing?  (still stuck on the fact that this is a client that takes one kind of client as a constructor arg, passes around a different kind of client that has a pointer to another client,  which comes back to this client.  < core dump>) :))

device/transport/mqtt/src/mqtt_twin_client.ts
anthonyvercolano: Well this depends on who is calling this level.  The service does indeed NOT want us to unsubscribe when when disconnect/close.  In general this could lead to a pattern when something "above" us here re-connects, it might actually want to unsubscribe to a topic without having done a subscribe first.  Now whether this will happen since this is somewhat of a private interface ....

---
In reply to: [170699009](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170699009) [](ancestors = 170699009,170698604,170587012)

device/transport/mqtt/devdoc/mqtt_twin_client_requirements.md
BertKleewein: The more I think about it, I'm not sure why twin_client is separated out.  Are you doing it to keep your .ts file smaller?  If so, this feels like a partial class.  Encapsulation just feels strange here.  

---
In reply to: [170618901](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170618901) [](ancestors = 170618901)

e2etests/test/twin_e2e_tests.js
```diff
@@ -190,27 +190,38 @@ delete nullMergeResult.tweedle;
...
+        } else if (props.$version === 2) {
+          assertObjectsAreEqual(newProps, deviceTwin.properties.desired);
+          done();
+        } else {
+          done(new Error('incorrect property version received - ' + props.$version));
+        }
+      });
 
       serviceTwin.update( { properties : { desired : newProps } }, function(err) {
         if (err) return done(err);
```
BertKleewein: scary re-ordering of test.   I'm not saying this change is bad, but why did you need to change this?  (I'm wondering if this is related to the unsubscribe)

device/transport/mqtt/src/mqtt_twin_client.ts
```diff
@@ -0,0 +1,266 @@
...
+    const messageHandler = this._onMqttMessage.bind(this);
+
+    /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_001: [The `MqttTwinClient` constructor shall immediately subscribe to the `message` event of the `client` object.]*/
+    this._mqtt.on('message', messageHandler);
+
+    this._topicFsm = new machina.BehavioralFsm({
+      initialState: 'unsubscribed',
+      states: {
+        unsubscribed: {
+          _onEnter: function (client: TopicSubscription, err: Error, callback: (err?: Error) => void): void {
```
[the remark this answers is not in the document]
BertKleewein: Also confused.  Why is this called a "client" (a term that is so overly used in our code that it's meaningless), when I think "topic" might be a better name since this code is subscribing to a topic.

---
In reply to: [170588449](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170588449) [](ancestors = 170588449)

device/transport/mqtt/devdoc/mqtt_requirements.md
pierreca: yeah i've been going back and forth on this - we do it in the AMQP transport too. On one hand it allows to have a separate state machine per feature and is keeping the code simple to understand for one given feature, on the other hand I don't like the dependency that these features have onto the 'connected' state of the main state machine and the fact that we have to wrap the methods in the main client. We could also have those "feature-clients" depend on the main client but then we'd have to wrap the MqttBase methods. i'm open to suggestions here, even though we'll probably address this outside of this PR

---
In reply to: [170613498](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170613498) [](ancestors = 170613498)

device/transport/mqtt/devdoc/mqtt_requirements.md
pierreca: true, i think it should just say "calls the enableTwinDEsiredPropertiesUpdate on the twin client"

---
In reply to: [170614894](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170614894) [](ancestors = 170614894)

device/transport/mqtt/devdoc/mqtt_requirements.md
pierreca: twin object. i'll change the requirement.

---
In reply to: [170615308](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170615308) [](ancestors = 170615308)

device/transport/mqtt/devdoc/mqtt_twin_client_requirements.md
pierreca: yes and yes. see my other comment. there's definitely a smell to this and I'm open to suggestions.

---
In reply to: [170704664](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170704664) [](ancestors = 170704664,170618901)

device/transport/mqtt/devdoc/mqtt_twin_client_requirements.md
pierreca: I thought it had to be integers!! yes I like GUIDs better let me double check everything with the service 'cause i don't like generating random ints.

---
In reply to: [170696059](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170696059) [](ancestors = 170696059)

device/transport/mqtt/src/mqtt_twin_client.ts
pierreca: the ability to store the state outside of the state machine logic (in the topic objects) which lets me use the same state machine for 2 topics.

---
In reply to: [170586642](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170586642) [](ancestors = 170586642)

device/transport/mqtt/src/mqtt_twin_client.ts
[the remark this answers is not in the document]
pierreca: I just got lazy and used whatever the machina.js doc used. I'll change it. topic sounds good.

---
In reply to: [170712943](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170712943) [](ancestors = 170712943,170588449)

device/transport/mqtt/src/mqtt_twin_client.ts
pierreca: I'll get to the bottom of this, thanks for pointing it out. may or may not be solved with this PR though.

---
In reply to: [170703095](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170703095) [](ancestors = 170703095,170699009,170698604,170587012)

device/transport/mqtt/src/mqtt_twin_client.ts
pierreca: ... you have excellent code comments and poor musical taste. I hope the renaming helped a little bit.
Unfortunately we can't use strings in place of objects because the client/topicSubscription object is actually used to store the state of the behavioral state machine

---
In reply to: [170701724](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170701724) [](ancestors = 170701724,170700995)

device/transport/mqtt/src/mqtt_twin_client.ts
pierreca: see [http://machina-js.org/](http://machina-js.org/) in the Going Further - Behavioral FSM section. The client/topicSubscription object is used to store the state of the state machine. we have 1 single state machine definition used by 2 topics, independently. that's why we carry this object around.

---
In reply to: [170703083](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170703083) [](ancestors = 170703083)

device/transport/mqtt/src/mqtt_twin_client.ts
pierreca: I will double check.

---
In reply to: [170699475](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170699475) [](ancestors = 170699475)

device/transport/mqtt/src/mqtt_twin_client.ts
pierreca: yep - one of the points of this refactoring is to address this issue. since the resource paths are now stored in the transport there is no need to worry about that anymore. mqtt will do its thing and amqp will do its thing, the device client won't have to care about that anymore.

---
In reply to: [170687317](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170687317) [](ancestors = 170687317)

e2etests/test/twin_e2e_tests.js
pierreca: the reason is that if you fire the update first, and then have the device subscribe, then the update to desired properties might happen before your subscription, which means that you'll never get an event for that - subscribing first and then calling serviceTwin.update makes sure we're subscribed first (well - sortof. since we don't wait on the SUBACK because deviceTwin.on is synchronous.... still. better this way i think.

---
In reply to: [170712640](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r170712640) [](ancestors = 170712640)

device/transport/mqtt/src/mqtt_twin_client.ts
pierreca: apparently we should unsubscribe when we disconnect.

---
In reply to: [171442744](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r171442744) [](ancestors = 171442744,170703095,170699009,170698604,170587012)

device/transport/mqtt/src/mqtt_twin_client.ts
pierreca: according to Robledo we should unsubscribe.

---
In reply to: [171443628](https://github.com/Azure/azure-iot-sdk-node/pull/236#discussion_r171443628) [](ancestors = 171443628,170699475)

pierreca:
superseded by #239

PR #237, opened 2018-02-28
CR feedback and unit tests for MQTT modules.

BertKleewein:
This is larger than I thought it would be.  

I did the following:
1) Removed the onInputMessage() and removeInputMessageHandler() functions, opting for on('inputMessage', function(inputName, msg)) instead.
2) Added enableInputMessages and disableInputMessages at the transport level.
3) Diverged code paths for sendEvent and sendOutputEvent.  A module can do both Event and OutputEvent.  Likewise, a device can do both.  
4) Diverged code paths for on('message') and on('inputMessage').
5) Added unit test coverage.

I have mixed feelings about this change.  It feels like I've added a lot of complexity that maybe isn't necessary.

### DIFF
--- a/common/transport/mqtt/devdoc/mqtt_base_requirements.md
+++ b/common/transport/mqtt/devdoc/mqtt_base_requirements.md
@@ -38,7 +38,7 @@ The `Mqtt` constructor receives the configuration parameters to configure the MQ
 
 ### MqttBase.connect(config, done)
 The `connect` method establishes a connection with the server using the config object passed in the arguments.
-**SRS_NODE_COMMON_MQTT_BASE_16_006: [** The `connect` method shall throw a ReferenceError if the config argument is falsy, or if one of the following properties of the config argument is falsy: deviceId, host, and one of sharedAccessSignature or x509.cert and x509.key. **]**
+**SRS_NODE_COMMON_MQTT_BASE_16_006: [** The `connect` method shall throw a ReferenceError if the config argument is falsy, or if one of the following properties of the config argument is falsy: uri, clientId, username, and one of sharedAccessSignature or x509.cert and x509.key. **]**
 
 **SRS_NODE_COMMON_MQTT_BASE_16_002: [** The `connect` method shall use the authentication parameters contained in the `config` argument to connect to the server. **]**
 
--- a/common/transport/mqtt/test/_mqtt_base_test.js
+++ b/common/transport/mqtt/test/_mqtt_base_test.js
@@ -14,9 +14,24 @@ var Message = require('azure-iot-common').Message;
 var endpoint = require('azure-iot-common').endpoint;
 var errors = require('azure-iot-common').errors;
 
+
+
+
+
 describe('MqttBase', function () {
+  var fakeConfig;
+
+  beforeEach(function() {
+    fakeConfig = {
+      uri: 'uri',
+      clientId: 'clientId',
+      username: 'username',
+      sharedAccessSignature: "sasToken"
+    }
+  });
+
   describe('#connect', function () {
-    /*Tests_SRS_NODE_COMMON_MQTT_BASE_16_006: [The `connect` method shall throw a ReferenceError if the config argument is falsy, or if one of the following properties of the config argument is falsy: deviceId, host, and one of sharedAccessSignature or x509.cert and x509.key.]*/
+    /*Tests_SRS_NODE_COMMON_MQTT_BASE_16_006: [The `connect` method shall throw a ReferenceError if the config argument is falsy, or if one of the following properties of the config argument is falsy: uri, clientId, username, and one of sharedAccessSignature or x509.cert and x509.key.]*/
     it('throws if config structure is falsy', function () {
       [null, undefined, '', 0].forEach(function (config) {
         var mqtt = new MqttBase('test', new FakeMqtt());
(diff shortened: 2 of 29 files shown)


common/transport/mqtt/devdoc/mqtt_base_requirements.md
```diff
@@ -38,7 +38,7 @@ The `Mqtt` constructor receives the configuration parameters to configure the MQ
 
 ### MqttBase.connect(config, done)
 The `connect` method establishes a connection with the server using the config object passed in the arguments.
-**SRS_NODE_COMMON_MQTT_BASE_16_006: [** The `connect` method shall throw a ReferenceError if the config argument is falsy, or if one of the following properties of the config argument is falsy: deviceId, host, and one of sharedAccessSignature or x509.cert and x509.key. **]**
+**SRS_NODE_COMMON_MQTT_BASE_16_006: [** The `connect` method shall throw a ReferenceError if the config argument is falsy, or if one of the following properties of the config argument is falsy: uri, clientId, username, and one of sharedAccessSignature or x509.cert and x509.key. **]**
```
pierreca: >throw [](start = 67, length = 5)

change is good but as a general remark, since these APIs are internal and we use typescript now, I would rather see strong typing than runtime checks in the future. Don't be surprised if you see me removing them in my PRs :)

common/transport/mqtt/test/_mqtt_base_test.js
```diff
@@ -14,9 +14,24 @@ var Message = require('azure-iot-common').Message;
 var endpoint = require('azure-iot-common').endpoint;
 var errors = require('azure-iot-common').errors;
 
+
+
+
+
```
pierreca: ok super nitpicky but watch out for unnecessary whitespace? I know, i'm nitpicky. sorry.

common/transport/mqtt/test/_mqtt_base_test.js
```diff
@@ -26,31 +41,15 @@ describe('MqttBase', function () {
...
-          host: "host.name",
-          deviceId: deviceId,
-          sharedAccessSignature: "sasToken"
-        };
-        var mqtt = new MqttBase(config, 'test', new FakeMqtt());
-        assert.throws(function () {
-          mqtt.connect(config, function () {});
-        }, ReferenceError, 'Invalid transport configuration');
+    ['uri', 'clientId', 'username'].forEach(function(falsyProperty) {
+      [null, undefined].forEach(function (falsyValue) {
```
pierreca: >undefined [](start = 13, length = 9)

empty string too? ''?

device/core/src/client.ts
```diff
@@ -65,6 +65,7 @@ export class Client extends EventEmitter {
   private blobUploadClient: BlobUploadClient; // TODO: wrong casing/naming convention
   private _c2dEnabled: boolean;
   private _methodsEnabled: boolean;
+  private _inputMessagesEnabled: boolean;
```
pierreca: >_inputMessagesEnabled [](start = 10, length = 21)

note for later: there's definitely a pattern emerging here. wondering if the BehavioralFsm would help.

device/transport/mqtt/devdoc/mqtt_requirements.md
```diff
@@ -348,3 +417,22 @@ interface MethodMessage {
 ```
 **]**
 
+### inputMessage event
+
+**SRS_NODE_DEVICE_MQTT_18_057: [** If there is a listener for the `inputMessage` event, a `inputMessage` event shall be emitted for each message received. **]**
```
pierreca: >If there is a listener for the `inputMessage` event, [](start = 35, length = 52)

you can emit even if there are no listeners, it is not going to crash. not sure if there's value if that first "if" statement since we're using enable/disable anyway?

device/transport/mqtt/test/_mqtt_receiver_test.js
```diff
@@ -329,5 +330,24 @@ describe('Mqtt as MqttReceiver', function () {
...
+    /*Tests_SRS_NODE_DEVICE_MQTT_18_058: [ When an `inputMessage` event is received, Mqtt shall extract the inputName from the topic according to the following convention: 'devices/<deviceId>/modules/<moduleId>/inputs/<inputName>' ]*/
+    /*Tests_SRS_NODE_DEVICE_MQTT_18_056: [ When an `inputMessage` event is emitted, the first parameter shall be the inputName and the second parameter shall be of type `Message`. ]*/
+    describe('#inputMessage', function() {
+        it('emits an inputMessage event with a Message object when there is a listener', function (done) {
+          fakeConfig.moduleId = 'moduleId';
+          var receiver = new Mqtt(fakeAuthenticationProvider, fakeMqttBase);
+          receiver.connect(function () {
+            receiver.on('inputMessage', function (inputName, msg) {
+              assert.strictEqual(inputName, 'fakeInputName')
+              assert.instanceOf(msg, Message);
```
pierreca: >assert.instanceOf(msg, Message); [](start = 14, length = 32)

is that working? you're not passing a message in the emit?

device/transport/mqtt/test/_mqtt_test.js
```diff
@@ -794,3 +923,9 @@ describe('Mqtt', function () {
     });
   });
 });
+
+
+
+
+
+
```
pierreca: nit: whitespaaace :-/

device/transport/mqtt/test/_mqtt_receiver_test.js
BertKleewein: Yep, this was copy/paste from the corresponding message test

---
In reply to: [171429827](https://github.com/Azure/azure-iot-sdk-node/pull/237#discussion_r171429827) [](ancestors = 171429827)

PR #240, opened 2018-03-06
refactor(amqp): use AmqpMessage instead of Message in ReceiverLink

pierreca:
# Description of the problem
in the AMQP transport `SenderLink` manipulates `AmqpMessage` objects but `ReceiverLink` operates with transport-agnostic `Message` object. this asymetry makes it complicated for features relying on `ReceiverLink` objects to inspect the message annotations etc. 

# Description of the solution
- Change `ReceiverLink` APIs to use the internal `AmqpMessage` type
- Change classes depending on `ReceiverLink` to do the `Message` translation themselves
- Create a `ServiceReceiver` class that wraps the `ReceiverLink` in the Service SDK so as not to break compatibility while isolating the `ReceiverLink` from the public API (and allow the changes)

*note: code coverage is going down in the service because there are areas of the generated javascript (From typescript) that are not tested at the beginning of the file - not something we care about*

### DIFF
--- a/device/transport/amqp/src/amqp_twin_client.ts
+++ b/device/transport/amqp/src/amqp_twin_client.ts
@@ -6,7 +6,7 @@
 import { EventEmitter } from 'events';
 import machina = require('machina');
 
-import { endpoint, Message, AuthenticationProvider } from 'azure-iot-common';
+import { endpoint, AuthenticationProvider } from 'azure-iot-common';
 import { Amqp as BaseAmqpClient, AmqpMessage, SenderLink, ReceiverLink } from 'azure-iot-amqp-base';
 import { TwinProperties } from 'azure-iot-device';
 
@@ -43,7 +43,7 @@ export class AmqpTwinClient extends EventEmitter {
   private _fsm: any;
   private _pendingTwinRequests: { [key: string]: (err: Error, twinProperties?: any) => void };
 
-  private _messageHandler: (message: Message) => void;
+  private _messageHandler: (message: AmqpMessage) => void;
   private _errorHandler: (err: Error) => void;
 
   constructor(authenticationProvider: AuthenticationProvider, client: any) {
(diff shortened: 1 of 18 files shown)


device/transport/amqp/src/amqp_twin_client.ts
```diff
@@ -289,17 +289,17 @@ export class AmqpTwinClient extends EventEmitter {
...
     /*Codes_SRS_NODE_DEVICE_AMQP_TWIN_16_021: [The `updateTwinReportedProperties` method shall monitor `Message` objects on the `ReceiverLink.on('message')` handler until a message with the same `correlationId` as the one that was sent is received.]*/
     /*Codes_SRS_NODE_DEVICE_AMQP_TWIN_16_029: [The `enableTwinDesiredPropertiesUpdates` method shall monitor `Message` objects on the `ReceiverLink.on('message')` handler until a message with the same `correlationId` as the one that was sent is received.]*/
     /*Codes_SRS_NODE_DEVICE_AMQP_TWIN_16_034: [The `disableTwinDesiredPropertiesUpdates` method shall monitor `Message` objects on the `ReceiverLink.on('message')` handler until a message with the same `correlationId` as the one that was sent is received.]*/
-    if (this._pendingTwinRequests[message.correlationId]) {
-      const pendingRequestCallback = this._pendingTwinRequests[message.correlationId];
-      delete this._pendingTwinRequests[message.correlationId];
+    if (this._pendingTwinRequests[message.properties.correlationId]) {
```
anthonyvercolano: >message.properties. [](start = 34, length = 19)

You took great care up at line 61 to determine that the message had properties before you accessed the correlationId.  No so here?

service/src/service_receiver.ts
```diff
@@ -0,0 +1,54 @@
...
+    /*Codes_SRS_NODE_SERVICE_RECEIVER_16_004: [The `abandon` method shall call the `abandon` method on the `ReceiverLink` object and pass it the `AmqpMessage` stored within the `transportObj` property of the `Message` object as well as the `done` callback passed as argument.]*/
+    this._receiver.abandon(message.transportObj, done);
+  }
+
+  reject(message: Message, done?: Client.Callback<results.MessageRejected>): void {
+    /*Codes_SRS_NODE_SERVICE_RECEIVER_16_005: [The `reject` method shall call the `reject` method on the `ReceiverLink` object and pass it the `AmqpMessage` stored within the `transportObj` property of the `Message` object as well as the `done` callback passed as argument.]*/
+    this._receiver.reject(message.transportObj, done);
+  }
+
+  detach(callback: (err?: Error) => void): void {
```
anthonyvercolano: >detach [](start = 2, length = 6)

I was surprised that I didn't see users of the detach in the service client needing to be adjusted in some small way.

device/transport/amqp/src/amqp_twin_client.ts
pierreca: well, we don't make the function call if we don't have one. so no need here, right?

---
In reply to: [172687346](https://github.com/Azure/azure-iot-sdk-node/pull/240#discussion_r172687346) [](ancestors = 172687346)

service/src/service_receiver.ts
pierreca: how so?

---
In reply to: [172690211](https://github.com/Azure/azure-iot-sdk-node/pull/240#discussion_r172690211) [](ancestors = 172690211)

PR #241, opened 2018-03-06
add edge module that can run in Docker, push CA down from app to TLS

BertKleewein:
Adding a sample that shows a docker container running an Edge module written in Node.js.  

To make this work, I needed to use the edgeHub's CA certificate to establish TLS between the module and edgeHub.  I added some plumbing to take the CA from the application all the way down to mqtt_base.

My one question centers around _how_ I add the CA in edge_sample_module.js.  Getting the config object from the auth provider and editing it before we connect seems skeevy.  Should we have a setOptions property that we can use to pass the CA down.

A possible moment of confusion will come from the fact that we're using SAK authentication and passing an X509 cert down to the transport.  This is completely valid.  The X509 CA cert that we pass down is used to validate the TLS connection between the module and the edgeHub.

Another moment of confusion will come from the fact that node.js doesn't use the trusted cert store that the OS provides.  Instead, node.js has a hardcoded cert store, so we can't simply set the edgeHub CA cert in the trusted store because Node doesn't use that. :(

### DIFF
--- a/common/core/src/authorization.ts
+++ b/common/core/src/authorization.ts
@@ -76,6 +76,10 @@ export interface TransportConfig {
    * IP address or internet name of the host machine working as a device or protocol gateway.  Used when communicating with Azure Edge devices.
    */
   gatewayHostName?: string;
+  /**
+   * Public certificate for the CA to use when validating the connection to the server
+  */
+  ca?: string;
 }
 
 /**
@@ -109,4 +113,5 @@ export interface X509 {
    * @private
    */
   keyFile?: string;
+
 }
--- a/device/edge-sample/edge_sample_module.js
+++ b/device/edge-sample/edge_sample_module.js
@@ -1,0 +1,66 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+
+var Protocol = require('azure-iot-device-mqtt').Mqtt;
+var Client = require('azure-iot-device').Client;
+var Message = require('azure-iot-device').Message;
+var SharedAccessKeyAuthenticationProvider = require('azure-iot-device').SharedAccessKeyAuthenticationProvider;
+var fs = require('fs');
+
+var connectionString = process.env.EdgeHubConnectionString;
+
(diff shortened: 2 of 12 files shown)


common/core/src/authorization.ts
```diff
@@ -109,4 +113,5 @@ export interface X509 {
    * @private
    */
   keyFile?: string;
+
 }
```
anthonyvercolano: Thought I'd nit the empty line before Pierre did.

anthonyvercolano:
>    */

nitting this empty line....

---
Refers to: common/core/src/authorization.ts:109 in 23a50e9. [](commit_id = 23a50e9baddf7a4c37253ee1ef3081f6b2685748, deletion_comment = False)

common/core/src/authorization.ts
```diff
@@ -76,6 +76,10 @@ export interface TransportConfig {
    * IP address or internet name of the host machine working as a device or protocol gateway.  Used when communicating with Azure Edge devices.
    */
   gatewayHostName?: string;
+  /**
+   * Public certificate for the CA to use when validating the connection to the server
+  */
+  ca?: string;
```
pierreca: wouldn't this thing belong in the x509 options? or are we keeping some x509 options for iot hub and others for edge?

device/edge-sample/edge_sample_module.js
```diff
@@ -0,0 +1,66 @@
...
+var fs = require('fs');
+
+var connectionString = process.env.EdgeHubConnectionString;
+
+var authProvider = SharedAccessKeyAuthenticationProvider.fromConnectionString(connectionString);
+authProvider.getDeviceCredentials(function(err, credentials) {
+  if (err) {
+    throw new Error('unexpected: getDeviceCredentials failure');
+  } else {
+    credentials.ca = fs.readFileSync(process.env.EdgeModuleCACertificateFile).toString('ascii');
```
pierreca: >credentials.ca  [](start = 4, length = 15)

I agree with your "skeevy" comment seeing this here. I think we have to figure out a way for the auth provider to return authentication for IoT Hub AND the edge hub. maybe with a completely new authentication provider type?

Issue #219, opened 2018-02-05
Is it possible to set contentType and encoding?


wmmihaa:
When using IoT Hub Routes we need to set **contentType** and **encoding** of the message or we will otherwise not be able to route the message to the appropriate end-point.

[The C# sample](https://azure.microsoft.com/en-us/blog/iot-hub-message-routing-now-with-routing-on-message-body/]) is pretty straight forward, but we can't get it to work for node-js: 

```
// C# - Working
using (var message = new Message(messageBytes)) 
{ 
    // Set message body type and content encoding. 
    message.ContentEncoding = "utf-8"; 
    message.ContentType = "application/json"; 
 
    // Add other custom application properties.  
    message.Properties["Status"] = "Active";  // Routing property
 
    await deviceClient.SendEventAsync(message); 
}
```

```
// JavaScript - Not working
message.contentEncoding = "utf-8"; 
message.contentType = "application/json"; 
message.properties.add("Status", "Active");  // Routing property
client.sendEvent(message);
```

We are currently using AMQP, but can switch to MQTT should it be required...

wmmihaa:
Should I understand, - tagging the issue "enhancement" means: -"No it's not supported at the moment"?

pierreca:
@wmmihaa it means it's not implemented yet and is on our backlog (as opposed to a bug that happens on a feature already implemented)

jimoneil:
FWIW, I posted at the [blog announcement from May](https://azure.microsoft.com/en-us/blog/iot-hub-message-routing-now-with-routing-on-message-body/) that this is not working.  It's not completely clear if this is AMQP only or MQTT.  I would recommend an update to the blog post since it is one of the more definitive resources for this feature.

pierreca:
This has been implemented and released now:
https://github.com/Azure/azure-iot-sdk-node/releases/tag/2018-3-9

yzhong94:
@jimoneil @wmmihaa , thank you for your contribution to our open-sourced project!  Please help us improve by filling out this 2-minute customer satisfaction survey.

virtualamjath:
can you provide sample to setup content Type and Encoding Type on message 

// JavaScript - Not working
message.contentEncoding = "utf-8"; 
message.contentType = "application/json"; 
message.properties.add("Status", "Active");  // Routing property
client.sendEvent(message);

this wont work

PR #242, opened 2018-03-07
feat(device): Add support for contentType and contentEncoding properties

pierreca:
#219 

# Description of the problem
Routing messages using the content of the body requires the ability for users to set the `contentType` and `contentEncoding` properties on the `Message`object. These properties have not been implemented nor has the code to encode them in different protocols

# Description of the solution
- Add `contentType` and `contentEncoding` properties to the `Message`
- Translate them into the appropriate formats for MQTT, AMQP and HTTP

### DIFF
--- a/common/core/src/message.ts
+++ b/common/core/src/message.ts
@@ -46,6 +46,16 @@ export class Message {
   ack: string;
 
   /**
+   * Content type property used to routes with the message body. Should be 'application/json'.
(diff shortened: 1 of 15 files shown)


virtualamjath:
can you provide me sample for while sending in data in MQTT protocol and Http to set the contetType and Encoding

PR #248, opened 2018-03-15
Make popper

anthonyvercolano:
# Description of the problem

It's a bit of a pain to do proof of possession for x509 certs.
# Description of the solution
 
Very simple app not for use in production that will use supplied signing cert and private key (in pem files) and create a signed cert with the supplied nonce as the subject.


pierreca:
1. change the name (use a search engine to find out why)
2. before you check this in, please look at what @BertKleewein did in `/provisioning/tools` and see if these couldn't be merged. we're starting to have cert-generating tools all over the place.

anthonyvercolano:
Just use the provisioning tools instead.  Closing this.

PR #249, opened 2018-03-15
update usage of create_test_cert.js to be more intuitive for verifica…

BertKleewein:
I took Tony's usage for his popper tool and integrated it into create_test_cert.js.  Since yargs is a bit difficult times, I made some yargs fixes along the way.

### DIFF
--- a/provisioning/tools/create_test_cert.js
+++ b/provisioning/tools/create_test_cert.js
@@ -9,57 +9,60 @@ var chalk = require('chalk');
 var camelcase = require('camelcase');
 var argv = require('yargs')
           .command('root [commonName]', 'create a root certificate', (yargs) => {
-            yargs.positional('commonName', { describe: 'name to use for cert', default: 'Test Root Cert' })
+            return yargs
+              .positional('commonName', { describe: 'name to use for cert', default: 'Test Root Cert' })
           })
           .command('intermediate <commonName> <parentCommonName>', 'create an intermediate cert', (yargs) => {
-            yargs.positional('commonName', { describe: 'name to use for cert' })
-            yargs.positional('parentCommonName', { describe: 'common name of parent cert' })
+            return yargs
+              .positional('commonName', { describe: 'name to use for cert' })
+              .positional('parentCommonName', { describe: 'common name of parent cert' });
           })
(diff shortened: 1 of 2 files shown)


provisioning/tools/create_test_cert.js
```diff
@@ -9,59 +9,62 @@ var chalk = require('chalk');
...
-            yargs.positional('commonName', { describe: 'name to use for cert' })
-            yargs.positional('parentCommonName', { describe: 'common name of parent cert' })
+            return yargs
+              .positional('commonName', { describe: 'name to use for cert' })
+              .positional('parentCommonName', { describe: 'common name of parent cert' });
           })
-          .command('verification <certToVerify> <verificationCode>', 'provide proof of possession for a cert using a verification code', (yargs) => {
-            yargs.positional('certToVerify', { describe: 'common name for certificate to verify' })
-            yargs.positional('verificationCode', { describe: 'verification code from portal' })
+          .command('verification [--ca root cert pem file name] [--key root cert key pem file name] [--nonce nonce]', 'provide proof of possession for a cert using a verification code', (yargs) => {
```
anthonyvercolano: >[--ca root cert pem file name] [](start = 33, length = 30)

The brackets make them appear to be optional.  They are not.  Note this should be fixed in the md file also.

BertKleewein: blame the makers of yargs for this.  Yargs parses this string and takes some meaning from it.  If I  enclose in <>,  yargs parses this and enforces the requirement at the top level and inside the command level.  So, I think, the .option calls consume the --key, --ca, and --nonce options, and then the .command call looks at the line and says, "hey, there's these 6 things missing.  You're missing the -ca, then you're missing the pem name, then you're missing the -key, then you're missing...blah blah blah".  At least that's the way it's behaving.  

I spent about 40 minutes trying to make this work before I figured this out.  Using brackets is the price I decided to pay to avoid spending any more time.  If you call incorrectly, the detailed usage for the command options shows up correctly, so it's a minor bug.  

---
In reply to: [174944497](https://github.com/Azure/azure-iot-sdk-node/pull/249#discussion_r174944497) [](ancestors = 174944497)

PR #254, opened 2018-03-20
fix(amqp) disconnection logic was causing "cannot read property write of null" errors

pierreca:
# Description of the problem
Because of some error in the code the build was sometimes hitting an error where something was trying to write on a socket that had been closed, causing a "cannot read property "write" of null" error.

# Description of the solution
A couple of things at play here:
1. the `amqp10` library, when it is trying to properly close the socket, emits the `disconnect` event _before_ completely closing and resetting its state machine which may lead to issues when retrying to connect before the disconnect is complete. that's what the `process.nextTick` call is for in amqp-base.
2. The error bubbled up with the `disconnect` event was swallowed, causing the SDK to try and disconnect cleanly (by trying to send detach frames on the socket even though the amqp10 library was already disconnecting and the socket dead: this caused the error previously described).
3. Looks like the C2D disconnection logic was missing a callback, causing a hang once the fix was implemented. fixed that too.
4. also shuffled a few parameters around to adhere to the convention that callbacks should always come last even in the internal state machine transition handlers.

### DIFF
--- a/device/transport/amqp/src/amqp.ts
+++ b/device/transport/amqp/src/amqp.ts
@@ -92,7 +92,7 @@ export class Amqp extends EventEmitter implements Client.Transport {
     this._amqp = baseClient || new BaseAmqpClient(false, 'azure-iot-device/' + packageJson.version);
     this._amqp.setDisconnectHandler((err) => {
       debug('disconnected event handler: ' + (err ? err.toString() : 'no error'));
-      this._fsm.handle('disconnect', () => {
+      this._fsm.handle('disconnectError', err, () => {
         this.emit('disconnect', getTranslatedError(err, 'AMQP client disconnected'));
       });
     });
@@ -137,7 +137,7 @@ export class Amqp extends EventEmitter implements Client.Transport {
       initialState: 'disconnected',
       states: {
         disconnected: {
-          _onEnter: (callback, err) => {
+          _onEnter: (err, callback) => {
             if (callback) {
               if (err) {
                 callback(err);
@@ -236,6 +236,11 @@ export class Amqp extends EventEmitter implements Client.Transport {
           disableMethods: (callback) => {
             // if we are disconnected the C2D link is already detached.
             callback();
+          },
+          disconnectError: (err, callback) => {
+            /*Codes_SRS_NODE_DEVICE_AMQP_16_080: [if the handler specified in the `setDisconnectHandler` call is called while the `Amqp` object is disconnected, the call shall be ignored.]*/
+            debug('ignoring disconnectError because already disconnected: ' + err.toString());
+            callback();
           }
         },
         connecting: {
@@ -244,7 +249,7 @@ export class Amqp extends EventEmitter implements Client.Transport {
             this._authenticationProvider.getDeviceCredentials((err, credentials) => {
               if (err) {
                 /*Codes_SRS_NODE_DEVICE_AMQP_16_055: [The `connect` method shall call its callback with an error if the callback passed to the `getDeviceCredentials` method is called with an error.]*/
-                this._fsm.transition('disconnected', connectCallback, translateError('AMQP Transport: Could not get credentials', err));
+                this._fsm.transition('disconnected', translateError('AMQP Transport: Could not get credentials', err), connectCallback);
               } else {
                 this._c2dEndpoint = endpoint.messagePath(encodeURIComponent(credentials.deviceId));
                 this._d2cEndpoint = endpoint.eventPath(credentials.deviceId);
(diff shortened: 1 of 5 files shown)


device/transport/amqp/src/amqp.ts
```diff
@@ -92,7 +92,7 @@ export class Amqp extends EventEmitter implements Client.Transport {
     this._amqp = baseClient || new BaseAmqpClient(false, 'azure-iot-device/' + packageJson.version);
     this._amqp.setDisconnectHandler((err) => {
       debug('disconnected event handler: ' + (err ? err.toString() : 'no error'));
-      this._fsm.handle('disconnect', () => {
+      this._fsm.handle('disconnectError', err, () => {
```
BertKleewein: >disconnectError [](start = 24, length = 15)

I'm struggling with this name.  Without thinking too deeply, my first reaction is that this is handling an error that occurred while disconnecting.  I _think_ you mean that it's handling a disconnect that happened because of an error, but that doesn't come though.  I can't come up with a good name off the top of my head, but I think you want an action called something like "serverInitiatedDisconnect," but I know that this description doesn't cover all cases.  Maybe "conectionDropped" or "connectionSevered" to indicate that the disconnect wasn't user initiated.  Or maybe just "nonUserInitiatedDisconnect"?

And if this handler could cover both cases (user-initiated and non-user-initiated), then I need to understand more and maybe we want to separate those two cases into two different handlers.

device/transport/amqp/src/amqp.ts
```diff
@@ -236,6 +236,11 @@ export class Amqp extends EventEmitter implements Client.Transport {
           disableMethods: (callback) => {
             // if we are disconnected the C2D link is already detached.
             callback();
+          },
+          disconnectError: (err, callback) => {
+            /*Codes_SRS_NODE_DEVICE_AMQP_16_080: [if the handler specified in the `setDisconnectHandler` call is called while the `Amqp` object is disconnected, the call shall be ignored.]*/
+            debug('ignoring disconnectError because already disconnected: ' + err.toString());
```
BertKleewein: >ignoring [](start = 19, length = 8)

does this swallow an error that we might want to forward up to higher layers?

device/transport/amqp/src/amqp.ts
```diff
@@ -297,15 +303,17 @@ export class Amqp extends EventEmitter implements Client.Transport {
               });
             }
           },
-          disconnect: (disconnectCallback) => this._fsm.transition('disconnecting', disconnectCallback),
+          disconnect: (disconnectCallback) => this._fsm.transition('disconnecting', null, disconnectCallback),
+          /*Codes_SRS_NODE_DEVICE_AMQP_16_081: [if the handler specified in the `setDisconnectHandler` call is called while the `Amqp` object is connecting or authenticating, the connection shall be stopped and an `disconnect` event shall be emitted with the error translated to a transport-agnostic error.]*/
+          disconnectError: (err, callback) => this._fsm.transition('disconnecting', err, callback),
```
BertKleewein: >disconnectError [](start = 10, length = 15)

I have a question here that comes back to the name of this action.  If this action happens because we've already been disconnected, is "disconnecting" the right state to go into?  If the network connection has already been severed, then we just want to clean up our internal housekeeping, but I don't think we want to call into amqp10 to do anything.  According to my guess, amqp already considers this disconnected, so why  risk problems by calling functions based on the assumption that parts of our connection might still be open?

device/transport/amqp/src/amqp.ts
```diff
@@ -387,10 +395,12 @@ export class Amqp extends EventEmitter implements Client.Transport {
             /*Codes_SRS_NODE_DEVICE_AMQP_16_042: [The `disableMethods` method shall call `detach` on the device method links and call its callback when these are successfully detached.]*/
             /*Codes_SRS_NODE_DEVICE_AMQP_16_043: [The `disableMethods` method shall call its `callback` with an `Error` if it fails to detach the device method links.]*/
             this._deviceMethodClient.detach(callback);
-          }
+          },
+          /*Codes_SRS_NODE_DEVICE_AMQP_16_082: [if the handler specified in the `setDisconnectHandler` call is called while the `Amqp` object is connected, the connection shall be disconnected and an `disconnect` event shall be emitted with the error translated to a transport-agnostic error.]*/
+          disconnectError: (err, callback) => this._fsm.transition('disconnecting', err, callback)
```
BertKleewein: >disconnectError [](start = 10, length = 15)

same question on what state we should transition into.

device/transport/amqp/src/amqp.ts
```diff
@@ -92,7 +92,7 @@ export class Amqp extends EventEmitter implements Client.Transport {
     this._amqp = baseClient || new BaseAmqpClient(false, 'azure-iot-device/' + packageJson.version);
     this._amqp.setDisconnectHandler((err) => {
```
anthonyvercolano: >this._amqp.setDisconnectHandler [](start = 4, length = 31)

Why do we even have this mechanism.  Should this have just been a listener?

pierreca: historical reasons. yes. out of the scope of this PR to refactor though.

---
In reply to: [175831933](https://github.com/Azure/azure-iot-sdk-node/pull/254#discussion_r175831933) [](ancestors = 175831933)

device/transport/amqp/src/amqp.ts
pierreca: it is on purpose. if we're already disconnected, we're already in the right state and the higher layers have already been notified (using the _onEnter callback calling/event emitting code). we should not be hitting that but i want to be defensive and log it just in case amqp10 does something funky. wouldn't be the first time we have a weird race condition.

---
In reply to: [175762815](https://github.com/Azure/azure-iot-sdk-node/pull/254#discussion_r175762815) [](ancestors = 175762815)

device/transport/amqp/src/amqp.ts
pierreca: yes (and that's what happens in the underlying amqp object). I guess the "disconnecting" state could be renamed to "cleaning up". if you look at the code in the `disconnecting._onEnter` you'll see that there are 2 paths depending if there's an error or not: one of them cleanly shuts down resources, the other one force-detaches everything (which doesn't send anything on the wire, it just resets state machines).

---
In reply to: [175764457](https://github.com/Azure/azure-iot-sdk-node/pull/254#discussion_r175764457) [](ancestors = 175764457)

device/transport/amqp/src/amqp.ts
pierreca: same answer then :)

---
In reply to: [175764784](https://github.com/Azure/azure-iot-sdk-node/pull/254#discussion_r175764784) [](ancestors = 175764784)

device/transport/amqp/src/amqp.ts
pierreca: amqpConnectionClosed?


---
In reply to: [175762418](https://github.com/Azure/azure-iot-sdk-node/pull/254#discussion_r175762418) [](ancestors = 175762418)

PR #257, opened 2018-03-27
Basic long-haul test harness & d2c test

pierreca:
Add a basic long-haul test harness as well as a d2c test.

The basic idea is to have intervals, timeouts and protocol configurable at build time, and spin a process or even a build for each test.

I don't know yet how this will evolve as we add more features to test (C2D, twins etc) so I kept it minimalistic. no use of test frameworks, fancy command line or anything else (yet)

### DIFF
--- a/longhaultests/src/iothub_longhaul.ts
+++ b/longhaultests/src/iothub_longhaul.ts
@@ -105,8 +105,11 @@ createDevice((err, deviceConnectionString) => {
     });
 
     const endTimeout = setTimeout(() => {
+      debug('end timeout: test completed successfully');
       stopSender(() => {
+        debug('sender stopped. deleting device');
         deleteDevice(() => {
+          debug('test device deleted from registry. Exiting.');
           process.exit(OK_EXIT_CODE);
         });
       });
@@ -118,6 +121,7 @@ createDevice((err, deviceConnectionString) => {
         debug('error starting d2c_sender: ' + err.toString());
         clearTimeout(endTimeout);
         deleteDevice(() => {
+          debug('test device deleted. Exiting.')
           process.exit(ERROR_EXIT_CODE);
         });
       } else {
--- a/longhaultests/src/iothub_longhaul.ts
+++ b/longhaultests/src/iothub_longhaul.ts
@@ -1,0 +1,128 @@
+#!/usr/bin/env node
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+import * as uuid from 'uuid';
+import * as async from 'async';
+import * as dbg from 'debug';
+const debug = dbg('longhaul:main');
+
+import { Registry, ConnectionString as HubConnectionString } from 'azure-iothub';
+import { ConnectionString as DeviceConnectionString } from 'azure-iot-device';
+
+import { D2CSender } from './d2c_sender';
+
+const MAX_EXECUTION_TIME = parseInt(process.env.MAX_EXECUTION_TIME_SECONDS) * 1000;
+const SEND_INTERVAL = parseInt(process.env.DEVICE_SEND_INTERVAL_SECONDS) * 1000;
+const SEND_TIMEOUT = parseInt(process.env.DEVICE_SEND_TIMEOUT_SECONDS) * 1000;
+
+const MAX_CREATE_TIME = 30000;
+const MAX_DELETE_TIME = 30000;
+const MAX_START_TIME = 30000;
+const MAX_STOP_TIME = 30000;
+
+const ERROR_EXIT_CODE = -1;
+const OK_EXIT_CODE = 0;
+
+const hubConnectionString = process.env.IOTHUB_CONNECTION_STRING;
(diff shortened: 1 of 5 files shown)


anthonyvercolano:
If I were to make a suggestion, whatever job you have running, set the debug environment variables to * for both our SDK and the lower level protocol transport.

longhaultests/src/iothub_longhaul.ts
```diff
@@ -0,0 +1,128 @@
...
+import * as async from 'async';
+import * as dbg from 'debug';
+const debug = dbg('longhaul:main');
+
+import { Registry, ConnectionString as HubConnectionString } from 'azure-iothub';
+import { ConnectionString as DeviceConnectionString } from 'azure-iot-device';
+
+import { D2CSender } from './d2c_sender';
+
+const MAX_EXECUTION_TIME = parseInt(process.env.MAX_EXECUTION_TIME_SECONDS) * 1000;
```
BertKleewein: >MAX_EXECUTION_TIME_SECONDS [](start = 48, length = 26)

Not sure about adding these env variables and the ones below.  I assume you're going to run these tests in a container with some container-creation script hat we haven't seen yet, and that's cool, but I'm wondering if some sort of config.json file might be a better place to hold these?  Like maybe launch the longhaul test and pass a config.json filename on the command line.  That way your execution parameters are decoupled from container creation.

longhaultests/src/iothub_longhaul.ts
```diff
@@ -0,0 +1,128 @@
...
+const MAX_STOP_TIME = 30000;
+
+const ERROR_EXIT_CODE = -1;
+const OK_EXIT_CODE = 0;
+
+const hubConnectionString = process.env.IOTHUB_CONNECTION_STRING;
+const registry = Registry.fromConnectionString(hubConnectionString);
+
+const deviceId = 'node-longhaul-sak-' + uuid.v4();
+let protocol;
```
BertKleewein: >p [](start = 4, length = 1)

Nit: this holds a constructor function, so it should be Protocol with a capital P.  (tslint has flagged this for me before, but you're not calling new() directly so you probably wouldn't see it).

longhaultests/src/iothub_longhaul.ts
```diff
@@ -0,0 +1,128 @@
...
+  break;
+  default:
+    debug('unknown protocol: ' + process.env.DEVICE_PROTOCOL);
+    process.exit(ERROR_EXIT_CODE);
+}
+/* tslint:enable:no-var-requires */
+
+const createDevice = (callback) => {
+  debug('creating device: ' + deviceId);
+  async.timeout(registry.create.bind(registry), MAX_CREATE_TIME)({ deviceId: deviceId }, (err, deviceInfo) => {
```
BertKleewein: >bind(registry) [](start = 32, length = 14)

E V I L!

BertKleewein: I'm trying to wrap my head around the necessity of the bind call here, and I'm also wondering if it would be cleaner to create all the timeout wrappers together in a separate function.  I'm thinking maybe build an entire wrapped object, like promisify does, only with async.timeout()  It wouldn't have to use forEach and wrap every method -- it could just be hardcoded functions/timeout values.  I just like the idea of having an object that looks exactly like the registry object, with the additional timeout functionality wrapping all the calls.    (side thought: would this make sense with retry?  Maybe too late now for the node SDK, but could be a useful paradigm)

---
In reply to: [177508442](https://github.com/Azure/azure-iot-sdk-node/pull/257#discussion_r177508442) [](ancestors = 177508442)

pierreca:
we'll have the ability to set this in the build definition directly.

---
In reply to: [376581181](https://github.com/Azure/azure-iot-sdk-node/pull/257#issuecomment-376581181) [](ancestors = 376581181)

longhaultests/src/iothub_longhaul.ts
pierreca: where (containers or not) it runs is the CI environment responsibility.
Regarding configuration, the reason I went with env variables is because i want to be able to set it up from the build definition page. I'll look if we can support something JSON later, for now I believe env variable are the way to go for build parameters.

---
In reply to: [177506818](https://github.com/Azure/azure-iot-sdk-node/pull/257#discussion_r177506818) [](ancestors = 177506818)

longhaultests/src/iothub_longhaul.ts
pierreca: in that case tslint will bark about it being uppercase because in this context, we're not calling it with new, and tslint rules dictate variable names to be camel cased.

---
In reply to: [177507571](https://github.com/Azure/azure-iot-sdk-node/pull/257#discussion_r177507571) [](ancestors = 177507571)

longhaultests/src/iothub_longhaul.ts
pierreca: definitely having the same thoughts on the side. I'll refactor as we "Generalize" this pattern to more features. right now i want to get it out there and get builds to run and succeed.

---
In reply to: [177514489](https://github.com/Azure/azure-iot-sdk-node/pull/257#discussion_r177514489) [](ancestors = 177514489,177508442)

PR #258, opened 2018-03-29
Add setOptions with HTTP agent and CA for TLS

BertKleewein:
1) formalize the structure for setOptions a little more.
2) Add specific options for HTTP agent & TLS CA (MQTT only for now).

### DIFF
--- a/common/transport/http/src/http.ts
+++ b/common/transport/http/src/http.ts
@@ -4,11 +4,10 @@
 'use strict';
 
 import { ClientRequest, IncomingMessage } from 'http';
-import { request, RequestOptions, Agent } from 'https';
+import { request, RequestOptions } from 'https';
 import { Message, X509 } from 'azure-iot-common';
 import dbg = require('debug');
 const debug = dbg('azure-iot-common.Http');
-const keepAliveAgent = new Agent({ keepAlive: true });
 
 /**
  * @private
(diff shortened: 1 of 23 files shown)


common/transport/http/src/http.ts
```diff
@@ -144,6 +149,14 @@ export class Http {
     return msg;
   }
 
+  /**
+   * @private
+   */
+  setOptions(options: any, callback: (err?: Error) => void): void {
```
pierreca: >setOptions(options: any, callback: (err?: Error) => void): void { [](start = 2, length = 65)

why make this asynchronous? no need now, and since it's private we can always change it later.

service/src/registry.ts
```diff
@@ -53,6 +54,9 @@ export class Registry {
     /*SRS_NODE_IOTHUB_REGISTRY_16_025: [The `Registry` constructor shall use `azure-iothub.RestApiClient` if no `restApiClient` argument is provided.]*/
     // This httpRequestBuilder parameter is used only for unit-testing purposes and should not be used in other situations.
     this._restApiClient = restApiClient || new RestApiClient(config, packageJson.name + '/' + packageJson.version);
+    if (this._restApiClient.setOptions) {
```
pierreca: >if (this._restApiClient.setOptions) { [](start = 4, length = 37)

same here

PR #266, opened 2018-04-04
Port of setOptions code from modules-preview into master.

BertKleewein:
This is a cherry-pick of one commit, plus some manual copying from modules-preview, plus the additional code to pass  an agent when connecting using MQTT-WS.


device/transport/http/src/http.ts
```diff
@@ -734,39 +739,9 @@ export class Http extends EventEmitter implements Client.Transport {
...
-   * Does not poll and instead rely on the user calling the `receive` method.
-   */
-  manualPolling?: boolean;
-  /**
-   * Boolean indicating whether only one message should be received all messages should be drained.
-   */
-  drain?: boolean;
+export interface HttpReceiverOptions extends _clientReceiverOptions {
+
 }
```
pierreca: I don't think you need this since it's not exported outside of the package.

PR #273, opened 2018-04-11
Move symlinks into local directory for easier cleanup use of multiple…

BertKleewein:
Move symlinks into local directory for easier cleanup use of multiple clones on a single machine

This moves the symlink folder under $node_root so that:
1) We don't have to call dev-teardown in the Windows jenkins job anymore.
2) We can have multiple clones and they don't collide because they don't share the same symlinks.

Note:  This allows us to keep the call to dev-teardown in our jenkins job because some jenkins builds will need it for a while (anyone except master, including the LTS branches).  If you want to force a teardown, call dev-teardown with the --force flag.

### DIFF
--- a/build/dev-teardown.sh
+++ b/build/dev-teardown.sh
@@ -5,6 +5,19 @@
 
 node_root=$(cd "$(dirname "$0")/.." && pwd)
 
+if [ -z "$1" ]; then
+  echo "Not removing any files because symlinks are in $node_root\npm-symlinks and teardown isn't necessary."
+  echo "If you want to tear down anyway, run dev-teardown.cmd with the --force flag"
+  exit 0
+fi
+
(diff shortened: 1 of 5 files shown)


build/dev-teardown.sh
```diff
@@ -5,6 +5,19 @@
...
+  echo "If you want to tear down anyway, run dev-teardown.cmd with the --force flag"
+  exit 0
+fi
+
+if [ "$1" != "--force" ]; then
+  echo "usage: dev-teardown.cmd [--force]"
+  exit 1
+fi
+
+export NPM_CONFIG_PREFIX=$node_root/npm-symlinks
```
pierreca: I don't think this is necessary. we can remove the call to dev-teardown in the build definition.

PR #274, opened 2018-04-13
fix(e2e:methods): reuse test device across device methods tests

pierreca:
# Description of the problem
device methods e2e tests create 1 device per test, which leads to some registry APIs errors

# Description of the solution
reuse the same device across the test suite (1 device / protocol)
also add logs to make it easier to debug failures later

### DIFF
--- a/e2etests/test/device_method.js
+++ b/e2etests/test/device_method.js
@@ -26,11 +26,11 @@ var hubConnectionString = process.env.IOTHUB_CONNECTION_STRING;
 ].forEach(function(protocolCtor) {
   describe('Device Methods over ' + protocolCtor.name, function() {
     this.timeout(120000);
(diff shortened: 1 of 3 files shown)


e2etests/test/device_method.js
```diff
@@ -42,33 +42,63 @@ var hubConnectionString = process.env.IOTHUB_CONNECTION_STRING;
...
+          debug('failed to delete device: ' +  deviceDescription.deviceId + ': ' + err.toString());
+          return done(err);
+        } else {
+          debug('device deleted: ' +  deviceDescription.deviceId);
+          return done();
+        }
+      });
+    });
+
+    // create a new device for every test
```
BertKleewein: >// create a new device for every test [](start = 4, length = 37)

outdated comment

PR #279, opened 2018-04-23
Modules connected to the hub can do D2C, but not C2D, inputs, or outp…

BertKleewein:
I was wrong when I did some this service API coding and E2E tests.  Modules connected to the hub can do D2C, but not C2D, inputs, or outputs.  Adjusting service API and E2E tests to account for this


service/src/client.ts
```diff
@@ -173,11 +173,10 @@ export class Client extends EventEmitter {
 
 
   /**
-   * @method            module:azure-iothub.Client#sendToModuleInput
+   * @method            module:azure-iothub.Client#sendToModule
    * @description       Sends a message to a device.
```
pierreca: >device [](start = 45, length = 6)

module?

service/src/client.ts
```diff
@@ -188,45 +187,42 @@ export class Client extends EventEmitter {
    *                              transport-specific response object useful for
    *                              logging or debugging.
    *
-   * @throws {ReferenceError}     If `deviceId`, `moduleId`, `inputName`, or `message` is null, undefined or empty.
+   * @throws {ReferenceError}     If `deviceId`, `moduleId`, or `message` is null, undefined or empty.
    */
-  sendToModuleInput(deviceId: string, moduleId: string, inputName: string, message: Message | Message.BufferConvertible, done?: Callback<results.MessageEnqueued>): void {
-    /*Codes_SRS_NODE_IOTHUB_CLIENT_18_007: [The `sendToModuleInput` method shall throw `ReferenceError` if the deviceId, moduleId, inputName, or message arguments are falsy.]*/
+  sendToModule(deviceId: string, moduleId: string, message: Message | Message.BufferConvertible, done?: Callback<results.MessageEnqueued>): void {
```
pierreca: >sendToModule [](start = 2, length = 12)

I'm confused this looks like C2D, i thought we didn't support that.

service/devdoc/iothub_client_requirements.md
```diff
@@ -109,32 +109,32 @@ The `send` method sends a cloud-to-device message to the service, intended for d
 
 **SRS_NODE_IOTHUB_CLIENT_16_030: [** The `send` method shall not throw if the `done` callback is falsy. **]**
 
-sendToModuleInput(deviceId: string, moduleId: string, inputName: string, message: Message | Message.BufferConvertible, done?: Callback<results.MessageEnqueued>): void;
-The `sendToModuleInput` method sends a message to the service, intended for delivery to the given input on the given module on the given device
+sendToModule(deviceId: string, moduleId: string, message: Message | Message.BufferConvertible, done?: Callback<results.MessageEnqueued>): void;
```
pierreca: >sendToModule [](start = 0, length = 12)

thinking out loud (sortof) ; if we're naming this "sendToModule" should we deprecate "send" and rename it "sendToDevice" ?
What are the other SDKs calling these APIs?

PR #280, opened 2018-04-23
Need feedback: user agent string v2 POC

BertKleewein:
Aargh.  I wrote a whole description in here then lost it.

This isn't a real PR.  I just want your opinion on device/transport mqtt.ts, amqp.ts, and http.ts.  Because getting the agent string requires a callback, I need to defer init of the base transports so it's outside of the device transport constructor.  This is my lazy way to do it.  How dirty does this make you feel?

### DIFF
--- a/common/core/src/utils.ts
+++ b/common/core/src/utils.ts
@@ -1,0 +1,32 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+'use strict';
+
+import * as os from 'os';
+// tslint:disable-next-line:no-var-requires
+const getos = require('getos');
+
+export function getAgentPlatformString(callback: (platformString?: string) => void): void {
+  let ver = [];
+
+  ver.push('node ' + process.version);
+
+  if (os.platform() === 'linux') {
+    getos((err, osVer) => {
+      if (err) {
+        ver.push('unknown');
+      } else {
+        // docker: { os: 'linux', dist: 'Debian', release: '8.10' }
+        // ubuntu: > { os: 'linux', dist: 'Ubuntu Linux', codename: 'xenial',  release: '16.04' }
+        ver.push(osVer.dist + ' ' +  osVer.release);
+      }
+      ver.push(os.arch());
+      callback(ver.join('; '));
+    });
+  } else {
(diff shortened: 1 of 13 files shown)


common/core/src/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+import * as os from 'os';
+// tslint:disable-next-line:no-var-requires
+const getos = require('getos');
+
+export function getAgentPlatformString(callback: (platformString?: string) => void): void {
+  let ver = [];
+
+  ver.push('node ' + process.version);
+
+  if (os.platform() === 'linux') {
```
anthonyvercolano: >'linux' [](start = 24, length = 7)

MacOS?

device/transport/amqp/src/amqp.ts
```diff
@@ -684,7 +693,9 @@ export class Amqp extends EventEmitter implements Client.Transport {
     }
 
     /*Codes_SRS_NODE_DEVICE_AMQP_16_020: [The `sendMethodResponse` response shall call the `AmqpDeviceMethodClient.sendMethodResponse` method with the arguments that were given to it.]*/
-    this._deviceMethodClient.sendMethodResponse(methodResponse, callback);
+    this._ensureBaseObject(() => {
+      this._deviceMethodClient.sendMethodResponse(methodResponse, callback);
+    });
```
anthonyvercolano: Twin too?

device/transport/mqtt/src/mqtt.ts
```diff
@@ -480,25 +489,27 @@ export class Mqtt extends EventEmitter implements Client.Transport {
     /*Codes_SRS_NODE_DEVICE_MQTT_16_015: [The `setOptions` method shall throw an `ArgumentError` if the `cert` property is populated but the device uses symmetric key authentication.]*/
     if (this._authenticationProvider.type === AuthenticationType.Token && options.cert) throw new errors.ArgumentError('Cannot set x509 options on a device that uses token authentication.');
 
-    this._mqtt.setOptions(options);
+    this._ensureBaseObject(() => {
```
anthonyvercolano: >this._ensureBaseObject [](start = 4, length = 22)

I haven't given it a lot of thought.  But, I do need to go wash up.

PR #283, opened 2018-04-25
rewrite amqpBase.connect to accept a config structure

BertKleewein:
This is a parallel change to the one I made for MQTT.  I'm doing this to prepare for user agent changes.  Instead of accepting individual parameters, I've adjusted amqpBase.connect to accept a config structure.  

Along the way, I removed the logic to automatically connect.  This mirrors the MQTT protocol and is unused anyway since the AMQP device, service, and provisioning transports all take care of doing this.

### DIFF
--- a/common/transport/amqp/package.json
+++ b/common/transport/amqp/package.json
@@ -36,7 +36,7 @@
     "alltest": "istanbul cover node_modules/mocha/bin/_mocha -- --reporter spec test/_*_test*.js",
     "ci": "npm -s run lint && npm -s run build && npm -s run alltest-min && npm -s run check-cover",
     "test": "npm -s run lint && npm -s run build && npm -s run alltest && npm -s run check-cover",
-    "check-cover": "istanbul check-coverage --statements 95 --branches 83 --functions 92 --lines 96"
+    "check-cover": "istanbul check-coverage --statements 95 --branches 82 --functions 92 --lines 96"
   },
   "engines": {
     "node": ">= 0.10"
--- a/common/transport/amqp/src/amqp.ts
+++ b/common/transport/amqp/src/amqp.ts
@@ -31,7 +31,6 @@ export type GenericAmqpBaseCallback<T> = (err: Error | null, result?: T) => void
  * @param   {String}    sdkVersionString        String identifying the type and version of the SDK used.
  */
 export class Amqp {
-  private uri: string;
   private _amqp: amqp10.Client;
   private _receivers: { [key: string]: ReceiverLink; } = {};
   private _senders: { [key: string]: SenderLink; } = {};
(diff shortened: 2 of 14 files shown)


common/transport/amqp/src/amqp.ts
```diff
@@ -222,7 +188,17 @@ export class Amqp {
             this._cbs.attach(callback);
           },
           putToken: (audience, token, callback) => {
-            this._cbs.putToken(audience, token, callback);
+            if (!this._cbs) {
```
anthonyvercolano: >if (!this._cbs) { [](start = 12, length = 17)

Really this is fine, but seems a little funny that a large part of the code that was removed and replaced above was now causing an error for performing operations out of order.  But it's ok here!

common/transport/amqp/package.json
```diff
@@ -36,7 +36,7 @@
     "alltest": "istanbul cover node_modules/mocha/bin/_mocha -- --reporter spec test/_*_test*.js",
     "ci": "npm -s run lint && npm -s run build && npm -s run alltest-min && npm -s run check-cover",
     "test": "npm -s run lint && npm -s run build && npm -s run alltest && npm -s run check-cover",
-    "check-cover": "istanbul check-coverage --statements 95 --branches 83 --functions 92 --lines 96"
+    "check-cover": "istanbul check-coverage --statements 95 --branches 82 --functions 92 --lines 96"
```
pierreca: >82 [](start = 71, length = 2)

I see small movements here and there so I don't say much because it's always one percent, but after 5 reviews, it may become 5%. There is no good reason to see these going down and if there is we need to know where that percent point is lost and why we shouldn't care.

common/transport/amqp/src/amqp.ts
```diff
@@ -136,59 +135,26 @@ export class Amqp {
...
-          attachReceiverLink: (endpoint, linkOptions, callback) => {
-            this._fsm.handle('connect', null, (err) => {
-              if (err) {
-                callback(err);
-              } else {
-                this._fsm.handle('attachReceiverLink', endpoint, linkOptions, callback);
-              }
-            });
-          },
+          attachSenderLink: (endpoint, linkOptions, callback) => callback(new errors.NotConnectedError()),
```
pierreca: >callback(new errors.NotConnectedError()) [](start = 65, length = 40)

I'm really not sure how I feel about this. On one hand, I'm all for more simple code and symmetry between MQTT and AMQP. on the other hand I wonder if the right thing is to align AMQP with MQTT and not the other way around. 
These layers should be independent. we should trust each layer to do the right thing as much as possible. I'm not going to question this that much for now especially since Tony is coming up with big changes at this level to incorporate rhea, but this feels like a philosophical change that slips through as part of an API adjustment that really should not require it (it's "just" a user agent string).

common/transport/amqp/src/amqp.ts
pierreca: Feels as if we're missing a state. maybe we should have 'connected' -> 'initializingCBS' -> 'authenticating' -> 'authenticated'

---
In reply to: [184350713](https://github.com/Azure/azure-iot-sdk-node/pull/283#discussion_r184350713) [](ancestors = 184350713)

common/transport/amqp/test/_amqp_test.js
```diff
@@ -31,24 +31,28 @@ describe('Amqp', function () {
 
     it('sets the session policy to not reestablish on failure', function () {
       var amqp = new Amqp();
+      void(amqp);
```
pierreca: nit: would these belong in beforeEach/afterEach clauses?

common/transport/amqp/src/amqp.ts
BertKleewein: I agree that it feels strange to fit in a philosophical change like this, and I feel like I dropped this on you without discussing beforehand, and I apologize for that.  

For context, here’s some of my rationale:  
1. At this layer,  we don’t know who we’re connecting to until the caller calls connect().  So, for example, it doesn’t make sense to call attachSenderLink before calling connect because we wouldn't know the destination URI.  The only time this code would work correctly is when the connection is established, then somehow disconnected.  (the device-level transport gets the destination URI at construction time, so it does make sense at that level).
2. The layers above take care of connecting before doing any of these operations, so, at best, this functionality is redundant.  At worst, this layer would be trying to circumvent the policy of upper layers, in effect deciding _when_ to reconnect.  
3. I agree that this could increase robustness by “just doing the right thing”, but it could also cover up bugs.  Because upper layers are responsible for reconnecting, this code would only come into play when the connection is broken but upper layers don’t know that it’s broken.  This seems like something that should be addressed rather than covered up by lower layers.  In effect, this would be a lower layer covering up the bugs in an upper layer, and it’s usually the other way around.
4. Because there’s no known  case where upper layers would address this functionality, the only coverage we get is at the unit test level.  As far as we know, this code _never_ gets executed in a product setting. 
5. Having 2 different layers be responsible for reconnecting just smells bad.

PR #284, opened 2018-04-30
Build and consumer user agent V2 for device SDK

BertKleewein:
build user agent string and consume it for all transports for the device SDK.

### DIFF
--- a/common/transport/amqp/package.json
+++ b/common/transport/amqp/package.json
@@ -7,16 +7,18 @@
   "main": "main.js",
   "typings": "index.d.ts",
   "dependencies": {
-    "azure-iot-common": "1.5.0",
     "amqp10": "^3.6.0",
     "amqp10-transport-ws": "^0.0.5",
     "async": "^2.5.0",
+    "azure-iot-common": "1.5.0",
     "bluebird": "^3.5.0",
     "debug": "^3.0.1",
+    "lodash.merge": "^4.6.1",
     "machina": "^2.0.1",
     "uuid": "^3.0.1"
   },
   "devDependencies": {
+    "@types/lodash.merge": "^4.6.3",
     "@types/node": "^8.0.28",
     "@types/async": "^2.0.42",
     "chai": "^4.1.2",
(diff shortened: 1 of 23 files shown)


common/transport/amqp/package.json
```diff
@@ -7,12 +7,13 @@
...
   "typings": "index.d.ts",
   "dependencies": {
-    "azure-iot-common": "1.5.0",
     "amqp10": "^3.6.0",
     "amqp10-transport-ws": "^0.0.5",
     "async": "^2.5.0",
+    "azure-iot-common": "1.5.0",
     "bluebird": "^3.5.0",
     "debug": "^3.0.1",
+    "lodash": "^4.17.10",
```
pierreca: >"lodash": "^4.17.10", [](start = 4, length = 21)

pls check we download the same version of lodash across all sdk packages

common/transport/amqp/src/amqp.ts
```diff
@@ -11,6 +11,7 @@ import { ClaimsBasedSecurityAgent } from './amqp_cbs';
 import { SenderLink } from './sender_link';
 import { ReceiverLink } from './receiver_link';
 import { AmqpLink } from './amqp_link_interface';
+import { _ } from 'lodash';
```
pierreca: >lodash [](start = 19, length = 6)

i think you can import only a subset of the package (lodash/something) to save on load times

device/core/src/utils.ts
```diff
@@ -0,0 +1,17 @@
...
+import { getAgentPlatformString } from 'azure-iot-common';
+
+// tslint:disable-next-line:no-var-requires
+const packageJson = require('../package.json');
+
+export function getUserAgentString(done: (agent: string) => void): void {
+  /*Codes_SRS_NODE_DEVICE_UTILS_18_001: [`getUserAgentString` shall call `getAgentPlatformString` to get the platform string.]*/
+  getAgentPlatformString((platformString) => {
+  /*Codes_SRS_NODE_DEVICE_UTILS_18_002: [`getUserAgentString` shall call its `callback` with a string in the form 'azure-iot-device/<packageJson.version>(<platformString>)'.]*/
+  done('azure-iot-device/' + packageJson.version + '(' + platformString + ')');
```
pierreca: >azure-iot-device [](start = 8, length = 16)

packageJson.name?

PR #286, opened 2018-05-08
fix: last put token should stop timer expiring old put tokens

anthonyvercolano:
# Description of the problem

If you open and close quite quickly, there will be a 10 second or so delay while a timer in the cbs code is queued to rundown expired put-tokens.

# Description of the solution
If the last put-token is completed stop the timer.  No point in leaving it around.

### DIFF
--- a/common/transport/amqp/src/amqp_cbs.ts
+++ b/common/transport/amqp/src/amqp_cbs.ts
@@ -139,6 +139,12 @@ export class ClaimsBasedSecurityAgent {
                         if (msg.properties.correlationId === this._putToken.outstandingPutTokens[i].correlationId) {
                           const completedPutToken = this._putToken.outstandingPutTokens[i];
                           this._putToken.outstandingPutTokens.splice(i, 1);
+                          //
+                          // If this was the last outstanding put token then get rid of the timer trying to clear out expiring put tokens.
+                          //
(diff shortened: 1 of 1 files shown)


common/transport/amqp/src/amqp_cbs.ts
```diff
@@ -139,6 +139,12 @@ export class ClaimsBasedSecurityAgent {
                         if (msg.properties.correlationId === this._putToken.outstandingPutTokens[i].correlationId) {
                           const completedPutToken = this._putToken.outstandingPutTokens[i];
                           this._putToken.outstandingPutTokens.splice(i, 1);
+                          //
+                          // If this was the last outstanding put token then get rid of the timer trying to clear out expiring put tokens.
+                          //
+                          if (this._putToken.outstandingPutTokens.length === 0) {
+                            clearTimeout(this._putToken.timeoutTimer);
```
BertKleewein: >timeoutTimer [](start = 56, length = 12)

Do you want to null this out after you clear it?

BertKleewein:
>             }

should also kill and null the timer here in case we didn't get a response before we detached.

---
Refers to: common/transport/amqp/src/amqp_cbs.ts:98 in 8b91702. [](commit_id = 8b91702fe44193334101e20591f6e085250c0614, deletion_comment = False)

PR #287, opened 2018-05-09
break Client into Client and ModuleClient

BertKleewein:
Following the C# pattern, I created an internalClient object and the Client (DeviceClient) and ModuleClient objects call into the internal client.


BertKleewein:
<ping>, this is due today.  Can someone (Tony?) review this please.

anthonyvercolano:
On it.  Sorry I missed it.

device/core/src/client.ts
```diff
@@ -1,885 +0,0 @@
...
-      throw new ReferenceError('transportCtor cannot be \'' + transportCtor + '\'');
-    }
-
-    /*Codes_SRS_NODE_DEVICE_CLIENT_16_090: [The `fromAuthenticationProvider` method shall pass the `authenticationProvider` object passed as argument to the transport constructor.]*/
-    /*Codes_SRS_NODE_DEVICE_CLIENT_16_091: [The `fromAuthenticationProvider` method shall return a `Client` object configured with a new instance of a transport created using the `transportCtor` argument.]*/
-    return new Client(new transportCtor(authenticationProvider), null, new BlobUploadClient(authenticationProvider));
-  }
-}
-
-export namespace Client {
```
pierreca: >Client [](start = 17, length = 6)

it looks like we're losing the content of this namespace which was exported as part of the public API of th package (i believe). At least from a typescript perspective this would be considered a breaking change (customers using the content of this namespace would lose some type definitions). I think this needs to be brought back somewhere.

PR #291, opened 2018-05-16
fix(service): URI-encode device id for method invocation (fix #289)

### LINKED ISSUE #289: Invoke Direct Method will fail when device name contains’#’
# Context
- **OS and version used:** Windows 10
- **Node.js version:** v10.1.0
- **npm version:** 6.0.1
- **SDK version:** azure-iothub@1.3.0

# Code sample exhibiting the issue:
Update `<connectionString>` with iot hub connection string, update `var deviceId = 'a#';` with device name contains’#’
```javascript
'use strict';

var Client = require('azure-iothub').Client;
var connectionString = '<connectionString>';

var methodName = 'writeLine';
var deviceId = 'a#';
var client = Client.fromConnectionString(connectionString);

var methodParams = {
    methodName: methodName,
    payload: 'a line to be written',
    timeoutInSeconds: 30
};

client.invokeDeviceMethod(deviceId, methodParams, function (err, result) {
    if (err) {
        console.error('Failed to invoke method \'' + methodName + '\': ' + err.message);
    } else {
        console.log(methodName + ' on ' + deviceId + ':');
        console.log(JSON.stringify(result, null, 2));
    }
});
```

# Console log of the issue:
```
D:\Try\Test>node D:\Try\Test\js\test3.js
Failed to invoke method 'writeLine': Error: <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN""http://www.w3.org/TR/html4/strict.dtd">
<HTML><HEAD><TITLE>Bad Request</TITLE>
<META HTTP-EQUIV="Content-Type" Content="text/html; charset=us-ascii"></HEAD>
<BODY><h2>Bad Request - Invalid URL</h2>
<hr><p>HTTP Error 400. The request URL is invalid.</p>
</BODY></HTML>
```

pierreca:
#289 

# Description of the problem
Device Ids weren't URI encoded when invoking device methods

# Description of the solution
URI-encode device ids :)

### DIFF
--- a/service/src/device_method.ts
+++ b/service/src/device_method.ts
@@ -63,7 +63,8 @@ export class DeviceMethod {
     /*Codes_SRS_NODE_IOTHUB_DEVICE_METHOD_16_008: [The `invokeOn` method shall throw a `ReferenceError` if `deviceId` is `null`, `undefined` or an empty string.]*/
     if (deviceId === null || deviceId === undefined || deviceId === '') throw new ReferenceError('deviceId cannot be \'' + deviceId + '\'');
 
-    const path = '/twins/' + deviceId + '/methods' + endpoint.versionQueryString();
+    /*Codes_SRS_NODE_IOTHUB_DEVICE_METHOD_16_017: [The `invokeOn` method shall uri-encode the device id.]*/
+    const path = '/twins/' + encodeURIComponent(deviceId) + '/methods' + endpoint.versionQueryString();
     const headers = {
       'Content-Type': 'application/json; charset=utf-8'
     };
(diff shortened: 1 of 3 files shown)


service/src/device_method.ts
```diff
@@ -63,7 +63,8 @@ export class DeviceMethod {
     /*Codes_SRS_NODE_IOTHUB_DEVICE_METHOD_16_008: [The `invokeOn` method shall throw a `ReferenceError` if `deviceId` is `null`, `undefined` or an empty string.]*/
     if (deviceId === null || deviceId === undefined || deviceId === '') throw new ReferenceError('deviceId cannot be \'' + deviceId + '\'');
 
-    const path = '/twins/' + deviceId + '/methods' + endpoint.versionQueryString();
+    /*Codes_SRS_NODE_IOTHUB_DEVICE_METHOD_16_017: [The `invokeOn` method shall uri-encode the device id.]*/
+    const path = '/twins/' + encodeURIComponent(deviceId) + '/methods' + endpoint.versionQueryString();
```
anthonyvercolano: > encodeURIComponent(deviceId) [](start = 28, length = 29)

Certainly the correct fix.

We have general problems with not uri encoding the device id.  It is certainly the reason that the previous bug is there.

I'm currently looking at the http sources to make sure those are correct and will look at the transports shortly.

PR #292, opened 2018-05-16
(fix) client Fix path calculation to encode device id (only)

anthonyvercolano:
# Description of the problem

Device Id's with special characters weren't getting through the paths correctly.
# Description of the solution

Encode the device id in the paths.

### DIFF
--- a/device/transport/http/src/http.ts
+++ b/device/transport/http/src/http.ts
@@ -9,7 +9,7 @@ const debug = dbg('azure-iot-device-http:Http');
 
 import { EventEmitter } from 'events';
 import { Http as Base } from 'azure-iot-http-base';
-import { endpoint, errors, results, Message, AuthenticationProvider, AuthenticationType, TransportConfig } from 'azure-iot-common';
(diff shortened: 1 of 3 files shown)


device/transport/http/src/http.ts
```diff
@@ -153,7 +153,7 @@ export class Http extends EventEmitter implements Client.Transport {
           /*Codes_SRS_NODE_DEVICE_HTTP_16_033: [if the `getDeviceCredentials` fails with an error, the Http request shall call its callback with that error]*/
           done(err);
         } else {
-          const path = endpoint.eventPath(config.deviceId);
+          const path = endpoint.eventPath(encodeUriComponentStrict(config.deviceId));
```
BertKleewein: >eventPath(encodeUriComponentStrict(config.deviceId)); [](start = 32, length = 53)

I had a similar change in the modules-preview branch.  I feel vindicated.

BertKleewein: Should we be using this strict version everywhere?  I didn't even know we had this until now and was using encodeUriComponent because I see it everywhere 

---
In reply to: [189098076](https://github.com/Azure/azure-iot-sdk-node/pull/292#discussion_r189098076) [](ancestors = 189098076)

PR #294, opened 2018-05-21
fix (client) - change method link to use receiver-settle-mode of first

anthonyvercolano:
# Description of the problem

Link credits for methods received on the client were being exhausted when sending a large number of requests.

# Description of the solution

Switch the receiver-settle-mode to first.  This really is a better mode than second.


device/transport/amqp/src/amqp_device_method_client.ts
```diff
@@ -88,7 +88,8 @@ export class AmqpDeviceMethodClient extends EventEmitter {
                     properties: {
                       'com.microsoft:api-version': endpoint.apiVersion,
                       'com.microsoft:channel-correlation-id': credentials.deviceId
-                    }
+                    },
+                    rcvSettleMode: 0
```
BertKleewein: >0 [](start = 35, length = 1)

grumble.  I would prefer a named value here, but....whatever.  maybe add a comment saying that it's settleFirst or whatever you want to call it.

PR #301, opened 2018-05-29
Add unix domain socket support to HTTP client

avranju:
# Checklist
- [X] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).
  - [X] I edited the corresponding document in the `devdoc` folder and added or modified requirements.

# Description of the problem

The HTTP wrapper in `common/transport/http` does not support Unix Domain Sockets right now. This is needed in order to be able to communicate with iotedged.

# Description of the solution

This PR adds support Unix Domain Sockets by leveraging the `socketPath` property that the underlying Node HTTP/HTTPS module supports.


common/transport/http/src/rest_api_client.ts
```diff
@@ -273,7 +274,8 @@ export class RestApiClient {
 
 export namespace RestApiClient {
     export interface TransportConfig {
-        host: string;
+        host?: string;
+        socketPath?: string;
```
pierreca: isn't that supposed to be `host?: string | { socketPath: string }` and no `socketPath` property?

avranju: That makes sense.

PR #303, opened 2018-05-31
update configuration API to match 2018-06-30 (GA) API

BertKleewein:
I won't be able to commit this until the 2018-06-30 API makes it into production, but it's ready to go otherwise

### DIFF
--- a/e2etests/test/testUtils.js
+++ b/e2etests/test/testUtils.js
@@ -82,8 +82,24 @@ function closeDeviceEventHubClients(deviceClient, eventHubClient, ehReceivers, d
   }
 }
 
+function getErrorDetailString(functionName, err) {
+  var detail = functionName + ' returned ' + (err ? err : 'success');
+  if (err && err.responseBody) {
+    if (err.response && err.response.headers) {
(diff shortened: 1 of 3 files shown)


e2etests/test/testUtils.js
```diff
@@ -82,8 +82,24 @@ function closeDeviceEventHubClients(deviceClient, eventHubClient, ehReceivers, d
...
+function getErrorDetailString(functionName, err) {
+  var detail = functionName + ' returned ' + (err ? err : 'success');
+  if (err && err.responseBody) {
+    if (err.response && err.response.headers) {
+      if (err.response.headers['content-type'].indexOf('json') !== -1) {
+        detail += '\n';
+        var body = JSON.parse(err.responseBody);
+        delete body.StackTrace;
+        detail += JSON.stringify(body, null, ' ');
+      }
```
pierreca: you could also enclose the JSON.parse in a try catch statement that catches SyntaxError and not worry about parsing headers

pierreca:
This is missing changes to the samples in `service/samples`.
Also, without the change to the API version, it could be checked in immediately.

PR #304, opened 2018-06-01
Add signature provider

avranju:
Factor out computation of SAS tokens into a separate `SignatureProvider` interface and implementation. This allows for additional flexibility when its time to add support for other kinds of signature providers -- specifically, the one provided by Azure IoT Edge where the signature computation is backed by `iotedged`.

### DIFF
--- a/device/core/devdoc/sak_signature_provider_requirement.md
+++ b/device/core/devdoc/sak_signature_provider_requirement.md
@@ -1,0 +1,25 @@
+# SharedAccessKeySignatureProvider Requirements
+
+# Overview
+
+The `SharedAccessKeySignatureProvider` class implements the `SignatureProvider` interface by using a shared access key based HMAC algorithm. This is used to provide credentials to transports when the user wants the device to authenticate using security tokens.
+
+# Public API
+
+# constructor(private _sharedAccessKey: string, tokenValidTimeInSeconds?: number)
+
+**SRS_NODE_SAK_SIG_PROVIDER_13_003: [** The `constructor` shall throw an `ArgumentError` if the `_sharedAccessKey` parameter is falsy. **]**
+
+**SRS_NODE_SAK_SIG_PROVIDER_13_004: [** The `constructor` shall save the `tokenValidTimeInSeconds` parameter if supplied. If not, it shall default to 3600 seconds (1 hour). **]**
+
+# sign(keyName: string, data: string, callback: (err: Error, signature: SharedAccessSignature) => void): void
+
+**SRS_NODE_SAK_SIG_PROVIDER_13_005: [** The `sign` method shall throw a `ReferenceError` if the `callback` parameter is falsy or is not a function. **]**
+
+**SRS_NODE_SAK_SIG_PROVIDER_13_006: [** The `sign` method invoke `callback` with a `ReferenceError` if the `data` parameter is falsy. **]**
+
+**SRS_NODE_SAK_SIG_PROVIDER_13_001: [** Every token shall be created with a validity period of `tokenValidTimeInSeconds` if specified when the constructor was called, or 1 hour by default. **]**
+
+**SRS_NODE_SAK_SIG_PROVIDER_13_002: [** Every token shall be created using the azure-iot-common.SharedAccessSignature.create method and then serialized as a string. The the expiration time of the token will be now + the token validity time, formatted as the number of seconds since Epoch (Jan 1st, 1970, 00:00 UTC). **]**
+
+**SRS_NODE_SAK_SIG_PROVIDER_13_007: [** `sign` shall invoke the callback with the result of calling `azure-iot-common.SharedAccessSignature.create`. **]**
--- a/device/core/src/sak_authentication_provider.ts
+++ b/device/core/src/sak_authentication_provider.ts
@@ -2,7 +2,11 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 import { EventEmitter } from 'events';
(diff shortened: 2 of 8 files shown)


device/core/devdoc/sak_signature_provider_requirement.md
```diff
@@ -0,0 +1,25 @@
...
+
+# sign(keyName: string, data: string, callback: (err: Error, signature: SharedAccessSignature) => void): void
+
+**SRS_NODE_SAK_SIG_PROVIDER_13_005: [** The `sign` method shall throw a `ReferenceError` if the `callback` parameter is falsy or is not a function. **]**
+
+**SRS_NODE_SAK_SIG_PROVIDER_13_006: [** The `sign` method invoke `callback` with a `ReferenceError` if the `data` parameter is falsy. **]**
+
+**SRS_NODE_SAK_SIG_PROVIDER_13_001: [** Every token shall be created with a validity period of `tokenValidTimeInSeconds` if specified when the constructor was called, or 1 hour by default. **]**
+
+**SRS_NODE_SAK_SIG_PROVIDER_13_002: [** Every token shall be created using the azure-iot-common.SharedAccessSignature.create method and then serialized as a string. The the expiration time of the token will be now + the token validity time, formatted as the number of seconds since Epoch (Jan 1st, 1970, 00:00 UTC). **]**
```
damonbarry: Nit: "the the"

avranju: Took a while for me to see it even _after_ you pointed it out! Tricksy eyeses!

device/core/src/sak_signature_provider.ts
```diff
@@ -0,0 +1,60 @@
...
+  errors
+} from 'azure-iot-common';
+
+export class SharedAccessKeySignatureProvider implements SignatureProvider {
+  private _tokenValidTimeInSeconds: number = 3600; // 1 hour
+
+  constructor(
+    private _sharedAccessKey: string,
+    tokenValidTimeInSeconds?: number
+  ) {
```
pierreca: nit: all our signatures are on the same single line in the rest of the code

device/core/src/sak_signature_provider.ts
```diff
@@ -0,0 +1,60 @@
...
+    // Codes_SRS_NODE_SAK_SIG_PROVIDER_13_003: [ The constructor shall throw an ArgumentError if the _sharedAccessKey parameter is falsy. ]
+    if (!this._sharedAccessKey) {
+      throw new errors.ArgumentError(
+        '_sharedAccessKey cannot be "' + this._sharedAccessKey + '".'
+      );
```
pierreca: and in the function calls too. i guess we're fine with long lines

device/core/src/sak_signature_provider.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    // Codes_SRS_NODE_SAK_SIG_PROVIDER_13_004: [ The constructor shall save the tokenValidTimeInSeconds parameter if supplied. If not, it shall default to 3600 seconds (1 hour). ]
+    if (tokenValidTimeInSeconds)
+      this._tokenValidTimeInSeconds = tokenValidTimeInSeconds;
```
pierreca: surprised the linter is not barking at you for skipping curly braces here.

device/core/src/sak_authentication_provider.ts
```diff
@@ -54,46 +62,50 @@ export class SharedAccessKeyAuthenticationProvider extends EventEmitter implemen
...
-    const sas = SharedAccessSignature.create(resourceUri, this._credentials.sharedAccessKeyName, this._credentials.sharedAccessKey, newExpiry);
-    this._credentials.sharedAccessSignature = sas.toString();
 
-    const nextRenewalTimeout = (this._tokenValidTimeInSeconds - this._tokenRenewalMarginInSeconds) * 1000;
-    this._renewalTimeout = setTimeout(() => this._renewToken(), nextRenewalTimeout);
-    /*Codes_SRS_NODE_SAK_AUTH_PROVIDER_16_005: [Every time a new token is created, the `newTokenAvailable` event shall be fired with the updated credentials.]*/
-    this.emit('newTokenAvailable', this._credentials);
+    this._signatureProvider.sign(this._credentials.sharedAccessKeyName, resourceUri, (err, sas) => {
+      if (err) {
+        this.emit('error', err);
```
pierreca: >this.emit('error', err); [](start = 8, length = 24)

i'd probably say to emit the error only if there is no callback?

pierreca:
I'm wondering if you saw the overload of the SharedAccessSignature.create() method that allows to passing a function that handles the signing (which is what we use for the TPM security provider). it feels like you want to enable something similar. Maybe we should catch up in person and see if there is a good way to come up with one pattern that works in all cases?

device/core/src/sak_signature_provider.ts
avranju: I used prettier to format all the new files I added. I can put it back into one line.

device/core/src/sak_signature_provider.ts
avranju: Ah. I'll add the braces. Yeah, linter didn't say anything.

device/core/src/sak_authentication_provider.ts
avranju: Sure.

device/core/src/sak_signature_provider.ts
pierreca: are there rules to configure features of prettier?

avranju: Yes, there's a `printWidth` setting that can be used to configure how long we want lines to be before it starts breaking them up.

avranju:
The changes to re-implement this as an authentication provider are non-trivial. I'll cancel this PR and create a new one.

PR #306, opened 2018-06-06
Add Iot Edge authentication provider

avranju:
This commit adds a new authentication provider that produces SAS tokens by delegating the call to the HTTP service implemented by Azure IoT Edge.

### DIFF
--- a/device/core/src/sak_authentication_provider.ts
+++ b/device/core/src/sak_authentication_provider.ts
@@ -14,7 +14,7 @@ import { AuthenticationProvider, AuthenticationType, ConnectionString, SharedAcc
 export class SharedAccessKeyAuthenticationProvider extends EventEmitter implements AuthenticationProvider {
   type: AuthenticationType = AuthenticationType.Token;
 
-  private _tokenValidTimeInSeconds: number = 3600;   // 1 hour
+  protected _tokenValidTimeInSeconds: number = 3600;   // 1 hour
   private _tokenRenewalMarginInSeconds: number = 900; // 15 minutes
 
   private _currentTokenExpiryTimeInSeconds: number;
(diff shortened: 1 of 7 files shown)


device/core/src/iotedged_authentication_provider.ts
```diff
@@ -0,0 +1,131 @@
...
+import { RestApiClient } from 'azure-iot-http-base';
+
+// tslint:disable-next-line:no-var-requires
+const packageJson = require('../package.json');
+
+export const WORKLOAD_API_VERSION = '2018-06-28';
+const DEFAULT_SIGN_ALGORITHM = 'HMACSHA256';
+const DEFAULT_KEY_ID = 'primary';
+
+export interface EdgedAuthConfig {
```
pierreca: >export [](start = 0, length = 6)

for all exported members there needs to be a /** @private */ jsdoc comment to avoid surfacing them in the auto-generated docs on docs.microsoft.com

device/core/src/iotedged_authentication_provider.ts
```diff
@@ -0,0 +1,131 @@
...
+  keyId: string;
+  algo: string;
+  data: string;
+}
+
+interface SignResponse {
+  digest: string;
+}
+
+export class IotedgedAuthenticationProvider extends SharedAccessKeyAuthenticationProvider implements AuthenticationProvider {
```
pierreca: >IotedgedAuthenticationProvider [](start = 13, length = 30)

nit: is that the same name as other SDKs? casing looks a bit weird (although i understand it and am not gonna fight it - just looking for consistency)

device/core/src/sak_authentication_provider.ts
```diff
@@ -45,7 +47,14 @@ export class SharedAccessKeyAuthenticationProvider extends EventEmitter implemen
...
     /*Codes_SRS_NODE_SAK_AUTH_PROVIDER_16_002: [The `constructor` shall start a timer that will automatically renew the token every (`tokenValidTimeInSeconds` - `tokenRenewalMarginInSeconds`) seconds if specified, or 45 minutes by default.]*/
-    this._renewToken();
+
+    if (startTimerOnNextTick) {
+      // we kick off the initial token creation during the next tick because if this code is being run from a sub-class then
+      // there might be additional initialization that needs to occur before we are ready to sign tokens
+      setImmediate(() => this._renewToken());
+    } else {
+      this._renewToken();
+    }
```
pierreca: that feels slightly sketchy. can you not document this parameter so that we revisit this later without it being a breaking change?

pierreca:
feels like the code coverage for this package is going to change - the goal is for it not to go down. you can compute it by running "npm -s run ci" or "npm -s run alltest". then please update the numbers in the package.json script section.

device/core/src/iotedged_authentication_provider.ts
avranju: Sure. Will add that.

device/core/src/iotedged_authentication_provider.ts
avranju: C# SDK has a "signature provider" and the implementation there calls it `HttpHsmSignatureProvider`. I think this is really just Azure IoT Edge specific. "HttpHsm" sounds too broad. I can rename the class to `IotEdgedAuthenticationProvider` if that's better. The reason I used the spelling as it is right now is because the service that implements the signing is called `iotedged`. I'll rename it to `IotEdgedAuthenticationProvider` though.

avranju:
Hmm. Good catch. I thought I had all the changes covered. Coverage seems to have gone down. I'll investigate.

device/core/src/sak_authentication_provider.ts
avranju: Yes, I am not super happy about this. I'll remove the documentation. Also, if you have another idea I am happy to change this.

device/core/src/iotedged_authentication_provider.ts
avranju: Checked with John. Apparently C calls it "edge" auth module or something along those lines. I'll align with that and call it `IotEdgeAuthenticationProvider`.

avranju:
Fixed the coverage by adding more tests to `_sak_authentication_provider_test.js`.

PR #309, opened 2018-06-08
Update registry_addUpdateRemoveDevicesSample.js

abramjstamper:
Status & authentication is now deprecated in the device description according to https://docs.microsoft.com/en-us/javascript/api/azure-iothub/device?view=azure-node-latest && https://docs.microsoft.com/en-us/javascript/api/azure-iothub/registry.devicedescription?view=azure-node-latest. However, from my testing, authentication is still accepted and commits the information passed by the client, but status is does not commit a change to the IoT Hub. The Device class, upon which this file is based, appears to be deprecated.

# Description of the problem

Sample code does not match the API in the documentation on Microsoft's website.

# Description of the solution

Modified the sample code.

### DIFF
--- a/service/samples/registry_addUpdateRemoveDevicesSample.js
+++ b/service/samples/registry_addUpdateRemoveDevicesSample.js
@@ -14,7 +14,6 @@ var registry = iothub.Registry.fromConnectionString(connectionString);
 var deviceAddArray = [
   {
     deviceId: 'Device1',
-    status: 'disabled',
     authentication: {
       symmetricKey: {
         primaryKey: new Buffer(uuid.v4()).toString('base64'),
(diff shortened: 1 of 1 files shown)


pierreca:
@abramjstamper thanks for filing a PR about this - I think the status field should still work, let me check and get back to this PR asap.

pierreca:
@abramjstamper 
looked into this - `status` and `authentication` are both updated correctly using these APIs, and those properties aren't deprecated, therefore I'll close the PR and leave the files "as is".

With regards to the deprecation of the `Device` class, it's just that in general javascript developers prefer to manipulate simple objects and JSON rather than a weird class with getters and setters, especially one with a typo in a property: that's why we're deprecating it *in the SDK*. it is absolutely still valid JSON for the service and those properties aren't going away any time soon.

Whether you use the `Device` class or not, you should be able to use the sample as it is. if it does not run successfully please open an issue rather than a PR and we'll try and help to find out what's not working.

*thank you for your contribution to our open-sourced project! Please help us improve by filling out [this 2-minute customer satisfaction survey](https://www.surveymonkey.com/r/Q5PLZ3V).*

PR #310, opened 2018-06-08
Add fromEnvironment support for ModuleClient

avranju:


device/core/src/module_client.ts
```diff
@@ -222,4 +222,25 @@ export class ModuleClient extends EventEmitter {
   static fromAuthenticationProvider(authenticationProvider: AuthenticationProvider, transportCtor: any): ModuleClient {
     return InternalClient.fromAuthenticationProvider(authenticationProvider, transportCtor, ModuleClient) as ModuleClient;
   }
+
+  /**
+   * @description         Creates an IoT Hub device client by using configuration
```
pierreca: > IoT Hub device [](start = 36, length = 15)

Edge module client?

pierreca: or just module client?

---
In reply to: [194502582](https://github.com/Azure/azure-iot-sdk-node/pull/310#discussion_r194502582) [](ancestors = 194502582)

device/core/src/module_client.ts
```diff
@@ -222,4 +222,25 @@ export class ModuleClient extends EventEmitter {
...
+   *                      information from the environment. If an environment
+   *                      variable called `EdgeHubConnectionString` or `IotHubConnectionString`
```
pierreca: >EdgeHubConnectionString [](start = 43, length = 23)

Why would these env variables not follow the same naming conventions as the other? IOTEDGE_CONNECTION_STRING and IOTHUB_CONNECTION_STRING?

device/core/src/internal_client.ts
```diff
@@ -642,6 +643,105 @@ export class InternalClient extends EventEmitter {
...
+    const authenticationProvider = new IotEdgeAuthenticationProvider(authConfig);
+
+    // Codes_SRS_NODE_INTERNAL_CLIENT_13_031: [ The fromEnvironment method shall return a new instance of the Client object. ]
+    return new clientCtor(
+      new transportCtor(authenticationProvider),
+      null,
+      new BlobUploadClient(authenticationProvider)
+    );
+  }
 }
```
pierreca: I think that all of this belongs only in the module_client file, and that the ModuleClient class could call only internalClient.fromConnectionString or internalClient.fromAuthenticationProvider?

avranju: Sure. Will move.

device/core/src/module_client.ts
avranju: Fixed.

device/core/src/module_client.ts
avranju: Not sure. The other SDKs seem to follow this convention. Agree that it should be consistent though. We shipped public preview of Edge with these variable names and it has continued from there.

device/core/src/module_client.ts
```diff
@@ -222,4 +223,90 @@ export class ModuleClient extends EventEmitter {
...
+      'IOTEDGE_MODULEID',
+      'IOTEDGE_IOTHUBHOSTNAME',
+      'IOTEDGE_AUTHSCHEME',
+      'IOTEDGE_MODULEGENERATIONID'
+    ];
+
+    keys.forEach((key) => {
+      if (!process.env[key]) {
+        throw new errors.PreconditionFailedError(
+          `Envrionment variable ${key} was not provided.`
```
pierreca: >Envrionment [](start = 11, length = 11)

typo

device/core/src/module_client.ts
```diff
@@ -222,4 +223,90 @@ export class ModuleClient extends EventEmitter {
...
+      'IOTEDGE_DEVICEID',
+      'IOTEDGE_MODULEID',
+      'IOTEDGE_IOTHUBHOSTNAME',
+      'IOTEDGE_AUTHSCHEME',
+      'IOTEDGE_MODULEGENERATIONID'
+    ];
+
+    keys.forEach((key) => {
+      if (!process.env[key]) {
+        throw new errors.PreconditionFailedError(
```
pierreca: >PreconditionFailedError [](start = 25, length = 23)

that's an error that is usually used for service related errors (especially invalid etags). I'd say this ought to be either a ReferenceError (because the key isn't found) or an ArgumentError (because that's what we throw when the connection string is invalid)

device/core/src/module_client.ts
```diff
@@ -222,4 +223,90 @@ export class ModuleClient extends EventEmitter {
...
+    // Codes_SRS_NODE_MODULE_CLIENT_13_030: [ The value for the environment variable IOTEDGE_AUTHSCHEME must be SasToken. ]
+
+    // we only support sas token auth scheme at this time
+    if (process.env.IOTEDGE_AUTHSCHEME !== 'SasToken') {
+      throw new errors.PreconditionFailedError(
+        `Authentication scheme ${
+          process.env.IOTEDGE_AUTHSCHEME
+        } is not a supported scheme.`
+      );
+    }
```
pierreca: same thing about the error: also, I don't see why we want the user to add this env variable if there's only one possible value. I'd advocate to set SasToken to be the default, and ask for the env variable only when we actually have the choice of multiple values. unnecessary burden otherwise.

device/core/src/module_client.ts
pierreca: ok i guess...

---
In reply to: [194575469](https://github.com/Azure/azure-iot-sdk-node/pull/310#discussion_r194575469) [](ancestors = 194575469)

device/core/test/_module_client_test.js
```diff
@@ -0,0 +1,155 @@
...
+
+      // Tests_SRS_NODE_MODULE_CLIENT_13_032: [ The fromEnvironment method shall create a new IotEdgeAuthenticationProvider object and pass this to the transport constructor. ]
+      // Tests_SRS_NODE_MODULE_CLIENT_13_031: [ The fromEnvironment method shall return a new instance of the ModuleClient object. ]
+      it('creates IotEdgeAuthenticationProvider', function() {
+        var transportStub = sinon.stub().returns(new EventEmitter());
+        var client = ModuleClient.fromEnvironment(transportStub);
+        assert.equal(transportStub.called, true);
+        var provider = transportStub.args[0][0];
+        assert.isOk(provider);
+        assert.equal(provider._authConfig.workloadUri, 'unix:///var/run/iotedge.w.sock');
```
pierreca: >equal [](start = 15, length = 5)

nit: strictEqual?

pierreca: >_authConfig [](start = 30, length = 11)

guessing we can't stub the IoTEdgeAuthenticationProvider? i'm not a fan of reaching within private members of another class in unit tests..

device/core/src/module_client.ts
avranju: I want to leave this in since we'll also support x509 auth scheme shortly after GA. I'd rather have this check be in here rather than forgetting to validate this later when the new scheme is added.

device/core/test/_module_client_test.js
avranju: Not sure how to do that here.

PR #311, opened 2018-06-11
Update ADM configuration objects

pierreca:
# Description of the problem
ADM/Configuration APIs are being updated for the GA API version

# Description of the solution
Update the objects.

Please note:
- still missing an e2e test
- CRUD e2e tests for configurations are still disabled
- the new API version change itself will happen only when it is deployed


pierreca:
just realized this is a duplicated of #303. closing

PR #316, opened 2018-06-14
Use inheritance instead of composition for clients

pierreca:
# Description of the problem
Using composition forces all code to go into the InternalClient class (so as to use retries, etc)

# Description of the solution
Change to inheritance, have only the device/module-specific code in the `Client` and `ModuleClient` classes and inherit from `InternalClient`


device/core/devdoc/device_client_requirements.md
```diff
@@ -0,0 +1,75 @@
+# Client requirements (Device client)
```
BertKleewein: >Client [](start = 2, length = 6)

Damn.  Did I really miss this file?  Aargh

device/core/src/module_client.ts
```diff
@@ -182,26 +114,75 @@ export class ModuleClient extends EventEmitter {
...
-   *                    connection string using the given transport type.
-   *
-   * @param {String}    connStr       A connection string which encapsulates "device
-   *                                  connect" permissions on an IoT hub.
-   * @param {Function}  Transport     A transport constructor.
-   *
-   * @throws {ReferenceError}         If the connStr parameter is falsy.
-   *
-   */
   static fromConnectionString(connStr: string, transportCtor: any): ModuleClient {
```
BertKleewein: >fromConnectionString [](start = 9, length = 20)

I don't necessarily like the way that Varun did his factory methods, but it feels like there should be a way to not copy/paste this code in both clients.  I don't know the right answer, but it's something to think about.

device/core/test/_client_retry_test.js
```diff
@@ -0,0 +1,109 @@
...
+    client.on('message', function() {});
+  });
+});
+
+[DeviceClient, ModuleClient].forEach(function (ClientCtor) {
+  describe(ClientCtor.name + ' Retry Logic', function () {
+    [
+      {
+        funcName: 'sendEvent',
+        funcParam: new Message('foo')
```
BertKleewein: >funcParam [](start = 8, length = 9)

It doesn't look like you're using funcParam anywhere.

device/core/test/_client_retry_test.js
```diff
@@ -0,0 +1,109 @@
...
+      },
+      {
+        funcName: 'abandon',
+        funcParam: new Message('foo')
+      },
+      {
+        funcName: 'setOptions',
+        funcParam: {}
+      }
+    ].forEach(function (testConfig) {
```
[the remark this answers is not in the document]
BertKleewein: Also, what about enabling twin and inputMessages?  Seems like that needs a retry test.

---
In reply to: [195806686](https://github.com/Azure/azure-iot-sdk-node/pull/316#discussion_r195806686) [](ancestors = 195806686)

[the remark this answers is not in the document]
BertKleewein: I see that you just moved existing tests.  I don't expect you to fix this.  Maybe we should add a techdebt task?

---
In reply to: [195807205](https://github.com/Azure/azure-iot-sdk-node/pull/316#discussion_r195807205) [](ancestors = 195807205,195806686)

device/core/test/_module_client_test.js
```diff
@@ -5,12 +5,43 @@
...
+  describe('#fromSharedAccessSignature', function () {
+    /*Tests_SRS_NODE_INTERNAL_CLIENT_16_030: [The fromSharedAccessSignature method shall return a new instance of the InternalClient object] */
+    it('returns an instance of ModuleClient', function () {
+      var client = ModuleClient.fromSharedAccessSignature(sharedAccessSignature, FakeTransport);
+      assert.instanceOf(client, ModuleClient);
+    });
+  });
+
+  describe('#fromAuthenticationProvider', function () {
+    /*Tests_SRS_NODE_INTERNAL_CLIENT_16_091: [The `fromAuthenticationProvider` method shall return a `InternalClient` object configured with a new instance of a transport created using the `transportCtor` argument.]*/
```
BertKleewein: >transport created using the `transportCtor` argument [](start = 161, length = 52)

For completeness, do you want to assert that it has a transport created using the transportCtor?

device/core/test/_module_client_test.js
```diff
@@ -5,12 +5,43 @@
...
 var ModuleClient = require('../lib/module_client').ModuleClient;
 var IotEdgeAuthenticationProvider = require('../lib/iotedge_authentication_provider').IotEdgeAuthenticationProvider;
 
-describe('ModuleClient', function() {
+describe('ModuleClient', function () {
+  var sharedKeyConnectionString = 'HostName=host;DeviceId=id;ModuleId=modId;SharedAccessKey=key';
+  var sharedAccessSignature = '"SharedAccessSignature sr=hubName.azure-devices.net/devices/deviceId/modules/moduleId&sig=s1gn4tur3&se=1454204843"';
+
+  describe('#fromConnectionString', function () {
+    /*Tests_SRS_NODE_INTERNAL_CLIENT_05_006: [The fromConnectionString method shall return a new instance of the InternalClient object, as by a call to new InternalClient(new Transport(...)).]*/
```
BertKleewein: >Tests_SRS_NODE_INTERNAL_CLIENT_05_006 [](start = 6, length = 37)

I hate renaming SRS names, but do you want to rename the SRS name?  Also, this refers to the InternalClient class.

BertKleewein: All comments in here that apply to one factory method apply to all factory methods. I'm not going to comment each time.

---
In reply to: [195808304](https://github.com/Azure/azure-iot-sdk-node/pull/316#discussion_r195808304) [](ancestors = 195808304)

device/core/test/_module_client_test.js
```diff
@@ -152,4 +183,117 @@ describe('ModuleClient', function() {
...
+          });
+        });
+      });
+    });
+  });
+
+  describe('#on(\'inputMessage\')', function () {
+    /*Tests_SRS_NODE_INTERNAL_CLIENT_18_012: [ The `inputMessage` event shall be emitted when an inputMessage is received from the IoT Hub service. ]*/
+    it('emits a message event when a message is received', function (done) {
+      var dummyTransport = new FakeTransport();
```
BertKleewein: >dummyTransport [](start = 10, length = 14)

The PC police say that this is an offensive name.

BertKleewein: Wait, that's my dev ID.  Was I the person who was being insensitive?

---
In reply to: [195808859](https://github.com/Azure/azure-iot-sdk-node/pull/316#discussion_r195808859) [](ancestors = 195808859)

device/core/test/_module_client_test.js
```diff
@@ -152,4 +183,117 @@ describe('ModuleClient', function() {
...
+        var fakeError = new Error('fake');
+        sinon.stub(dummyTransport, testConfig.enableFunc).callsFake(function (callback) { callback(fakeError); });
+        var client = new ModuleClient(dummyTransport);
+        client.on('error', function (err) {
+          assert.strictEqual(err, fakeError);
+          testCallback();
+        })
+
+        // Calling 'on' twice to make sure it's called only once on the receiver.
+        // It works because the test will fail if the test callback is called multiple times, and it's called for every time the testConfig.eventName event is subscribed on the receiver.
```
BertKleewein: remove these 2 comments please.  It looks like copypasta from the 18_014 test above.

device/core/test/_module_client_test.js
```diff
@@ -152,4 +183,117 @@ describe('ModuleClient', function() {
...
+        client.on(testConfig.eventName, function () { });
+        assert.isTrue(dummyTransport[testConfig.enableFunc].calledOnce);
+      });
+
+      /*Tests_SRS_NODE_INTERNAL_CLIENT_18_017: [ The client shall emit an `error` if connecting the transport fails while subscribing to `inputMessage` events. ]*/
+      it('emits an error if it fails to stop listening for messages', function (testCallback) {
+        var dummyTransport = new FakeTransport();
+        var fakeError = new Error('fake');
+        sinon.spy(dummyTransport, testConfig.enableFunc);
+        sinon.stub(dummyTransport, testConfig.disableFunc).callsFake(function (callback) { callback(fakeError); });
```
BertKleewein: >sinon.stub(dummyTransport, testConfig.disableFunc).callsFake(function (callback) { callback(fakeError); }); [](start = 8, length = 107)

It's embarrassing to see how many mistakes I'm finding in my own tests.  This case is actually testing the case where the transport fails while _unsubscribing_.  I don't care if you add a new SRS.  I'm just shocked at what I see looking at my own work from a different perspective.

device/core/test/_internal_client_test.js
```diff
@@ -18,441 +16,335 @@ var Message = require('azure-iot-common').Message;
...
-      var client = new InternalClient(new FakeTransport());
-      client.setOptions({}, done);
-    });
-
-    /*Tests_SRS_NODE_INTERNAL_CLIENT_16_044: [The `done` callback shall be invoked with a standard javascript `Error` object and no result object if the client could not be configured as requested.]*/
-    it('calls the done callback with an error when it failed to configured the transport', function(done) {
-      var failingTransport = new FakeTransport();
-      sinon.stub(failingTransport, 'setOptions').callsFake(function (options, done) {
-        done(new Error('dummy error'));
+    describe('#setOptions', function () {
```
BertKleewein: >describe('#setOptions', function () { [](start = 4, length = 37)

nit: adding this space seems arbitrary.  Should we do a one-time prettifier pass?

device/core/test/_internal_client_test.js
```diff
@@ -18,441 +16,335 @@ var Message = require('azure-iot-common').Message;
...
-    it('throws if transport arg is falsy', function () {
-      [null, undefined, '', 0].forEach(function (transport) {
-        assert.throws(function () {
-          return new InternalClient(transport);
-        }, ReferenceError, 'transport is \'' + transport + '\'');
+[ModuleClient, DeviceClient].forEach(function (ClientCtor) {
+  describe(ClientCtor.name, function () {
+    var sharedKeyConnectionString = 'HostName=host;DeviceId=id;SharedAccessKey=key';
+    describe('#constructor', function () {
+      /*Tests_SRS_NODE_INTERNAL_CLIENT_05_001: [The InternalClient constructor shall throw ReferenceError if the transport argument is falsy.]*/
```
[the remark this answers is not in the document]
BertKleewein: OK, the drift isn't as big as I thought.  I see why most of these are INTERNAL_CLIENT requirements being tested at the DeviceClient/ModuleClient level, but I think there's still some sloppiness here (as with any big code move, it's impossible to get them all 100% correct).  

---
In reply to: [195811526](https://github.com/Azure/azure-iot-sdk-node/pull/316#discussion_r195811526) [](ancestors = 195811526)

device/core/test/_device_client_test.js
```diff
@@ -0,0 +1,257 @@
...
+var results = require('azure-iot-common').results;
+var X509AuthenticationProvider = require('../lib/x509_authentication_provider').X509AuthenticationProvider;
+var Client = require('../lib/device_client').Client;
+
+describe('Device Client', function () {
+  var sharedKeyConnectionString = 'HostName=host;DeviceId=id;SharedAccessKey=key';
+  var sharedAccessSignature = '"SharedAccessSignature sr=hubName.azure-devices.net/devices/deviceId&sig=s1gn4tur3&se=1454204843"';
+
+  describe('#fromConnectionString', function () {
+    /*Tests_SRS_NODE_DEVICE_CLIENT_05_006: [The fromConnectionString method shall return a new instance of the InternalClient object, as by a call to new InternalClient(new Transport(...)).]*/
```
BertKleewein: >InternalClient [](start = 154, length = 14)

Similar comment as in the ModuleClient factory methods.  the requirements refer to InternalClient still.

device/core/test/_device_client_test.js
```diff
@@ -0,0 +1,257 @@
...
+  describe('#fromAuthenticationProvider', function () {
+    /*Tests_SRS_NODE_DEVICE_CLIENT_16_091: [The `fromAuthenticationProvider` method shall return a `InternalClient` object configured with a new instance of a transport created using the `transportCtor` argument.]*/
+    it('returns an instance of Client', function () {
+      var client = Client.fromAuthenticationProvider({}, FakeTransport);
+      assert.instanceOf(client, Client);
+    });
+  });
+
+
+  describe('#uploadToBlob', function() {
```
BertKleewein: >uploadToBlob [](start = 13, length = 12)

random comment -- I haven't looked at the code yet, but does the blob upload functionality also need retry tests?

device/core/test/_device_client_test.js
```diff
@@ -0,0 +1,257 @@
...
+      // Calling 'on' twice to make sure it's called only once on the receiver.
+      // It works because the test will fail if the test callback is called multiple times, and it's called for every time the 'message' event is subscribed on the receiver.
+      client.on('message', function () { });
+      client.on('message', function () { });
+      assert.isTrue(dummyTransport.enableC2D.calledOnce);
+    });
+
+    /*Tests_SRS_NODE_DEVICE_CLIENT_16_005: [The client shall stop listening for messages from the service whenever the last listener unsubscribes from the ‘message’ event.]*/
+    it('stops listening for messages when the last listener has unsubscribed', function (testCallback) {
+      var dummyTransport = new FakeTransport();
```
BertKleewein: >dummyTransport [](start = 10, length = 14)

ah, you wrote this test and I copied it.  You're the source of the non-PC variable names.  tsk. tsk. tsk.

device/core/src/device_client.ts
```diff
@@ -21,9 +25,14 @@ import { Twin  } from './twin';
...
  * to create an IoT Hub device client.
  */
-export class Client extends EventEmitter {
-  private _internalClient: InternalClient;
-
+export class Client extends InternalClient {
+  private _c2dEnabled: boolean;
+  private _deviceDisconnectHandler: (err?: Error, result?: any) => void;
+  /**
+   * @private
```
BertKleewein: >* @private [](start = 3, length = 10)

nit: I think the docs.microsoft.com jsdoc parsers are pretty good about honoring the private scoping on the variable.  From what I've seen, they're pretty good about pulling everything (like types and parameters) out of .ts function prototypes, so most of the duplication that you have to put in jsdoc for vanilla javascript isn't necessary in typescript.

device/core/src/module_client.ts
```diff
@@ -210,50 +191,39 @@ export class ModuleClient extends EventEmitter {
    *
    * @throws {ReferenceError}         If the connStr parameter is falsy.
    *
+   * @returns {module:azure-iothub.Client}
```
BertKleewein: >* @returns {module:azure-iothub.Client} [](start = 3, length = 39)

To quote DT: "wrong!".  the @method and @description tags you put in aren't necessary.  If you need proof, I think you can  look at some of the provisioning doc at docs.microsoft.com and see how much detail it pulled out of the prototype.

device/core/src/module_client.ts
pierreca: there's only a bit of code that is shared. 2 things that motivated me to not refactor:
1. client constructors may evolve separately (including with separate parameters
2. not all auth mechanisms are supported everywhere which leads to minor differences

we could refactor argument checking (meh) and maybe instantiating the auth provider.. not that big of a gain imho. definitely a thing to keep in mind in the future though

---
In reply to: [195805036](https://github.com/Azure/azure-iot-sdk-node/pull/316#discussion_r195805036) [](ancestors = 195805036)

device/core/test/_client_retry_test.js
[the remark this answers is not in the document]
pierreca: code coverage keeps track of this for us :)

---
In reply to: [195807455](https://github.com/Azure/azure-iot-sdk-node/pull/316#discussion_r195807455) [](ancestors = 195807455,195807205,195806686)

PR #317, opened 2018-06-15
Compare authentication scheme case-insensitively

LazarusX:
# Checklist

N/A

# Description of the problem

iotedged changes the authentication scheme environment variable from `SasToken` to `sasToken`, which causes the module to fail.

# Description of the solution


pierreca:
taken care of with #318 - thanks!

PR #319, opened 2018-06-15
Add question mark to MQTT username

BertKleewein:
Adding question mark to MQTT username so modules-over-mqtt works against iothub


device/transport/mqtt/src/mqtt.ts
```diff
@@ -657,7 +657,7 @@ export class Mqtt extends EventEmitter implements DeviceTransport {
     let baseConfig: MqttBaseTransportConfig = {
       uri: 'mqtts://' + (credentials.gatewayHostName || credentials.host),
       username: credentials.host + '/' + clientId +
-        '/' + endpoint.versionQueryString().substr(1) +
+        '/' + endpoint.versionQueryString() +
```
anthonyvercolano: >endpoint.versionQueryString() [](start = 14, length = 29)

I'll approve it, but I guess I'm wondering why you went through the effort of removing it originally.

PR #322, opened 2018-06-21
ModuleClient.invokeMethod (works only with Edge)

pierreca:
- Added a new object `MethodClient` in the `azure-iot-device` package
- Introduce the new public API `ModuleClient.invokeMethod`
- Use the new `MethodClient` in `ModuleClient.invokeMethod`
- Override `InternalClient.setOptions` in `ModuleClient` so that it updates the `MethodClient` too
- Add the ability for the `RestApiClient` to use a custom CA cert

@avranju FYI - you don't actually have to review if you don't have the time :)

### DIFF
--- a/device/core/devdoc/device_method/method_client_requirements.md
+++ b/device/core/devdoc/device_method/method_client_requirements.md
@@ -1,0 +1,49 @@
+# DirectMethodClient requirements
+
+The `DirectMethodClient` class is in charge of making edgeHub method API calls. It is a private class that shall be used by the `ModuleClient` to make method calls and the SDK user should only use the `ModuleClient` object.
+
+## Overview
+```typescript
+export class DirectMethodClient {
+  constructor(authProvider: AuthenticationProvider);
+  invokeMethod(deviceId: string, moduleId: string, methodParams: DirectMethodParams, callback: DirectMethodCallback): void;
+  setOptions(options: any): void;
+  updateSharedAccessSignature(sharedAccessSignature: string, callback: (err?: Error) => void): void;
+}
+```
+
+## API
+*Note: this API is internal only and is used solely by the `ModuleClient` object.*
+
+### invokeMethod(deviceId: string, moduleId: string, methodParams: DirectMethodParams, callback: DirectMethodCallback): void
+
+**SRS_NODE_DEVICE_METHOD_CLIENT_16_006: [** The `invokeMethod` method shall get the latest credentials by calling `getDeviceCredentials` on the `AuthenticationProvider` object. **]**
+
+**SRS_NODE_DEVICE_METHOD_CLIENT_16_007: [** The `invokeMethod` method shall create a `RestApiClient` object if it does not exist. **]**
+
+**SRS_NODE_DEVICE_METHOD_CLIENT_16_015: [** The `invokeMethod` method shall update the shared access signature of the `RestApiClient` by using its `updateSharedAccessSignature` method and the credentials obtained with the call to `getDeviceCredentials` (see `SRS_NODE_DEVICE_METHOD_CLIENT_16_006`). **]**
+
+**SRS_NODE_DEVICE_METHOD_CLIENT_16_008: [** The `invokeMethod` method shall call its callback with an `Error` if it fails to get the latest credentials from the `AuthenticationProvider` object. **]**
+
+**SRS_NODE_DEVICE_METHOD_CLIENT_16_009: [** The `invokeMethod` method shall call the `setOptions` method on the `RestApiClient` with its options as argument to make sure the CA certificate is populated. **]**
+
+**SRS_NODE_DEVICE_METHOD_CLIENT_16_010: [** The `invokeMethod` method shall construct the HTTP request path as `/twins/encodeUriComponentStrict(<targetDeviceId>)/methods` if the target is a device. **]**
+
+**SRS_NODE_DEVICE_METHOD_CLIENT_16_011: [** The `invokeMethod` method shall construct the HTTP request path as `/twins/encodeUriComponentStrict(<targetDeviceId>)/modules/encodeUriComponentStrict(<targetModuleId>)/methods` if the target is a module. **]**
+
+**SRS_NODE_DEVICE_METHOD_CLIENT_16_012: [** The `invokeMethod` method shall call `RestApiClient.executeApiCall` with:
+- `POST` for the HTTP method argument.
+- `path` as defined in `SRS_NODE_DEVICE_METHOD_CLIENT_16_010` and `SRS_NODE_DEVICE_METHOD_CLIENT_16_011`
+- 2 custom headers:
+  - `Content-Type` shall be set to `application/json`
+  - `x-ms-edge-moduleId` shall be set to `<deviceId>/<moduleId>` with `deviceId` and `moduleId` being the identifiers for the current module (as opposed to the target module)
+- the stringified version of the `MethodParams` object as the body of the request
+- a timeout value in milliseconds that is the sum of the `connectTimeoutInSeconds` and `responseTimeoutInSeconds` parameters of the `MethodParams` object. **]**
+
+**SRS_NODE_DEVICE_METHOD_CLIENT_16_013: [** The `invokeMethod` method shall call its callback with an error if `RestApiClient.executeApiCall` fails. **]**
+
+**SRS_NODE_DEVICE_METHOD_CLIENT_16_014: [** The `invokeMethod` method shall call its callback with the result object if the call to `RestApiClient.executeApiCall` succeeds. **]**
+
+### setOptions(options: any): void
+
+**SRS_NODE_DEVICE_METHOD_CLIENT_16_001: [** The `setOptions` method shall merge the options passed in argument with the existing set of options used by the `MethodClient`. **]**
--- a/device/core/devdoc/module_client_requirements.md
+++ b/device/core/devdoc/module_client_requirements.md
@@ -82,8 +82,32 @@ class ModuleClient extends InternalClient {
 
 **SRS_NODE_MODULE_CLIENT_18_022: [** The `sendOutputEventBatch` method shall not throw if the `callback` is not passed. **]**
 
+### setOptions
+**SRS_NODE_MODULE_CLIENT_16_042: [** The `setOptions` method shall throw a `ReferenceError` if the options object is falsy. **]**
 
-**SRS_NODE_MODULE_CLIENT_16_096: [** The `setRetryPolicy` method shall call the `setRetryPolicy` method on the twin if it is set and pass it the `policy` object. **]**
+**SRS_NODE_MODULE_CLIENT_16_043: [** The `done` callback shall be invoked with no parameters when it has successfully finished setting the client and/or transport options. **]**
+
+**SRS_NODE_MODULE_CLIENT_16_044: [** The `done` callback shall be invoked with a standard javascript `Error` object and no result object if the client could not be configured as requested. **]**
+
+**SRS_NODE_MODULE_CLIENT_16_098: [** The `setOptions` method shall call the `setOptions` method with the `options` argument on the `MethodClient` object of the `ModuleClient`. **]**
+
+### invokeMethod(deviceId: string, moduleIdOrMethodParams: string | DirectMethodParams, methodParamsOrCallback: DirectMethodParams | DirectMethodCallback, callback?: DirectMethodCallback): void
+
+**Other valid signatures:**
+```typescript
+invokeMethod(deviceId: string, methodParams: DirectMethodParams, callback: DirectMethodCallback): void;
+invokeMethod(deviceId: string, moduleId: string, methodParams: DirectMethodParams, callback: DirectMethodCallback): void;
+```
+
+**SRS_NODE_MODULE_CLIENT_16_093: [** `invokeMethod` shall throw a `ReferenceError` if the `deviceId` argument is falsy. **]**
+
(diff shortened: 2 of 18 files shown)


device/core/src/device_method/method_client.ts
```diff
@@ -0,0 +1,115 @@
...
+          callback();
+        } else {
+          getUserAgentString((userAgentString) => {
+            const transportConfig: RestApiClient.TransportConfig = {
+              host: creds.gatewayHostName,
+              sharedAccessSignature: creds.sharedAccessSignature
+            };
+
+            this._restApiClient = new RestApiClient(transportConfig, userAgentString);
+            this._restApiClient.updateSharedAccessSignature(creds.sharedAccessSignature);
```
avranju: Is this update call necessary if the token was passed in through the `transportConfig` param?

device/core/src/module_client.ts
```diff
@@ -290,12 +370,14 @@ export class ModuleClient extends InternalClient {
         callback(err);
       } else {
         const transport = new transportCtor(authenticationProvider);
-
         // Codes_SRS_NODE_MODULE_CLIENT_13_035: [ If the client is running in edge mode then the IotEdgeAuthenticationProvider.getTrustBundle method shall be invoked to retrieve the CA cert and the returned value shall be set as the CA cert for the transport via the transport's setOptions method passing in the CA value for the ca property in the options object. ]
         transport.setOptions({ ca });
 
+        const methodClient = new MethodClient(authenticationProvider);
+        methodClient.setOptions({ ca });
```
avranju: Shouldn't this CA get set for the `MethodClient` from the `_fromEnvironmentNormal` function too?

device/core/src/device_method/method_client.ts
pierreca: that's a good point...

---
In reply to: [197006491](https://github.com/Azure/azure-iot-sdk-node/pull/322#discussion_r197006491) [](ancestors = 197006491)

device/core/src/device_method/method_client.ts
```diff
@@ -0,0 +1,115 @@
...
+      }
+    });
+  }
+
+  setOptions(options: any): void {
+    /*Codes_SRS_NODE_DEVICE_METHOD_CLIENT_16_001: [The `setOptions` method shall merge the options passed in argument with the existing set of options used by the `MethodClient`.]*/
+    this._options = _.merge(this._options, options);
+  }
+
+  updateSharedAccessSignature(sharedAccessSignature: string, callback: (err?: Error) => void): void {
```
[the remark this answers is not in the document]
pierreca: You're right... I had that as I was thinking of an old historical code path - but since we have AuthenticationProviders, this problem is solved elegantly and I actually don't need it.
(user calls ModuleClient.updateSharedAccessSignature - which in turn will call the transport updateSharedAccessSignature method, which will update the AuthenticationProvider. since getDeviceCredentials is called before any API call, it'll always be up to date and I don't need this function at all. brilliant :) thanks for pointing it out!

---
In reply to: [197007121](https://github.com/Azure/azure-iot-sdk-node/pull/322#discussion_r197007121) [](ancestors = 197007121)

device/core/src/module_client.ts
pierreca: aren't methods supported only in the Edge environment?

---
In reply to: [197007661](https://github.com/Azure/azure-iot-sdk-node/pull/322#discussion_r197007661) [](ancestors = 197007661)

pierreca: my bad, my misunderstanding. I was confused and thought the _fromEnvironmentNormal method was for non-edge environment (iot hub). it actually refers to env variables as opposed to edgelet. will fix.

---
In reply to: [197012544](https://github.com/Azure/azure-iot-sdk-node/pull/322#discussion_r197012544) [](ancestors = 197012544,197007661)

device/core/devdoc/module_client_requirements.md
```diff
@@ -82,8 +82,32 @@ class ModuleClient extends InternalClient {
...
-**SRS_NODE_MODULE_CLIENT_16_096: [** The `setRetryPolicy` method shall call the `setRetryPolicy` method on the twin if it is set and pass it the `policy` object. **]**
+**SRS_NODE_MODULE_CLIENT_16_043: [** The `done` callback shall be invoked with no parameters when it has successfully finished setting the client and/or transport options. **]**
+
+**SRS_NODE_MODULE_CLIENT_16_044: [** The `done` callback shall be invoked with a standard javascript `Error` object and no result object if the client could not be configured as requested. **]**
+
+**SRS_NODE_MODULE_CLIENT_16_098: [** The `setOptions` method shall call the `setOptions` method with the `options` argument on the `MethodClient` object of the `ModuleClient`. **]**
+
+### invokeMethod(deviceId: string, moduleIdOrMethodParams: string | DirectMethodParams, methodParamsOrCallback: DirectMethodParams | DirectMethodCallback, callback?: DirectMethodCallback): void
+
+**Other valid signatures:**
```
BertKleewein: Interesting.  So we can call a device method with a module identity?  Is that strange?  Do other SDKs allow this?

device/core/devdoc/module_client_requirements.md
```diff
@@ -82,8 +82,32 @@ class ModuleClient extends InternalClient {
...
+
+**SRS_NODE_MODULE_CLIENT_16_093: [** `invokeMethod` shall throw a `ReferenceError` if the `deviceId` argument is falsy. **]**
+
+**SRS_NODE_MODULE_CLIENT_16_094: [** `invokeMethod` shall throw a `ReferenceError` if the `moduleIdOrMethodParams` argument is falsy. **]**
+
+**SRS_NODE_MODULE_CLIENT_16_095: [** `invokeMethod` shall throw a `ReferenceError` if the `deviceId` and `moduleIdOrMethodParams` are strings and the `methodParamsOrCallback` argument is falsy. **]**
+
+**SRS_NODE_MODULE_CLIENT_16_096: [** `invokeMethod` shall throw a `ArgumentError` if the `methodName` property of the `DeviceMethodParams` argument is falsy. **]**
+
+**SRS_NODE_MODULE_CLIENT_16_097: [** `invokeMethod` shall call the `invokeMethod` API of the `DeviceMethodClient` API that was created for the `ModuleClient` instance. **]**
```
BertKleewein: >DeviceMethodClient [](start = 94, length = 18)

nit: this is the only place in this doc that mentions the DeviceMethodClient object.  seems strange.

device/core/devdoc/device_method/method_client_requirements.md
```diff
@@ -0,0 +1,47 @@
...
+```
+
+## API
+*Note: this API is internal only and is used solely by the `ModuleClient` object.*
+
+### invokeMethod(deviceId: string, moduleId: string, methodParams: DirectMethodParams, callback: DirectMethodCallback): void
+
+**SRS_NODE_DEVICE_METHOD_CLIENT_16_006: [** The `invokeMethod` method shall get the latest credentials by calling `getDeviceCredentials` on the `AuthenticationProvider` object. **]**
+
+**SRS_NODE_DEVICE_METHOD_CLIENT_16_007: [** The `invokeMethod` method shall create a `RestApiClient` object if it does not exist and update its shared access signature. **]**
```
BertKleewein: >The `invokeMethod` method shall create a `RestApiClient` object if it does not exist and update its shared access signature [](start = 44, length = 123)

This is a vaguely worded requirement because it's not clear if it always updates the SAS, or only if it creates the new RestApiClient object.  At the very least, this requirement describes two flows that probably deserves 2 test cases.  I needed to read the code to makes sure it did what I thought.  I recommend having 2 requirements: 1) It creates the RestApiClient if it doesn't exist (with uses a valid SAS), or 2) if the RestApiClient already exists, it updates the SAS.

device/core/src/module_client.ts
```diff
@@ -126,11 +131,76 @@ export class ModuleClient extends InternalClient {
...
+    const actualCallback     = typeof methodParamsOrCallback === 'function' ? methodParamsOrCallback : callback;
+
+
+    /*Codes_SRS_NODE_MODULE_CLIENT_16_096: [`invokeMethod` shall throw a `ArgumentError` if the `methodName` property of the `DeviceMethodParams` argument is falsy.]*/
+    if (!(actualMethodParams as MethodParams).methodName) {
+      throw new errors.ArgumentError('the name property of the methodParams argument cannot be \'' + (actualMethodParams as MethodParams).methodName + '\'');
+    }
+
+    /*Codes_SRS_NODE_MODULE_CLIENT_16_097: [`invokeMethod` shall call the `invokeMethod` API of the `DeviceMethodClient` API that was created for the `ModuleClient` instance.]*/
+    this._methodClient.invokeMethod(deviceId, actualModuleId, actualMethodParams as MethodParams, actualCallback);
```
BertKleewein: >_methodClient [](start = 9, length = 13)

If _methodClient is null, presumably because this is a moduleClient talking to IoTHub, we should fail earlier with a more meaningful error.

device/core/src/module_client.ts
```diff
@@ -126,11 +131,76 @@ export class ModuleClient extends InternalClient {
...
+    if (!deviceId) {
+      throw new ReferenceError('deviceId cannot be \'' + deviceId + '\'');
+    }
+    /*Codes_SRS_NODE_MODULE_CLIENT_16_094: [`invokeMethod` shall throw a `ReferenceError` if the `moduleIdOrMethodParams` argument is falsy.]*/
+    if (!moduleIdOrMethodParams) {
+      if (typeof methodParamsOrCallback === 'object') {
+        throw new ReferenceError('moduleId cannot be \'' + moduleIdOrMethodParams + '\'');
+      } else {
+        throw new ReferenceError('methodParams cannot be \'' + moduleIdOrMethodParams + '\'');
+      }
```
BertKleewein: for maintainability and readability, I would collapse these into one exception "second parameter cannot be 'blah'.  Otherwise, it could be confusing to the caller because you're trying to infer the meaning of the second parameter based on the type of the third parameter.  
This particular group is functions is just hard to read and grok.  Maybe not too important since it's just parameter checking, but I would need to take out a piece of paper just to make sure it works correctly and that's a bad sign.

BertKleewein: Maybe this would be easier if you moved the overloads (with their confusing names) into actualX values and then verified those instead.  I think that would be way way easier to understand.  

---
In reply to: [197229768](https://github.com/Azure/azure-iot-sdk-node/pull/322#discussion_r197229768) [](ancestors = 197229768)

device/core/src/module_client.ts
```diff
@@ -178,7 +258,7 @@ export class ModuleClient extends InternalClient {
     }
 
     /*Codes_SRS_NODE_MODULE_CLIENT_05_006: [The fromConnectionString method shall return a new instance of the Client object, as by a call to new Client(new transportCtor(...)).]*/
-    return new ModuleClient(new transportCtor(authenticationProvider), null);
+    return new ModuleClient(new transportCtor(authenticationProvider), new MethodClient(authenticationProvider));
```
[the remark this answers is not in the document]
BertKleewein: Note: it's technically possible to have an EdgeModule that doesn't use fromEnvironment().  I'm using this for my E2E test work because it's way easier to debug a module outside of a container.  Whether or not we allow/support this scenario is another question.  

---
In reply to: [197231663](https://github.com/Azure/azure-iot-sdk-node/pull/322#discussion_r197231663) [](ancestors = 197231663)

device/core/src/module_client.ts
BertKleewein: Yeah, I tried finding _fromEnvironmentAbnormal, but couldn't. :)

---
In reply to: [197197439](https://github.com/Azure/azure-iot-sdk-node/pull/322#discussion_r197197439) [](ancestors = 197197439)

device/core/devdoc/module_client_requirements.md
pierreca: yes - as long as it's a downstream device. think about the ML module running a workload at the edge on telemetry sent by downstream devices and then triggering something on that device.

---
In reply to: [197223596](https://github.com/Azure/azure-iot-sdk-node/pull/322#discussion_r197223596) [](ancestors = 197223596)

device/core/devdoc/module_client_requirements.md
pierreca: missed it. will fix.

---
In reply to: [197224262](https://github.com/Azure/azure-iot-sdk-node/pull/322#discussion_r197224262) [](ancestors = 197224262)

device/core/src/module_client.ts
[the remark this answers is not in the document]
pierreca: yeah it is possible and that's why i'm creating the MethodClient everywhere. which is also why it's never supposed to be null (see your earlier comment) or that'd be a bug. I feel it's simpler that way for now, until we get a good way to identify whether we're connecting to an Edge gateway or not... I'm not sure what we can do about it. One way would be to lazily initialize it when invokeMethod is called maybe? but then we have to check for its existence everywhere (setOptions etc)

---
In reply to: [197233150](https://github.com/Azure/azure-iot-sdk-node/pull/322#discussion_r197233150) [](ancestors = 197233150,197231663)

device/core/src/module_client.ts
pierreca: should never be null. if it were it'd be a bug which is why i'm fine not testing it.

---
In reply to: [197227257](https://github.com/Azure/azure-iot-sdk-node/pull/322#discussion_r197227257) [](ancestors = 197227257)

PR #325, opened 2018-06-21
fix(service): Fold invokeModuleMethod into invokeDeviceMethod

pierreca:
During the API review it was decided that there would be no invokeModuleMethod on the service client and instead the invokeDeviceMethod would be overloaded to also address the case of modules.

### DIFF
--- a/service/test/_client_test.js
+++ b/service/test/_client_test.js
@@ -201,39 +201,15 @@ describe('Client', function () {
     });
   });
 
-  describe('#invokeModuleMethod', function() {
-    /*Tests_SRS_NODE_IOTHUB_CLIENT_18_001: [The `invokeModuleMethod` shall throw a `ReferenceError` if `deviceId` or `methodId` is falsy. ]*/
-    [undefined, null, ''].forEach(function(badArg) {
-      it('throws if deviceId is ' + badArg, function() {
-        var client = new Client({}, {});
-        assert.throws(function() {
-          client.invokeModuleMethod(badArg, 'moduleId', { }, 42, function() {});
-        }, ReferenceError);
-      });
-    });
-
-    /*Tests_SRS_NODE_IOTHUB_CLIENT_18_001: [The `invokeModuleMethod` shall throw a `ReferenceError` if `deviceId` or `methodId` is falsy. ]*/
-    [undefined, null, ''].forEach(function(badArg) {
-      it('throws if moduleId is ' + badArg, function() {
-        var client = new Client({}, {});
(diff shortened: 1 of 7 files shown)


service/test/_client_test.js
```diff
@@ -201,39 +201,15 @@ describe('Client', function () {
...
       /*Tests_SRS_NODE_IOTHUB_CLIENT_16_013: [The `invokeDeviceMethod` method shall call the `done` callback with a `null` first argument, the result of the method execution in the second argument, and the transport-specific response object as a third argument.]*/
-      /*Tests_SRS_NODE_IOTHUB_CLIENT_18_002: [The `invokeModuleMethod` method shall initialize a new `DeviceMethod` instance with `methodParams` values passed in the arguments. ]*/
-      /*Tests_SRS_NODE_IOTHUB_CLIENT_18_003: [The `invokeModuleMethod` method shall call `invokeOnModule` on the new `DeviceMethod` instance. ]*/
-      /*Tests_SRS_NODE_IOTHUB_CLIENT_18_005: [The `invokeModuleMethod` method shall call the `done` callback with a `null` first argument, the result of the method execution in the second argument, and the transport-specific response object as a third argument. ]*/
+      /*Tests_SRS_NODE_IOTHUB_CLIENT_18_003: [If `moduleIdOrMethodParams` is a string the `invokeDeviceMethod` method shall call `invokeOnModule` on the new `DeviceMethod` instance. ]*/
```
anthonyvercolano: >invokeOnModule [](start = 130, length = 14)

Where is this?

pierreca: ? I dont' understand the comment

---
In reply to: [197173487](https://github.com/Azure/azure-iot-sdk-node/pull/325#discussion_r197173487) [](ancestors = 197173487)

anthonyvercolano: You have a requirement that says  invokeOnModule.  I don't see an invocation of invokeOnModule

---
In reply to: [197259039](https://github.com/Azure/azure-iot-sdk-node/pull/325#discussion_r197259039) [](ancestors = 197259039,197173487)

pierreca: it's in the invokeDeviceMethod code. the test is written so that the invocation of the executeApiCAll on the RestApiClient is considered enough of a validation. it's true that it may not be now. We can rework the test to be more explicit.

---
In reply to: [197259805](https://github.com/Azure/azure-iot-sdk-node/pull/325#discussion_r197259805) [](ancestors = 197259805,197259039,197173487)

PR #326, opened 2018-06-21
Rename onDeviceMethod to onMethod for ModuleClient

pierreca:
# Description of the problem
The ModuleClient should use onMethod instead of onDeviceMethod (the latter being confusing from a naming perspective)

### DIFF
--- a/e2etests/test/module_methods.js
+++ b/e2etests/test/module_methods.js
@@ -52,7 +52,7 @@ describe('module methods', function() {
         };
 
         debug('adding method handler for ' + methodName);
-        testModule.deviceClient.onDeviceMethod(methodName, function(request, response) {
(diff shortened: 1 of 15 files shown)


e2etests/test/module_methods.js
```diff
@@ -52,7 +52,7 @@ describe('module methods', function() {
         };
 
         debug('adding method handler for ' + methodName);
-        testModule.deviceClient.onDeviceMethod(methodName, function(request, response) {
+        testModule.deviceClient.onMethod(methodName, function(request, response) {
```
BertKleewein: >onMethod [](start = 32, length = 8)

device\samples\simple_sample_module_method.js needs to be updated too.  while you're at it, we should probably change Client->ModuleClient in that sample (and maybe the other samples in that directory)

PR #330, opened 2018-06-25
fix(device:sak_auth_prov): Start timer and emit events only when requested

pierreca:
# Description of the problem

In `SharedAccessKeyAuthenticationProvider` the timer to renew the SAS token is started immediately by the constructor - this generates events while the device or module is already connecting which generates spurious disconnect/reconnect cycles in MQTT.

On top of that starting that timer at that time is not necessary since the transport will automatically call `getDeviceCredentials` when it needs credentials.

Also when `getDeviceCredentials` is called with a callback it still emits an event, which isn't useful - events are only useful when renewed because of a timer tick.

# Description of the solution

- Move code around to only emit events when called by the timer, otherwise pass credentials (or errors) through callbacks.
- Remove timer start from constructor.


BertKleewein:
>   constructor(credentials: TransportConfig, tokenValidTimeInSeconds?: number, tokenRenewalMarginInSeconds?: number, startTimerOnNextTick?: boolean) {

remove this if unused.

---
Refers to: device/core/src/sak_authentication_provider.ts:33 in 28f1090. [](commit_id = 28f109016ff9f1a45e8e75b55bd86dea2caa656d, deletion_comment = False)

pierreca:
>   constructor(credentials: TransportConfig, tokenValidTimeInSeconds?: number, tokenRenewalMarginInSeconds?: number, startTimerOnNextTick?: boolean) {

what is unused?

---
In reply to: [400101570](https://github.com/Azure/azure-iot-sdk-node/pull/330#issuecomment-400101570) [](ancestors = 400101570)

---
Refers to: device/core/src/sak_authentication_provider.ts:33 in 28f1090. [](commit_id = 28f109016ff9f1a45e8e75b55bd86dea2caa656d, deletion_comment = False)

device/core/src/sak_authentication_provider.ts
```diff
@@ -81,11 +79,7 @@ export class SharedAccessKeyAuthenticationProvider extends EventEmitter implemen
     return (this._currentTokenExpiryTimeInSeconds - currentTimeInSeconds) < this._tokenRenewalMarginInSeconds;
   }
 
-  private _renewToken(callback?: (err: Error, credentials?: TransportConfig) => void): void {
-    if (this._renewalTimeout) {
-      clearTimeout(this._renewalTimeout);
-    }
-
```
BertKleewein: race condition?  If you call getDeviceCredentials right before the timer fires, could the timer fire during the signature operation?

device/core/src/sak_authentication_provider.ts
```diff
@@ -104,27 +98,41 @@ export class SharedAccessKeyAuthenticationProvider extends EventEmitter implemen
...
-        const nextRenewalTimeout = (this._tokenValidTimeInSeconds - this._tokenRenewalMarginInSeconds) * 1000;
-        this._renewalTimeout = setTimeout(() => this._renewToken(), nextRenewalTimeout);
+        callback(null, this._credentials);
+      }
+    });
+  }
+
+  private _expiryTimerHandler(): void {
+    if (this._renewalTimeout) {
+      clearTimeout(this._renewalTimeout);
```
BertKleewein: >clearTimeout(this._renewalTimeout); [](start = 6, length = 35)

set to null otherwise _scheduleNextExpireTimeout will call clearTimeout again.

device/core/src/sak_authentication_provider.ts
```diff
@@ -104,27 +98,41 @@ export class SharedAccessKeyAuthenticationProvider extends EventEmitter implemen
...
+    }
+
+    this._renewToken((err) => {
+      if (!err) {
+        this._scheduleNextExpiryTimeout();
         /*Codes_SRS_NODE_SAK_AUTH_PROVIDER_16_005: [Every time a new token is created, the `newTokenAvailable` event shall be fired with the updated credentials.]*/
         this.emit('newTokenAvailable', this._credentials);
```
BertKleewein: >this.emit [](start = 8, length = 9)

maybe dangerous and unnecessary, but do you want to call emit inside setImmediate so the stack has been popped?

BertKleewein: I mean dangerous to change and maybe we don't need to change it.

---
In reply to: [197950701](https://github.com/Azure/azure-iot-sdk-node/pull/330#discussion_r197950701) [](ancestors = 197950701)

device/core/src/sak_authentication_provider.ts
pierreca: You could but in theory these if() statements to clear/restart the timer should protect us from this.

---
In reply to: [197949683](https://github.com/Azure/azure-iot-sdk-node/pull/330#discussion_r197949683) [](ancestors = 197949683)

device/core/src/sak_authentication_provider.ts
pierreca: why would I pop the stack? isn't it easier to debug when we have the full stack?

---
In reply to: [197950859](https://github.com/Azure/azure-iot-sdk-node/pull/330#discussion_r197950859) [](ancestors = 197950859,197950701)

device/core/src/sak_authentication_provider.ts
```diff
@@ -42,16 +42,6 @@ export class SharedAccessKeyAuthenticationProvider extends EventEmitter implemen
     if (this._tokenValidTimeInSeconds <= this._tokenRenewalMarginInSeconds) {
       throw new errors.ArgumentError('tokenRenewalMarginInSeconds must be less than tokenValidTimeInSeconds');
     }
-
```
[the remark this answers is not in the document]
pierreca: haa correct

---
In reply to: [197948904](https://github.com/Azure/azure-iot-sdk-node/pull/330#discussion_r197948904) [](ancestors = 197948904)

pierreca:
>   constructor(credentials: TransportConfig, tokenValidTimeInSeconds?: number, tokenRenewalMarginInSeconds?: number, startTimerOnNextTick?: boolean) {

startTimerOnNextTick as Raj guessed. good point. not sure why the linter/compiler doesn't pick it up.

---
In reply to: [400102701](https://github.com/Azure/azure-iot-sdk-node/pull/330#issuecomment-400102701) [](ancestors = 400102701,400101570)

---
Refers to: device/core/src/sak_authentication_provider.ts:33 in 28f1090. [](commit_id = 28f109016ff9f1a45e8e75b55bd86dea2caa656d, deletion_comment = False)

device/core/src/sak_authentication_provider.ts
BertKleewein: It is much easier to debug, but if this event handler makes it into user code in the current state, then our code is blocked on user code.



---
In reply to: [197951178](https://github.com/Azure/azure-iot-sdk-node/pull/330#discussion_r197951178) [](ancestors = 197951178,197950859,197950701)

device/core/src/sak_authentication_provider.ts
BertKleewein: It's a small window, but it's there .
user calls getDeviceCredentials, getDeviceCredentials calls _renewToken,  after _renewToken returns, the timer is cleared and restarted.
However, if the timer goes off during _sign (the most likely place in a small window), _expiryTimerHandler will call _renewToken and schedule a new timer.
There's nothing stopping the expiry handler from renewing the token even though the user flow is already renewing it.
(this is all very unlikely and only matters if any code in this flow yields and lets the timer fire)





---
In reply to: [197950960](https://github.com/Azure/azure-iot-sdk-node/pull/330#discussion_r197950960) [](ancestors = 197950960,197949683)

device/core/src/sak_authentication_provider.ts
pierreca: aren't we the only users of this though?

---
In reply to: [197996374](https://github.com/Azure/azure-iot-sdk-node/pull/330#discussion_r197996374) [](ancestors = 197996374,197951178,197950859,197950701)

device/core/src/sak_authentication_provider.ts
pierreca: oh I see - this being said, whichever piece of code finishes last will restart the timer, right? so it's a race but it doesn't block us. in any case, it feels like maybe we need a bit more state management now that this thing is async.

---
In reply to: [197997041](https://github.com/Azure/azure-iot-sdk-node/pull/330#discussion_r197997041) [](ancestors = 197997041,197950960,197949683)

BertKleewein: I think you end up with a double-renew and then the second piece of code schedules the timer.   Not a big deal.  

---
In reply to: [197997354](https://github.com/Azure/azure-iot-sdk-node/pull/330#discussion_r197997354) [](ancestors = 197997354,197997041,197950960,197949683)

pierreca: I could cancel the timer as soon as I call getDeviceCredentials, that way I'm sure a signing operation isn't triggered by a timer, but I don't know if I can do that the other way around. 

---
In reply to: [197997631](https://github.com/Azure/azure-iot-sdk-node/pull/330#discussion_r197997631) [](ancestors = 197997631,197997354,197997041,197950960,197949683)

device/core/src/sak_authentication_provider.ts
BertKleewein: Good point.

(I'm not arguing to change this.  I'm just going through the mental exercise here.)

We're the only ones who use this event, but there's potential reentrancy issues here.  In device\transport\mqtt, the newTokenAvailable handler disconnects the transport.  I can't look in the debugger now, but it's possible that the base mqtt object is already disconnected by the time this call to emit() returns.  Now, here's where I'm waving my arms and being all theoretical, if the user happened to call Client.setOption in an event handler originally triggered by common/mqtt, then there's some event triggered by the base transport down deep in the stack, which ends up triggering a token renewal up here, which ends up triggering a disconnection higher in the stack in the middle of the emit call.  The base transport _should_ be hardened against stuff like this, but, well, you know how that goes.  

On the other hand, maybe I'm just confusing myself with too many different threading models in my head at the same time.  

The code was fine before without the setImmediate, so we shouldn't change it now.  

---
In reply to: [197997124](https://github.com/Azure/azure-iot-sdk-node/pull/330#discussion_r197997124) [](ancestors = 197997124,197996374,197951178,197950859,197950701)

Issue #285, opened 2018-05-04
clientFromConnectionString causes process to hang


gulipek95:
# Context
- **OS and version used:** Windows 10
- **Node.js version:** v8.9.4
- **npm version:** 5.6.0

# Description of the issue:
This issue came up when we were adding unit tests that used the azure-iot-device-amqp module, and the tests never finish executing. Everything else is cleaned, connections closed, but the client instance creates some hidden processes. To make sure it was the instantiation, I created a simple .js file with the content below, and after the connection is closed, it continues to run (max I waited was 5 min, but I think that is plenty enough) until I Ctrl+C.

# Code sample exhibiting the issue (this is the whole js file):
const azureClient = require('azure-iot-device-amqp');
const connectionString = <my valid connection string>;

const client = azureClient.clientFromConnectionString(connectionString);

client.open((error, result) => {
  console.log(`open error is: ${error}`);
  console.log(`open result is: ${result}`);

  console.log('about to close the client');
  client.close((err, res) => {
    console.log(`close error is: ${err}`);
    console.log(`close result is: ${res}`);

    console.log('closed the client');
  });
});

# Console log of the issue:
open error is: null
open result is: [object Object]
about to close the client
close error is: null
close result is: [object Object]
closed the client

.. and the code hangs forever, doesn't exist.

pierreca:
Thanks for reporting this. I believe it’s related to the timer that is started in the SharedAccessKeyAuthenticationProvider class. One workaround would be to use x509 certificates instead, while we validate this hypothesis and fix the bug.

gulipek95:
You're welcome - for x509 certificate, where can I use/set it to be the cert?

anthonyvercolano:
simple_sample_device_x509 in the azure-iot-sdk-node/device/samples directory demonstrates.

In the build/tools directory there is a tool called create_device_certs.js that will create a self signed cert/private keythat you could use.  The C sdk has a more sophisticated tool chain for creating a cert chain if you want.

https://github.com/Azure/azure-iot-sdk-c/blob/master/tools/CACertificates/CACertificateOverview.md


Or, you could just add a process.exit() and that would probably work too.

I'll follow up on Monday with investigation results.

gulipek95:
Cheers! Process.exit is how I am handling it right now - will wait for the updates, thank you.

anthonyvercolano:
[...]

PR #336, opened 2018-07-13
Add stop() method to AuthenticationProvider to enable shutting down timers when disconnecting/exiting

pierreca:
https://github.com/Azure/azure-iot-sdk-node/issues/285

# Description of the problem
The `SharedAccessKeyAuthenticationProvider` and the `PollingStateMachine` had timers running that could not be stopped. This was preventing the node process from exiting even after a user called `close()` on the client object.
 
# Description of the solution
- Introduced a `stop()` method on the token-based authentication providers.
- Called this from all transport disconnection paths
- Fixed all the tests that were preventing Mocha from exiting because of that.


gulipek95:
@pierreca thank you very much for addressing the issue and following up!

PR #343, opened 2018-07-28
fix(mqtt-base): Fix #307: error event sometimes emitted after mqttClient.removeAllListeners() is called

### LINKED ISSUE #307: Unhandled 'error' event - at MqttClient._handleConnack
- **list of installed packages:** 
```
azure-iot-device@1.4.1
azure-iot-device-mqtt@1.4.1
azure-iot-mqtt-base@1.4.0
mqtt@2.17.0
```

Very often our application is terminated by an unhandled error event coming from the SDK, always with the following output:

```
events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: Connection refused: Not authorized
    at MqttClient._handleConnack (/app/node_modules/mqtt/lib/client.js:893:15)
    at MqttClient._handlePacket (/app/node_modules/mqtt/lib/client.js:329:12)
    at work (/app/node_modules/mqtt/lib/client.js:271:12)
    at Writable.writable._write (/app/node_modules/mqtt/lib/client.js:281:5)
    at doWrite (/app/node_modules/mqtt/node_modules/readable-stream/lib/_stream_writable.js:428:64)
    at writeOrBuffer (/app/node_modules/mqtt/node_modules/readable-stream/lib/_stream_writable.js:417:5)
    at Writable.write (/app/node_modules/mqtt/node_modules/readable-stream/lib/_stream_writable.js:334:11)
    at TLSSocket.ondata (_stream_readable.js:639:20)
    at emitOne (events.js:116:13)
    at TLSSocket.emit (events.js:211:7)
    at addChunk (_stream_readable.js:263:12)
    at readableAddChunk (_stream_readable.js:250:11)
    at TLSSocket.Readable.push (_stream_readable.js:208:10)
    at TLSWrap.onread (net.js:607:20
```

Inspecting the SDK code, the error handler for the MQTT client is registered as soon as it's returned by `connect()`. Checking the lib MQTT code, there's not synchronous path from `connect()` to the point where the error is emitted (`handleConnack()`), so there's no possibility that it's being emitted before the client is returned.

The error handler is unregistered before the client is closed in the following part of `mqtt_base.ts`:

```
  private _disconnectClient(forceDisconnect: boolean, callback: () => void): void {
    this._mqttClient.removeAllListeners();
    /* Codes_SRS_NODE_COMMON_MQTT_BASE_16_001: [The disconnect method shall call the done callback when the connection to the server has been closed.] */
    this._mqttClient.end(forceDisconnect, callback);
  }
```

If `forceDisconnect` is false (which seems to be the case for most flows in the SDK), the client does close the underlying TCP stream when `end()` is called, thus a `connAck` packet can still be parsed and the observed error will emitted, even though we already unregistered the error handler.

MQTT lib does not provide a way to find out when the underlying stream has been effectively closed and it's safe to unregister the error handler, so we might have to request this functionality there.

AB#7366634

pierreca:
#307 

# Description of the problem
Because we call `mqttClient.removeAllListeners` before we call `mqttClient.end()` there is a possibility for an error to be emitted while/after the call to end() has finished, which leads to unhandled error exceptions.

# Description of the solution
The solution is in multiple parts:
- we need to keep listening to errors until we know no error can be emitted anymore, which means when the socket has been closed: which materializes itself with a `close` event which we don't monitor once we're connected: so we need to start handling this
- because we are leaving events "attached" to an `mqttClient` that we are disconnecting, there is a possibility, if there is a bug in the MQTT library, that the `close` event is never called and we end up with a leak or worse: that we reconnect successfully with a new mqttClient but that the old client emits an event after that: because of that we need to separate the event handlers fired after we call end() from the event handlers called while the mqttClient is actively monitored by the FSM.

### DIFF
--- a/common/transport/mqtt/src/mqtt_base.ts
+++ b/common/transport/mqtt/src/mqtt_base.ts
@@ -21,10 +21,26 @@ export class MqttBase extends EventEmitter {
   private _mqttClient: MqttClient;
   private _fsm: any;
   private _options: any;
+  private _closeEventFsmHandler: () => void;
+  private _errorEventFsmHandler: (err?: Error) => void;
+  private _messageEventFsmHandler: (topic: string, payload: any) => void;
 
   constructor(mqttprovider?: any) {
     super();
     this.mqttprovider = mqttprovider ? mqttprovider : require('mqtt');
+    this._messageEventFsmHandler = (topic, payload) => {
+      process.nextTick(() => {
+        this.emit('message', topic, payload);
+      });
+    };
+
+    this._closeEventFsmHandler = () => {
+      this._fsm.handle('closeEvent');
+    };
+
+    this._errorEventFsmHandler = (err) => {
+      this._fsm.handle('errorEvent', err);
+    };
 
     this._fsm = new machina.Fsm({
       namespace: 'mqtt-base',
@@ -76,10 +92,6 @@ export class MqttBase extends EventEmitter {
         },
         connected: {
           _onEnter: (connectCallback, conack) => {
-            this._mqttClient.on('close', () => {
-              debug('close event received from mqtt.js client - no error');
-              this._fsm.handle('closeEvent');
-            });
             connectCallback(null, new results.Connected(conack));
           },
           connect: (callback) => callback(null, new results.Connected()),
@@ -107,6 +119,9 @@ export class MqttBase extends EventEmitter {
           },
           closeEvent: () => {
             this._fsm.transition('disconnected', undefined, new errors.NotConnectedError('Connection to the server has been closed.'));
+          },
+          errorEvent: (err) => {
+            this._fsm.transition('disconnected', undefined, err);
           }
         },
         disconnecting: {
@@ -122,11 +137,15 @@ export class MqttBase extends EventEmitter {
             /*Codes_SRS_NODE_COMMON_MQTT_BASE_16_033: [The `updateSharedAccessSignature` method shall disconnect and reconnect the mqtt client with the new `sharedAccessSignature`.]*/
             /*Codes_SRS_NODE_COMMON_MQTT_BASE_16_035: [The `updateSharedAccessSignature` method shall call the `callback` argument with no parameters if the operation succeeds.]*/
             /*Codes_SRS_NODE_COMMON_MQTT_BASE_16_036: [The `updateSharedAccessSignature` method shall call the `callback` argument with an `Error` if the operation fails.]*/
+            debug('disconnecting MQTT client');
             this._disconnectClient(false, () => {
+              debug('reconnecting MQTT client');
               this._connectClient((err, connack) => {
                 if (err) {
+                  debug('Could not reconnect MQTT client');
                   this._fsm.transition('disconnected', callback, err);
                 } else {
+                  debug('MQTT client reconnected');
                   this._fsm.transition('connected', callback, connack);
                 }
               });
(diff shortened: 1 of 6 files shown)


pierreca:
cc @periclesroalves

common/transport/mqtt/src/mqtt_base.ts
```diff
@@ -253,39 +272,57 @@ export class MqttBase extends EventEmitter {
...
+      oldMqttClient = undefined;
+    });
+
+    oldMqttClient.removeListener('disconnect', this._closeEventFsmHandler);
+    oldMqttClient.removeListener('offline', this._closeEventFsmHandler);
+    oldMqttClient.removeListener('close', this._closeEventFsmHandler);
+    oldMqttClient.removeListener('message', this._messageEventFsmHandler);
+    oldMqttClient.removeListener('error', this._errorEventFsmHandler);
+
+    oldMqttClient.end(forceDisconnect, callback);
```
anthonyvercolano: >forceDisconnect [](start = 22, length = 15)

So if forceDisconnect is false, one might never see the close event?

common/transport/mqtt/test/_mqtt_base_test.js
```diff
@@ -526,5 +526,33 @@ describe('MqttBase', function () {
       });
       fakeMqtt.emit('connect');
     });
+
+    it('disconnects and emits an error if an error is received while connected', function (testCallback) {
+      var fakeMqtt = new FakeMqtt();
```
anthonyvercolano: >FakeMqtt [](start = 25, length = 8)

The fakeMqtt class references an instance variable subscribeShouldFail that is only referenced, never set.

common/transport/mqtt/test/_mqtt_base_test.js
```diff
@@ -526,5 +526,33 @@ describe('MqttBase', function () {
...
+      var mqttBase = new MqttBase(fakeMqtt);
+      var fakeError = new Error('fake');
+
+      mqttBase.on('error', function (err) {
+        assert.strictEqual(err, fakeError);
+        assert.isTrue(fakeMqtt.end.notCalled);
+        mqttBase.connect(fakeConfig, function () {
+          // second call proves that the client was disconnected the first time.
```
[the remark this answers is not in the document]
pierreca: added a better comment

---
In reply to: [206242075](https://github.com/Azure/azure-iot-sdk-node/pull/343#discussion_r206242075) [](ancestors = 206242075)

common/transport/mqtt/test/_mqtt_base_test.js
pierreca: unrelated but removed.

---
In reply to: [206242605](https://github.com/Azure/azure-iot-sdk-node/pull/343#discussion_r206242605) [](ancestors = 206242605)

pierreca:
>         const err = error || new errors.NotConnectedError('Unable to establish a connection');

unrelated but added event name.

---
In reply to: [408924538](https://github.com/Azure/azure-iot-sdk-node/pull/343#issuecomment-408924538) [](ancestors = 408924538)

---
Refers to: common/transport/mqtt/src/mqtt_base.ts:270 in 306f25f. [](commit_id = 306f25f023da5430d6d09ff681a4eb41d2b1ba08, deletion_comment = False)

common/transport/mqtt/src/mqtt_base.ts
pierreca: I believe we probably already got a close event if that were the case. forceDisconnect is used when the socket has already failed.

---
In reply to: [206239361](https://github.com/Azure/azure-iot-sdk-node/pull/343#discussion_r206239361) [](ancestors = 206239361)

pierreca:
closing as terribly outdated.

PR #344, opened 2018-08-06
Migration to the rhea AMQP library

anthonyvercolano:
# Description of the problem
Need to switch from amq10.

# Description of the solution
Latest work on utilizing the rhea library.


common/transport/amqp/package.json
```diff
@@ -25,12 +24,19 @@
...
     "sinon": "^4.0.2",
+    "sinon-test": "^2.2.1",
     "tslint": "^5.1.0",
     "typescript": "2.5.2"
   },
   "scripts": {
     "lint": "tslint --project . -c ../../../tslint.json",
     "build": "tsc",
+    "all_together": "istanbul cover node_modules/mocha/bin/_mocha -- --reporter spec test/_amqp_cbs_test.js test/_amqp_common_errors_test.js test/_amqp_link_test.js test/_amqp_message_test.js test/_sender_link_test.js test/_receiver_link_test.js",
+    "sender_link": "istanbul cover node_modules/mocha/bin/_mocha -- --reporter spec test/_sender_link_test.js",
```
pierreca: use a glob pattern instead? test/_*_test.js?

common/transport/amqp/typings.json
```diff
@@ -1,8 +1,4 @@
 {
   "name": "azure-iot-amqp-base",
-  "main": "index.d.ts",
-  "dependencies": {
-    "amqp10": "registry:npm/amqp10#3.3.1+20170124180914",
-    "bluebird": "registry:npm/bluebird#3.5.0+20170314181206"
-  }
+  "main": "index.d.ts"
 }
```
pierreca: I think we can remove this file alltogether.

pierreca:
> const amqp10Client = new amqp10.AmqpClient(null);

this sample probably needs to be adjusted

---
Refers to: common/transport/amqp/devdoc/amqp_cbs_requirements.md:15 in 90bae3c. [](commit_id = 90bae3c1d5365f09ebb5dc9b8a82c73ef6ef3f47, deletion_comment = False)

pierreca:
> **SRS_NODE_AMQP_CBS_16_001: [** The `constructor` shall instantiate a `SenderLink` object for the `$cbs` endpoint using a custom policy `{encoder: function(body) { return body;}}` which forces the amqp layer to send the token as an amqp value in the body. **]**

is that still valid?

---
Refers to: common/transport/amqp/devdoc/amqp_cbs_requirements.md:37 in 90bae3c. [](commit_id = 90bae3c1d5365f09ebb5dc9b8a82c73ef6ef3f47, deletion_comment = False)

common/transport/amqp/devdoc/amqp_cbs_requirements.md
```diff
@@ -93,8 +93,8 @@ and a body containing `<sasToken>`. **]**
 
 ### events
 
-**SRS_NODE_AMQP_CBS_16_016: [** If either the sender or receiver link emits an `error` event, remaining links shall be detached. **]**
-
+**SRS_NODE_AMQP_CBS_06_001: [** If in the attached state, either the sender or the receiver links gets an error, an error of `azure-iot-amqp-base:error-indicated` will have been indicated on the container object and the cbs will remain in the attached state.  The owner of the cbs MUST detach.
```
pierreca: that does not sound like a test or a test scenario.

common/transport/amqp/devdoc/amqp_message.md
```diff
@@ -77,20 +75,20 @@ The `toMessage` static method takes an AMQP message from the underlying library
 
 **SRS_NODE_IOTHUB_AMQPMSG_16_002: [** The `toMessage` method shall return a `Message` object. **]**
 
-**SRS_NODE_IOTHUB_AMQPMSG_16_003: [** The `toMessage` method shall set the `Message.correlationId` property to the `AmqpMessage.properties.correlationId` value if it is present. **]**
+**SRS_NODE_IOTHUB_AMQPMSG_16_003: [** The `toMessage` method shall set the `Message.correlationId` property to the `AmqpMessage.correlation_id` value if it is present. **]**
```
pierreca: no buffer translation on this part?

common/transport/amqp/devdoc/amqp_requirements.md
```diff
@@ -48,8 +48,7 @@ A string containing the version of the SDK used for telemetry purposes**]**
 
 **SRS_NODE_COMMON_AMQP_16_042: [** The Amqp constructor shall create a new `amqp10.Client` instance and configure it to:
```
pierreca: >amqp10.Client [](start = 76, length = 13)

this should probably be adjusted for rhea-specific object.

common/transport/amqp/devdoc/receiver_link_requirements.md
```diff
@@ -4,71 +4,60 @@
...
+**SRS_NODE_AMQP_RECEIVER_LINK_16_020: [** The `attach` method shall call the `callback` if the link was successfully attached **]**
 
-**SRS_NODE_AMQP_RECEIVER_LINK_16_008: [** If the `amqp10.AmqpClient` fails to create the link the `callback` function shall be called with this error object. **]**
+**SRS_NODE_AMQP_RECEIVER_LINK_06_001: [** The `attach` method shall immediately invoke the `callback` if already in an attached state. **]**
 
-**SRS_NODE_AMQP_RECEIVER_LINK_16_020: [** The `attach` method shall call the `callback` if the link was successfully attached **]**
+**SRS_NODE_AMQP_RECEIVER_LINK_06_002: [** If the `detach` method is invoked on the `ReceiverLink` while still attaching, the ReceiverLink shall detach.  It will indicate the error to the callback for the `detach` as well as the callback to the `attach`. **]**
+
+**SRS_NODE_AMQP_RECEIVER_LINK_06_003: [** If the `forceDetach` method is invoked on the `ReceiverLink` while still attaching, the ReceiverLink shall detach.  It will indicate the error to the callback to the `attach`. **]**
```
pierreca: > It will indicate the error to the callback to the `attach`. [](start = 157, length = 60)

this sounds very high-level/generic. it should describe what is being called with what arguments.

common/transport/amqp/devdoc/sender_link_requirements.md
```diff
@@ -4,73 +4,65 @@
...
-**SRS_NODE_AMQP_SENDER_LINK_16_009: [** The `detach` method shall detach the link created by the `amqp10.AmqpClient` underlying object. **]**
+**SRS_NODE_AMQP_SENDER_LINK_06_005: [** If the `SenderLink` is already in the detached state an invocation of `detach` shall immediately invoke the callback with the (potentially) supplied error parameter. **]**
+
+**SRS_NODE_AMQP_SENDER_LINK_16_009: [** The `detach` method shall detach the link created by `rhea`. **]**
+
+**SRS_NODE_AMQP_SENDER_LINK_06_009: [** If `forceDetach` invoked while detaching, the detach will be completed with the error supplied to the `forceDetach` or an error indicating that the `detach` was preempted by the `forceDetach`. **]**
+
+**SRS_NODE_AMQP_SENDER_LINK_06_010: [** An error occurring during a detach will be indicated in the error result of the `detach`. **]**
+
+**SRS_NODE_AMQP_SENDER_LINK_06_011: [** If `detach` invoked while already detaching, it's callback will be invoked with an error.  Whatever caused the original detaching will proceed. **]**
```
pierreca: > Whatever caused the original detaching will proceed. [](start = 130, length = 53)

what does that mean? I'm not sure it adds to the test description or understanding of the situation. try to think of requirements as descriptions for the target as a black box, regardless of the calling context.

common/transport/amqp/src/amqp.ts
```diff
@@ -3,23 +3,20 @@
 
 'use strict';
 
-import * as amqp10 from 'amqp10';
 import * as machina from 'machina';
+import urlParser = require('url');
```
pierreca: >require('url') [](start = 19, length = 14)

 why use require instead of import from?

common/transport/amqp/src/amqp.ts
```diff
@@ -34,12 +31,90 @@ export type GenericAmqpBaseCallback<T> = (err: Error | null, result?: T) => void
...
+  //
+  // Top of the rhea object hierarchy.
+  //
+  private _rheaContainer: Container;
+  //
+  // Rhea object used to represent the connection.  The connection object is a child of the container object.
+  // It is an event emitter.  Numerous handlers listen to manage setting up a connection and act upon error events
+  // associated with the connection.
+  //
+  private _rheaConnection: Connection;
```
pierreca: >_rheaConnection [](start = 10, length = 15)

- aren't the connection and the session part of the container or available in a context object?
- is there a risk that we are keeping references to things that the container has already gotten rid of?

common/transport/amqp/src/amqp.ts
```diff
@@ -34,12 +31,90 @@ export type GenericAmqpBaseCallback<T> = (err: Error | null, result?: T) => void
...
+  // A dictionary to handle looking up THIS transport's (NOT rhea) link objects.  The key is the actual name of the link.
+  // The value is the link object.
+  //
   private _senders: { [key: string]: SenderLink; } = {};
+  //
+  // We utilize this property to hold the callback that was provided when entering a state.  This callback
+  // is invoked when an operation is completed.  Due to the event driven nature of rhea, multiple handlers
+  // will be invoked while in a state.  We need a place to keep the callback as event progression occurs.
+  //
+  private _connectionCallback: (err: Error | AmqpError, result?: any) => void;
```
pierreca: >_connectionCallback: (err: Error | AmqpError, result?: any) => void; [](start = 10, length = 68)

This is pretty much a global variable. have you explored the possbility of passing the callback from one state to the next using arguments?

common/transport/amqp/src/amqp.ts
```diff
@@ -48,70 +123,85 @@ export class Amqp {
...
+    - not reattach sender and receiver links on failure]*/
 
-    const amqpErrorHandler = (err) => {
-      debug('amqp10 client error: ' + err.toString());
+    this._rheaContainer = rheaCreateContainer();
+    this._rheaContainer.on('error', (context: EventContext) => {
+      debug('Container received a error event' + this._getErrorName(context.connection.error));
+    });
+
+    this._rheaContainer.on('azure-iot-amqp-base:error-indicated', (err: AmqpError) => {
```
pierreca: >this._rheaContainer.on('azure-iot-amqp-base:error-indicated' [](start = 4, length = 60)

this is a risky design pattern because we do not control the nature/type of the rhea container (what if it isn't an EventEmitter anymore?). it tells me we lack something in the current patterns that _we_ use. can we work on identifying the gap in our code, rather than rely on the modification of someone else's? or is rhea missing a thing and we need to ask for a change?

common/transport/amqp/src/amqp.ts
```diff
@@ -48,70 +123,85 @@ export class Amqp {
...
-      process.nextTick(() => {
-        this._fsm.handle('amqpDisconnected');
-      });
     });
 
+    // //
+    // // This error handler listens for events from THIS transport SenderLinks or ReceiverLinks.
+    // // These error indications would be as a result of an error indication from the associated
+    // // rhea links that are not associated with a specific request.
+    // //
```
pierreca: nit: double-commented lines

common/transport/amqp/src/amqp.ts
```diff
@@ -48,70 +123,85 @@ export class Amqp {
...
+    // const amqpErrorHandler = (err) => {
+    //   debug('amqpErrorHandler invoked ' + this._getErrorName(err));
+    //   this._fsm.handle('amqpError', err);
+    // };
```
pierreca: remove?

common/transport/amqp/src/amqp.ts
```diff
@@ -48,70 +123,85 @@ export class Amqp {
...
+
+    const rheaErrorHandler = (context: EventContext) => {
+      debug('rhea error event handler');
+      this._fsm.handle('error', context);
+    };
+
+    const connectionErrorHandler = (context: EventContext) => {
+      debug('connection error event handler');
+      this._fsm.handle('connection_error', context);
```
pierreca: >connection_error [](start = 24, length = 16)

are those the exact names of the events we are receiving? do we need to wrap those in a function or can we just use this._fsm.handle directly as the event handler?

common/transport/amqp/src/amqp.ts
```diff
@@ -145,44 +235,190 @@ export class Amqp {
...
-                this._fsm.transition('connected', connectCallback, result);
-                return null;
-              })
-              .catch((err) => {
-                this._amqp.removeListener(_amqpClientError, connectErrorHandler);
-                /*Codes_SRS_NODE_COMMON_AMQP_16_003: [The `connect` method shall call the `done` callback if the connection fails.] */
-                this._fsm.transition('disconnected', connectCallback, connectError || err);
-              });
+          _onEnter: (connectionParameters, connectCallback) => {
+            this._rheaContainer.options.sender_options = {properties: {'com.microsoft:client-version': this._config.userAgentString}, reconnect: false};
```
pierreca: >{properties: {'com.microsoft:client-version': this._config.userAgentString}, reconnect: false}; [](start = 57, length = 95)

nit: please make those multi-lines for readability

pierreca:
>           detachSenderLink: (endpoint, callback) => this._safeCallback(callback),

why are some of those using safeCallback and not others? 

---
Refers to: common/transport/amqp/src/amqp.ts:232 in 90bae3c. [](commit_id = 90bae3c1d5365f09ebb5dc9b8a82c73ef6ef3f47, deletion_comment = False)

common/transport/amqp/src/amqp.ts
```diff
@@ -145,44 +235,190 @@ export class Amqp {
...
+            //
+            // According to the rhea maintainers, one can depend on that fact that no actual network activity
+            // will occur until the nextTick() after the call to connect.  Because of that, one can
+            // put the event handlers on the rhea connection object returned from the connect call and be assured
+            // that the listeners are in place BEFORE any possible events will be emitted on the connection.
+            //
+            this._rheaConnection = this._rheaContainer.connect(connectionParameters);
+            manageConnectionHandlers('on');
+          },
+          'connection_open': (context: EventContext) => {
```
pierreca: >connection_open [](start = 11, length = 15)

nit: no quotes necessary

common/transport/amqp/src/amqp.ts
```diff
@@ -145,44 +235,190 @@ export class Amqp {
...
+            this._sessionCloseOccurred = false;
+            this._connectionCloseOccurred = false;
+            //
+            // According to the rhea maintainers, one can depend on that fact that no actual network activity
+            // will occur until the nextTick() after the call to connect.  Because of that, one can
+            // put the event handlers on the rhea connection object returned from the connect call and be assured
+            // that the listeners are in place BEFORE any possible events will be emitted on the connection.
+            //
+            this._rheaConnection = this._rheaContainer.connect(connectionParameters);
+            manageConnectionHandlers('on');
```
pierreca: >manageConnectionHandlers('on'); [](start = 12, length = 31)

nit: naming is a little generic - try to find something more descriptive: installConnectionEventHandlers?

common/transport/amqp/src/amqp.ts
```diff
@@ -145,44 +235,190 @@ export class Amqp {
...
+            let err = this._indicatedConnectionError;
+            let callback = this._connectionCallback;
+            this._indicatedConnectionError = undefined;
+            this._connectionCallback = undefined;
+            this._connectionCloseOccurred = true;
+            manageConnectionHandlers('removeListener');
+            this._fsm.transition('disconnected', callback, err);
+          },
+          'connection_error': (context: EventContext) => {
+            debug('In the connection_error handler for the connecting state: ' + context.connection.name + ' with error of: ' + this._getErrorName(context.connection.error));
```
pierreca: >In the connection_error handler for the connecting state [](start = 19, length = 56)

nit: those long-form sentences make it harder to parse logs than coming up with a standard string format such as:
"State: <state_name>; Event: <event_name>; Connection: <connection_name>"

common/transport/amqp/src/amqp.ts
```diff
@@ -145,44 +235,190 @@ export class Amqp {
...
+            debug('Deferring transition until the connection_close occurs.');
```
pierreca: >debug('Deferring transition until the connection_close occurs.'); [](start = 12, length = 65)

you're not deferring anything here, i think the debug statement is misleading especially considering how "defer" means something else in the context of machina.js

common/transport/amqp/src/amqp.ts
```diff
@@ -145,44 +235,190 @@ export class Amqp {
...
+          },
+          'connection_close': (context: EventContext) => {
+            debug('In the connection_close handler for the connecting_session state with indicated error of: ' + this._getErrorName(this._indicatedConnectionError));
+            debug('Transition to disconnecting');
+            let err = this._indicatedConnectionError;
+            let callback = this._sessionCallback;
+            //
+            // We lie about session close coming in.  Thing is that if we are here we don't actually have
+            // a good session set up.  This way we won't wait around for a session end that probably won't come.
+            //
```
pierreca: this is the "global state" anti-pattern coming back to bite you. you now have to manage consistency of that state which is a collection of disjointed variables.

common/transport/amqp/src/amqp.ts
```diff
@@ -239,8 +477,9 @@ export class Amqp {
           },
           attachReceiverLink: (endpoint: string, linkOptions: any, done: GenericAmqpBaseCallback<ReceiverLink>): void => {
             debug('creating receiver link for: ' + endpoint);
-            this._receivers[endpoint] = new ReceiverLink(endpoint, linkOptions, this._amqp);
+            this._receivers[endpoint] = new ReceiverLink(endpoint, linkOptions, this._rheaSession);
             const permanentErrorHandler = (err) => {
+              debug('receiver link invokes perm error handler for attachReceiverLink: ' + endpoint + ': ' + err.toString());
```
pierreca: >perm [](start = 43, length = 4)

nit: permanent?

common/transport/amqp/src/amqp.ts
```diff
@@ -311,19 +551,52 @@ export class Amqp {
...
+                if (this._sessionCloseOccurred) {
                   callback();
-                  return null;
-                }).catch((err) => {
-                  debug('amqp10 failed to cleanly disconnect: ' + err.toString());
+                } else {
+                  this._sessionCallback = callback; // So that the session_close handler for this disconnecting state can invoke the callback.
+                }
+              }
+            };
```
pierreca: this tells me "disconnecting" needs to be broken down into 2 states at least (for session and connection).

common/transport/amqp/src/amqp.ts
```diff
@@ -568,17 +866,29 @@ export class Amqp {
...
   }
 
+  private _getErrorName(err: any): string {
+    if (err) {
+      if (err.condition) {
+        return '(amqp error) ' + err.condition;
+      } else if (err.hasOwnProperty('name')) {
+        return '(javascript error) ' + err.name;
+      } else {
+        return 'this is not an error type I understand';
```
pierreca: >'this is not an error type I understand' [](start = 15, length = 40)

I'm not sure how this helps anyone. who is "I"? why not?

common/transport/amqp/package.json
anthonyvercolano: Actually all of these can be removed.  Just was adding them as I was going through the unit test exercise.

---
In reply to: [209123651](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209123651) [](ancestors = 209123651)

common/transport/amqp/typings.json
anthonyvercolano: Done.

---
In reply to: [209123932](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209123932) [](ancestors = 209123932)

common/transport/amqp/devdoc/amqp_cbs_requirements.md
anthonyvercolano: There is an actual test for this with the text:

remains in the attached state when error emitted to container from link

---
In reply to: [209124407](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209124407) [](ancestors = 209124407)

common/transport/amqp/devdoc/amqp_message.md
anthonyvercolano: Hmmm, perhaps there should be another requirement or two.  If the amqp property correlation_id is of type buffer AND is of length 16 it will do a rhea uuid type to string coversion.  Otherwise it will simply set the agnostic correlationId to the amqp correlation_id.

---
In reply to: [209124547](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209124547) [](ancestors = 209124547)

common/transport/amqp/devdoc/receiver_link_requirements.md
anthonyvercolano: Changing text to: With the error supplied to the forceDetach, the `attach` callback will also be invoked.

---
In reply to: [209125261](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209125261) [](ancestors = 209125261)

common/transport/amqp/devdoc/sender_link_requirements.md
```diff
@@ -4,73 +4,65 @@
...
 
-**SRS_NODE_AMQP_SENDER_LINK_16_006: [** The `SenderLink` object should subscribe to the `detached` event of the newly created `amqp10` link object. **]**
+**SRS_NODE_AMQP_SENDER_LINK_06_007: [** The `attach` method shall immediately invoke the `callback` if already in an attached state. **]**
 
-**SRS_NODE_AMQP_SENDER_LINK_16_007: [** The `SenderLink` object should subscribe to the `errorReceived` event of the newly created `amqp10` link object. **]**
+**SRS_NODE_AMQP_SENDER_LINK_06_001: [** If the `detach` method is invoked on the `SenderLink` while still attaching, the SenderLink shall detach.  It will indicate the error to the callback for the `detach` as well as the callback to the `attach` **]**
 
-**SRS_NODE_AMQP_SENDER_LINK_16_008: [** If the `amqp10.AmqpClient` fails to create the link the `callback` function shall be called with this error object. **]**
+**SRS_NODE_AMQP_SENDER_LINK_06_003: [** If the `forceDetach` method is invoked on the `SenderLink` while still attaching, the SenderLink shall detach.  It will indicate the error to the callback to the `attach` and emit an `error` event on the `SenderLink`.**]**
```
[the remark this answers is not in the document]
anthonyvercolano: Both do.  Changed requirments for sender and recevier to read:  With the error supplied to the forceDetach, the `attach` callback will also be invoked.  If the error is NOT falsy it will also be emitted as the argument to the `error` event.] */
        it('forceDetach while attaching invokes the attach callback with an error (if provided) and emits an `error` event with the (potentially) supplied error.

---
In reply to: [209125785](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209125785) [](ancestors = 209125785)

common/transport/amqp/src/amqp.ts
anthonyvercolano: I vaguely remember adding this.  I think I had some difficulty doing from?  I'll go back and redo and see what happens.

---
In reply to: [209127671](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209127671) [](ancestors = 209127671)

common/transport/amqp/src/amqp.ts
anthonyvercolano: Don't believe that there is a direct property from the container to the connection.

As for the context,  You only have those when handling and error.

---
In reply to: [209127965](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209127965) [](ancestors = 209127965)

common/transport/amqp/src/amqp.ts
anthonyvercolano: The whole issue here to me was the event driven nature of rhea.  This was the best I could come up with in dealing with the multi event nature of how rhea works.

---
In reply to: [209129265](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209129265) [](ancestors = 209129265,209128149)

common/transport/amqp/src/amqp.ts
```diff
@@ -48,70 +123,85 @@ export class Amqp {
...
-        forever: false
-      }
-    }, amqp10.Policy.EventHub));
+    - not reattach sender and receiver links on failure]*/
 
-    const amqpErrorHandler = (err) => {
-      debug('amqp10 client error: ' + err.toString());
+    this._rheaContainer = rheaCreateContainer();
+    this._rheaContainer.on('error', (context: EventContext) => {
+      debug('Container received a error event' + this._getErrorName(context.connection.error));
```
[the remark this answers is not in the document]
anthonyvercolano: This was particularly handy during development.  Always indicated a bug on my part.  It should be removed.

---
In reply to: [209129776](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209129776) [](ancestors = 209129776)

common/transport/amqp/src/amqp.ts
anthonyvercolano: To change the container or for that matter connections, links, sessions from event emitters would be a HUGE breaking architectural change for rhea.  Not. Going. To. Happen.

I need to emit an event up to this class.    This is what I have.  Give me an alternative and I'll consider.

---
In reply to: [209129983](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209129983) [](ancestors = 209129983)

common/transport/amqp/src/amqp.ts
anthonyvercolano: Done

---
In reply to: [209130673](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209130673) [](ancestors = 209130673)

common/transport/amqp/src/amqp.ts
anthonyvercolano: Done.

---
In reply to: [209130704](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209130704) [](ancestors = 209130704)

common/transport/amqp/src/amqp.ts
anthonyvercolano: Why did I put these there?
Removed/

---
In reply to: [209131428](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209131428) [](ancestors = 209131428)

common/transport/amqp/src/amqp.ts
anthonyvercolano: Not for me when I was debugging the code.  Will remove the debug.

---
In reply to: [209132055](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209132055) [](ancestors = 209132055)

common/transport/amqp/src/amqp_cbs.ts
```diff
@@ -217,22 +225,25 @@ export class ClaimsBasedSecurityAgent {
...
-              to: '$cbs',
-              messageId: uuid.v4(),
-              reply_to: 'cbs'
-            };
+            amqpMessage.to = '$cbs';
+            //
+            // Just as a reminder:  For cbs, the message id and the correlation id
+            // always stayed as string values.  They never went on over the wire
+            // as the amqp uuid type.
+            //
```
pierreca: minor nit: the way this comment is phrased ("Just as a reminder", "never went over the wire") refers to tribal knowledge and seem to indicate a context that someone in the future won't have. I would suggest simply stating the fact/protocol: "For CBS, the message id and correlation id are encoded as string".

common/transport/amqp/src/amqp_cbs.ts
```diff
@@ -274,6 +285,7 @@ export class ClaimsBasedSecurityAgent {
             const links = [this._senderLink, this._receiverLink];
             async.each(links, (link, callback) => {
               if (link) {
+                debug('while detaching for  link ');
```
pierreca: >  link  [](start = 42, length = 7)

minor nit: you have extra spaces there.

common/transport/amqp/src/amqp_link_interface.ts
```diff
@@ -1,5 +1,7 @@
+import { AmqpError } from 'rhea';
+
 export interface AmqpLink {
   attach: (callback: (err?: Error) => void) => void;
-  detach: (callback: (err?: Error) => void) => void;
+  detach: (callback: (err?: Error) => void, err?: Error | AmqpError) => void;
```
pierreca: >callback: (err?: Error) => void, err?: Error | AmqpError [](start = 11, length = 56)

argument order should be reversed. I'm ok with breaking the API, a lot less ok with breaking the callback pattern

common/transport/amqp/src/amqp_message.ts
```diff
@@ -3,14 +3,22 @@
...
+/* tslint:disable:variable-name */
+/* tslint:enable:variable-name: [true, "check-format", allow-leading-underscore", "ban-keywords", "allow-snake-case"]*/
 
 function encodeUuid(uuidString: string): any {
   const uuidRegEx = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/;
   let uuid;
   if (typeof uuidString === 'string' && uuidString.match(uuidRegEx)) {
-    uuid = (<any>amqp10).Type.uuid(uuidString);
+    //
+    // SERIALIZATION DEPENDENCY!!
```
pierreca: >SERIALIZATION DEPENDENCY!! [](start = 7, length = 26)

nit: that specific line does not help

common/transport/amqp/src/amqp.ts
anthonyvercolano: It is stating exactly what I meant.  Will remove these debugs.

---
In reply to: [209132217](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209132217) [](ancestors = 209132217)

common/transport/amqp/src/receiver_link.ts
```diff
@@ -21,57 +28,119 @@ const debug = dbg('azure-iot-amqp-base:ReceiverLink');
...
+  //
+  // When we receive a message from the service we send it up to the application.  We have no say as to how we
+  // dispose of the message.
+  //
+  // We need for the application to indicate the disposition.
+  //
+  // We create a dictionary that is keyed by the amqpMessage.  The value of the dictionary is the delivery record associated
+  // (by rhea) of this message.
+  //
+  // So when a message is received we place in the dictionary.  When the application sends back down the message we look it up
```
pierreca: >sends back down the message [](start = 85, length = 27)

you mean "when the user of the SDK settles the message by accepting, abandoning or rejecting it"?

common/transport/amqp/src/receiver_link.ts
```diff
@@ -21,57 +28,119 @@ const debug = dbg('azure-iot-amqp-base:ReceiverLink');
...
+  // in the dictionary, remove it if found, and issue the disposition.  Note we will indicate an error to the callback (if provided),
+  // if we can't find the item in the dictionary.
+  //
+  // Note also, if the application does NOT provide an indication on what to do with the message this dictionary will NOT
+  // remove the item.  This, could be considered a leak.  If so, we could implement a cleaner for the list.  However, for
+  // now this is treated as an application error.
+  //
+  private _unDisposedDeliveries: DeliveryRecord[];
```
pierreca: >_unDisposedDeliveries [](start = 10, length = 21)

nit: casing is weird.

common/transport/amqp/src/amqp.ts
anthonyvercolano: I disagree.  We are dealing with a fairly bizarre fever dream of error conditions.  But we deal with it.  Run downs are always hard.

---
In reply to: [209132891](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209132891) [](ancestors = 209132891)

common/transport/amqp/src/receiver_link.ts
```diff
@@ -21,57 +28,119 @@ const debug = dbg('azure-iot-amqp-base:ReceiverLink');
...
+    const receiverErrorHandler  = (context: EventContext): void => {
+        this._fsm.handle('receiverErrorEvent', context);
     };
 
-    this._errorHandler = (err: Error): void => {
-      debug('handling error event: ' + err.toString());
-      this._fsm.handle('forceDetach', err);
+    /*Codes_SRS_NODE_AMQP_RECEIVER_LINK_16_012: [If a `message` event is emitted by the `rhea` link object, the `ReceiverLink` object shall emit a `message` event with the same content.]*/
+    const receiverMessageHandler = (context: EventContext): void => {
+      this._unDisposedDeliveries.push({msg: context.message as any, delivery: context.delivery});
```
pierreca: > as any [](start = 59, length = 7)

why do we need to cast this?

common/transport/amqp/src/receiver_link.ts
```diff
@@ -21,57 +28,119 @@ const debug = dbg('azure-iot-amqp-base:ReceiverLink');
...
       states: {
         detached: {
           _onEnter: (callback, err) => {
-            this._linkObject = null;
+            this._rheaReceiver = null;
+            this._rheaReceiverName = null;
 
             if (callback) {
-              callback(err);
+              this._safeCallback(callback,err);
```
pierreca: >,err [](start = 41, length = 4)

nit: spacing. also: what's the use of _safeCallback here?

pierreca: it seems you're using safeCallback almost everywhere. it blows up the stack and ruins the debugging experience, also makes writing unit tests much harder. why is it necessary to nextTick all that stuff? or to test if a callback is there or not, considering it should always be there? in this specific case, you're even going to test if the callback is not undefined twice before calling it


---
In reply to: [209352713](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209352713) [](ancestors = 209352713)

common/transport/amqp/src/receiver_link.ts
```diff
@@ -80,98 +149,236 @@ export class ReceiverLink  extends EventEmitter implements AmqpLink {
...
+            this._receiverCloseOccurred = true;
+            this._fsm.transition('detached', callback, error);
+          },
+          attach: (null),
+          detach: (callback, err) => {
+            /*Codes_SRS_NODE_AMQP_RECEIVER_LINK_06_002: [If the `detach` method is invoked on the `ReceiverLink` while still attaching, the ReceiverLink shall detach.  It will indicate the error to the callback for the `detach` as well as the callback to the `attach`.] */
+            debug('Detaching while attaching of rhea receiver link ' + this._linkAddress);
+            manageReceiverHandlers('removeListener');
+            //
+            // We may have a callback outstanding on the request that started the attaching.
```
pierreca: >We may have a callback outstanding on the request that started the attaching. [](start = 15, length = 77)

this sentence is hard to understand (try reading it out loud :))

common/transport/amqp/src/receiver_link.ts
```diff
@@ -80,98 +149,236 @@ export class ReceiverLink  extends EventEmitter implements AmqpLink {
...
+            // We will signal to that callback that an error occurred. We will also invoke the callback supplied
+            // for this detach.
```
pierreca: wait... we have 2 callbacks to call here? that's really weird. there should be a way to call a single callback that unrolls all the way to the original operation.

common/transport/amqp/src/receiver_link.ts
```diff
@@ -80,98 +149,236 @@ export class ReceiverLink  extends EventEmitter implements AmqpLink {
...
+            let error = err || this._indicatedError || new Error('Unexpected link detach while attaching');
+            let attachingCallback = this._attachingCallback;
+            this._indicatedError = undefined;
+            this._attachingCallback = undefined;
+            if (attachingCallback) {
+              attachingCallback(error);
             }
-
-            this._fsm.transition('detached', undefined, err);
+            this._fsm.transition('detached', callback, error);
```
pierreca: something feels awkward about this.
1. you have to maintain your global state separately of the FSM state which is error prone
2. you're calling 2 callbacks, for attaching and then passing another callback around to detach (which will call that callback with the error you pass down). Are there 2 callers with 2 concurrent operations that each need their own callback?

common/transport/amqp/src/receiver_link.ts
```diff
@@ -80,98 +149,236 @@ export class ReceiverLink  extends EventEmitter implements AmqpLink {
...
-            this._safeCallback(callback, err);
+            if (callback) callback(err);
+          },
+          receiverErrorEvent: (context: EventContext) => {
+            debug('in receiver attached state - error event for ' + context.receiver.name + ' error is: ' + this._getErrorName(context.receiver.error));
+            this._indicatedError = context.receiver.error;
+            //
+            // We don't transition at this point in that we are guaranteed that the error will be followed by a receiver_close
+            // event.
+            //
```
pierreca: does that need an intermediate state that is "errored but not closed yet"?

common/transport/amqp/src/amqp.ts
anthonyvercolano: Done

---
In reply to: [209131170](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209131170) [](ancestors = 209131170)

common/transport/amqp/src/amqp.ts
anthonyvercolano: Gone.

---
In reply to: [209133104](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209133104) [](ancestors = 209133104)

common/transport/amqp/src/amqp_message.ts
```diff
@@ -3,14 +3,22 @@
 
 'use strict';
 
+import { uuid_to_string as rheaUuidToString, string_to_uuid as rheaStringToUuid, types as rheaTypes, message as rheaMessage } from 'rhea';
```
BertKleewein: nit: couldn't you do import * as rhea from 'rhea' then used reah.uuid_to_string, etc, instead of redefining each individual name?

common/transport/amqp/src/amqp.ts
anthonyvercolano: I'm assuming that it's not simply the }; that is signaling you here.

Really couldn't disagree more.  These are fairly intertwined concepts, for instance the session code is going to have to deal with the fact that the connection has already been disconnected where for the happy path you should be able to do the session stuff before doing the connection stuff.  Doesn't work that way.  A boatload of cases have come through this code.   Trying to separate them will only produce very repetitive code and LESS systematic treatment.

---
In reply to: [209133207](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209133207) [](ancestors = 209133207)

common/transport/amqp/src/amqp_message.ts
BertKleewein: additional comment: rheaMessage is not accurate.  maybe call this messageUtils or something similar?

---
In reply to: [209365550](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209365550) [](ancestors = 209365550)

common/transport/amqp/src/amqp.ts
anthonyvercolano: This code was ONLY used in debug statements.  I here refers to this function.  It's there because while I (in this case me) couldn't think of what would fall through to the else, it seemed more complete to put some kind of return for the else.

Changed test to 'unknown error type'

---
In reply to: [209133721](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209133721) [](ancestors = 209133721)

common/transport/amqp/src/amqp_message.ts
```diff
@@ -141,47 +144,59 @@ export class AmqpMessage {
...
+    if (amqpMessage.absolute_expiry_time) {
+      msg.expiryTimeUtc = amqpMessage.absolute_expiry_time;
+    }
 
-      /*Codes_SRS_NODE_IOTHUB_AMQPMSG_16_004: [The `toMessage` method shall set the `Message.messageId` property to the `AmqpMessage.properties.messageId` value if it is present.]*/
-      if (amqpMessage.properties.messageId) {
-        msg.messageId = amqpMessage.properties.messageId;
+    //
+    // SERIALIZATION DEPENDENCY!!
+    // The rhea library will de-serialize an encoded uuid (0x98) as a 16 byte buffer.
```
BertKleewein: >(0x98) [](start = 58, length = 6)

Silly question, and ignore me if this is dumb, but is this 0x98 type identifier available to you here, or has it already been discarded at this point?  (I assume it's been discarded otherwise you wouldn't have written the code this way, but I still have to ask).

common/transport/amqp/src/amqp_message.ts
```diff
@@ -141,47 +144,59 @@ export class AmqpMessage {
...
-      /*Codes_SRS_NODE_IOTHUB_AMQPMSG_16_003: [The `toMessage` method shall set the `Message.correlationId` property to the `AmqpMessage.properties.correlationId` value if it is present.]*/
-      if (amqpMessage.properties.correlationId) {
-        msg.correlationId = amqpMessage.properties.correlationId;
+    /*Codes_SRS_NODE_IOTHUB_AMQPMSG_16_003: [The `toMessage` method shall set the `Message.correlationId` property to the `AmqpMessage.correlation_id` value if it is present.]*/
+    if (amqpMessage.correlation_id) {
+      if (((amqpMessage.correlation_id as any) instanceof Buffer) && (amqpMessage.correlation_id.length === 16)) {
+        msg.correlationId = rheaUuidToString(amqpMessage.correlation_id as any);
+      } else {
+        msg.correlationId = amqpMessage.correlation_id;
       }
```
BertKleewein: nit: could have a helper to deserialize a GUID since this code appears to repeat at least twice.

anthonyvercolano:
> **SRS_NODE_AMQP_CBS_16_001: [** The `constructor` shall instantiate a `SenderLink` object for the `$cbs` endpoint using a custom policy `{encoder: function(body) { return body;}}` which forces the amqp layer to send the token as an amqp value in the body. **]**

No.

---
In reply to: [411948200](https://github.com/Azure/azure-iot-sdk-node/pull/344#issuecomment-411948200) [](ancestors = 411948200)

---
Refers to: common/transport/amqp/devdoc/amqp_cbs_requirements.md:37 in 90bae3c. [](commit_id = 90bae3c1d5365f09ebb5dc9b8a82c73ef6ef3f47, deletion_comment = False)

common/transport/amqp/src/amqp_message.ts
BertKleewein: like maybe call it decodeUuid to mirror encodeUuid.  (or rename them to serialize/deserialize, but I like the symmetry here.)

---
In reply to: [209371927](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209371927) [](ancestors = 209371927)

common/transport/amqp/src/amqp_cbs.ts
anthonyvercolano: Well this is where the tribe is being told the story.  This of course has never been well specified.  I will restate with the far less chatty: For CBS, the message id and correlation id are encoded as string.

---
In reply to: [209348680](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209348680) [](ancestors = 209348680)

common/transport/amqp/src/amqp_cbs.ts
anthonyvercolano: Not anymore.

---
In reply to: [209348926](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209348926) [](ancestors = 209348926)

common/transport/amqp/src/amqp_message.ts
anthonyvercolano: Removed here and everywhere.

---
In reply to: [209349954](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209349954) [](ancestors = 209349954)

common/transport/amqp/src/amqp.ts
BertKleewein: I like that idea.  It _seems_ like we could maybe knock out 20 lines of code and a level of indirection that might not be necessary, but it doesn't seem to hurt as is and it gives us a place for breakpoints/additional debug code if necessary.  I think we should leave this as-is unless there's a sound reason to change it besides "I think we could be a little more clever and a little smaller here".  

---
In reply to: [209130807](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209130807) [](ancestors = 209130807)

BertKleewein: And I know how much pain Tony went through to make this right, so let's not make him revisit his pain :)

---
In reply to: [209374344](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209374344) [](ancestors = 209374344,209130807)

common/transport/amqp/src/receiver_link.ts
anthonyvercolano: changed on => from

---
In reply to: [209359158](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209359158) [](ancestors = 209359158)

common/transport/amqp/src/amqp.ts
```diff
@@ -145,44 +235,190 @@ export class Amqp {
...
+            // put the event handlers on the rhea connection object returned from the connect call and be assured
+            // that the listeners are in place BEFORE any possible events will be emitted on the connection.
+            //
+            this._rheaConnection = this._rheaContainer.connect(connectionParameters);
+            manageConnectionHandlers('on');
+          },
+          'connection_open': (context: EventContext) => {
+            debug('in the connection open handler for the connection state: ' + context.connection.name);
+            this._rheaConnection = context.connection;
+            let callback = this._connectionCallback;
```
BertKleewein: >let callback = this._connectionCallback; [](start = 12, length = 40)

Aaargh.  I've been stuck with this pattern before.  I hate saving callbacks like this.  It just seems so error prone.  So @#$&() frustrating.

common/transport/amqp/test/_amqp_cbs_test.js
```diff
@@ -27,60 +27,66 @@ describe('ClaimsBasedSecurityAgent', function() {
...
-      fakeAmqpClient.createSender = sinon.stub().resolves(fakeSender);
-      fakeAmqpClient.createReceiver = sinon.stub().resolves(fakeReceiver);
-
-      var cbs = new CBS(fakeAmqpClient);
+      var fakeRheaSession = new EventEmitter();
+      var cbs = new CBS(fakeRheaSession);
+      var fakeRheaLink = new EventEmitter();
+      var fakeContext = {receiver: fakeRheaLink};
+      fakeRheaSession.open_receiver = () => {};
+      sinon.stub(fakeRheaSession, 'open_receiver').callsFake(() => {process.nextTick( () => {fakeRheaLink.emit('receiver_open', fakeContext)});return fakeRheaLink;});
```
pierreca: >process.nextTick( () => {fakeRheaLink.emit('receiver_open', fakeContext)});return fakeRheaLink; [](start = 68, length = 95)

multiline please

common/transport/amqp/src/receiver_link.ts
anthonyvercolano: Yes for 2.

---
In reply to: [209360358](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209360358) [](ancestors = 209360358)

common/transport/amqp/test/_amqp_cbs_test.js
```diff
@@ -27,60 +27,66 @@ describe('ClaimsBasedSecurityAgent', function() {
...
-      fakeAmqpClient.createReceiver = sinon.stub().resolves(new EventEmitter()),
-      fakeAmqpClient.createSender = sinon.stub().resolves(new EventEmitter())
+      var fakeRheaSession = new EventEmitter();
+
+      var cbs = new CBS(fakeRheaSession);
+
+      var fakeRheaLink = new EventEmitter();
+      var fakeContext = {receiver: fakeRheaLink};
+      fakeRheaSession.open_receiver = () => {};
+      sinon.stub(fakeRheaSession, 'open_receiver').callsFake(() => {process.nextTick( () => {fakeRheaLink.emit('receiver_open', fakeContext)});return fakeRheaLink;});
```
pierreca: >{process.nextTick( () => {fakeRheaLink.emit('receiver_open', fakeContext)});return fakeRheaLink;}); [](start = 67, length = 99)

same thing - multiline please :)

common/transport/amqp/src/receiver_link.ts
anthonyvercolano: See reponse to 2 below.

---
In reply to: [209359414](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209359414) [](ancestors = 209359414)

common/transport/amqp/test/_amqp_cbs_test.js
```diff
@@ -27,60 +27,66 @@ describe('ClaimsBasedSecurityAgent', function() {
...
+      cbs._senderLink.attach = sinon.stub().callsArg(0);
+      cbs._senderLink.detach = sinon.stub().callsArg(0);
+      cbs._receiverLink.attach = sinon.stub().callsArgWith(0);
+      cbs._receiverLink.detach = sinon.stub().callsArgWith(0);
```
pierreca: a lot of this seems to repeat. I would suggest creating a stubLinks(cbs) helper function and just call this everywhere. then override the default stubs only when necessary.

common/transport/amqp/src/receiver_link.ts
anthonyvercolano: as the comment above says we are guaranteed that close will follow.  It's not at all clear to me how error information is propagated around.  Even so, to introduce another state would double code in these sections.

---
In reply to: [209360561](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209360561) [](ancestors = 209360561)

common/transport/amqp/src/amqp.ts
BertKleewein: I was hoping there was some way to pass information via the EventContext, but I don't see anything easy.  Aargh.  My comment is just expressing frusturation.  

---
In reply to: [209376307](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209376307) [](ancestors = 209376307)

common/transport/amqp/src/amqp.ts
```diff
@@ -145,44 +235,190 @@ export class Amqp {
...
+        },
+        connecting_session: {
+          _onEnter: (connectCallback, result) => {
+            this._sessionCallback = connectCallback;
+            this._sessionResult = result;
+            //
+            // According to the rhea maintainers, one can depend on that fact that no actual network activity
+            // will occur until the nextTick() after the call to create_session.  Because of that, one can
+            // put the event handlers on the rhea session object returned from the create_session call and be assured
+            // that the listeners are in place BEFORE any possible events will be emitted on the session.
```
BertKleewein: This is neat, but why not add the handlers before calling create_session?  You seem to be going to great lengths to justify that the placement of this call shouldn't cause any problems, but I don't know why you put it after and not before.

common/transport/amqp/test/_amqp_cbs_test.js
```diff
@@ -154,53 +196,91 @@ describe('ClaimsBasedSecurityAgent', function() {
...
+      });
+      var amqpSenderForceDetachSpy = sinon.spy(cbs._senderLink, 'forceDetach');
+      var amqpReceiverForceDetachSpy = sinon.spy(cbs._receiverLink, 'forceDetach');
+      cbs._senderLink.detach = sinon.stub();
+      cbs._receiverLink.detach = sinon.stub();
+
+      sinon.stub(fakeRheaReceiverLink, "on").callsFake(function (eventName) {
         if (eventName === 'message') {
           // sender is attached, receiver is attaching since we're registering for messages. good time to trigger a fake forceDetach
+          assert(fakeRheaSession.open_sender.calledOnce, 'open sender NOT called once');
```
pierreca: >'open sender NOT called once' [](start = 57, length = 29)

nit: this is _very_ obvious from the first half of the line. we could make it even more obvious by using a different test tool/syntax if you want. those messages feel redundant

device/transport/amqp/src/amqp_twin_client.ts
```diff
@@ -139,40 +141,40 @@ export class AmqpTwinClient extends EventEmitter {
...
+                } else {
+                  this._endpoint = endpoint.deviceTwinPath(credentials.deviceId);
+                }
+                /*Codes_SRS_NODE_DEVICE_AMQP_TWIN_06_006: [When a listener is added for the `response` event, and the `post` event is NOT already subscribed, upstream and downstream links are established via calls to `attachReceiverLink` and `attachSenderLink`.] */
+                /*Codes_SRS_NODE_DEVICE_AMQP_TWIN_06_012: [When a listener is added for the `post` event, and the `response` event is NOT already subscribed, upstream and downstream links are established via calls to `attachReceiverLink` and `attachSenderLine`.] */
+                /*Codes_SRS_NODE_DEVICE_AMQP_TWIN_16_036: [The same correlationId shall be used for both the sender and receiver links.]*/
+                const linkCorrelationId: string  = uuid.v4().toString();
+                this._client.attachSenderLink( this._endpoint, this._generateTwinLinkProperties(linkCorrelationId), (senderLinkError?: Error, senderTransportObject?: any): void => {
+                  if (senderLinkError) {
```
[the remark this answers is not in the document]
anthonyvercolano: No, but I was in the middle of debugging something a long time ago, and reversed to help me (lame i know).  Then I never changed it back.  I could change this, but it never seemed important.

---
In reply to: [209361869](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209361869) [](ancestors = 209361869)

common/transport/amqp/src/amqp_message.ts
anthonyvercolano: You are very smart.

---
In reply to: [209371716](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209371716) [](ancestors = 209371716)

common/transport/amqp/src/amqp.ts
```diff
@@ -126,15 +216,15 @@ export class Amqp {
               }
             } else if (this._disconnectHandler) {
               debug('calling upper layer disconnect handler');
-              this._disconnectHandler(err);
+              debug('error passed to disconnect handler is: ' + this._getErrorName(err || new errors.NotConnectedError('rhea: connection disconnected')));
+              this._disconnectHandler(err || new errors.NotConnectedError('rhea: connection disconnected'));
```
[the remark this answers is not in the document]
anthonyvercolano: This is invoking a higher level disconnection handler.  It always expects to see an error.  Even on a clean disconnection.

---
In reply to: [209374803](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209374803) [](ancestors = 209374803)

common/transport/amqp/src/amqp.ts
anthonyvercolano: Or less error prone.

---
In reply to: [209382324](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209382324) [](ancestors = 209382324,209376307)

common/transport/amqp/src/amqp.ts
anthonyvercolano: Because the create session is creating the emitter that the handlers are being placed on.

---
In reply to: [209383003](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209383003) [](ancestors = 209383003)

common/transport/amqp/src/amqp.ts
```diff
@@ -145,44 +235,190 @@ export class Amqp {
...
+            debug('In the session_close handler for the connected state with indicated error of: ' + this._getErrorName(this._indicatedSessionError));
+            debug('Transition to disconnecting');
+            let err = this._indicatedSessionError;
+            this._indicatedSessionError = undefined;
+            this._sessionCloseOccurred = true;
+            this._fsm.transition('disconnecting', null, err);
+          },
+          'connection_error': (context: EventContext) => {
+            debug('In the connection_error handler for the connected state with error of: ' + this._getErrorName(context.connection.error));
+            debug('Deferring transition until the connection_close occurs.');
```
BertKleewein: > debug('Deferring transition until the connection_close occurs.'); [](start = 11, length = 66)

vague uncomfortable feeling.  It looks like you're assuming that connection_close will maybe always happen after connection_error.  Is this a safe assumption?  Is the connection in some sort of "broken and waiting to close" state here?  Or is the state more vague than that?

common/transport/amqp/src/amqp.ts
```diff
@@ -145,44 +235,190 @@ export class Amqp {
...
+            this._indicatedConnectionError = context.connection.error;
+          },
+          'connection_close': (context: EventContext) => {
+            debug('In the connection_close handler for the connected state with indicated error of: ' + this._getErrorName(this._indicatedConnectionError));
+            debug('Transition to disconnecting');
+            let err = this._indicatedConnectionError;
```
BertKleewein: >et err = this._indicatedConnectionError; [](start = 13, length = 40)

Similar vague uncomfortable feeling.  Does this assume that connection_error always happens before connection_close if you're in this state?

device/transport/amqp/src/amqp_twin_client.ts
```diff
@@ -139,40 +141,40 @@ export class AmqpTwinClient extends EventEmitter {
...
+                      /* Codes_SRS_NODE_DEVICE_AMQP_TWIN_06_022: [If an error occurs on establishing the upstream or downstream link then the `error` event shall be emitted.] */
+                      this._fsm.transition('detached', senderLinkError, attachCallback);
                   } else {
-                    this._endpoint = endpoint.deviceTwinPath(credentials.deviceId);
+                    this._senderLink = senderTransportObject;
+                    this._senderLink.on('error', this._errorHandler);
+                    this._client.attachReceiverLink( this._endpoint, this._generateTwinLinkProperties(linkCorrelationId), (receiverLinkError?: Error, receiverTransportObject?: any): void => {
+                      if (receiverLinkError) {
+                        this._fsm.transition('detached', receiverLinkError, attachCallback);
```
BertKleewein: >this._fsm.transition('detached', receiverLinkError, attachCallback); [](start = 24, length = 68)

This probably predates your code, but if the receiver link fails to attach, it will leave the sender link in an attached state.  Does this cause any problems?

common/transport/amqp/test/_amqp_cbs_test.js
anthonyvercolano: Hmm, I was feeling guilty that all of my asserts DON'T have the textual comment.  I thought that they were mostly there to help the reporting when the assert fails.  I thought that otherwise the error report from the unit test run is a bit 'terse' and not particularly helpful when an actual failure occurs.

---
In reply to: [209383243](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209383243) [](ancestors = 209383243)

common/transport/amqp/src/amqp.ts
anthonyvercolano: No.  See comment above.

---
In reply to: [209389010](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209389010) [](ancestors = 209389010)

common/transport/amqp/src/amqp.ts
anthonyvercolano: Connection close ALWAYS will be called after the connection error.  Note though that connection_close is NOT always preceded by connection_error.  This statement is also true for all *_error and *_close events used throughout rhea.  Other than the general pain of the event handling, I don't consider this a deficiency.

This actually makes fairly reasonable sense if you examine how errors are presented as specified by the AMQP performative specification.  The way that errors show up is VIA a performative that causes the *_close events. 

---
In reply to: [209388831](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209388831) [](ancestors = 209388831)

common/transport/amqp/src/amqp.ts
pierreca: I understand that some of these might be unavoidable given how difficult it is to pass things around from one state to the next - nevertheless I think some of those booleans are there to split a state in 2 fake states - and I think the state machine could take care of that for us. 
we can leave this for later but this is something we will *need* to revisit as it is guaranteed to come back and haunt us, and to get much worse quick if we have bugs to fix later. 
Before we check this is let's at least take some time on a whiteboard to map the FSM states and how events influence transitions between them. the FSM really is a great way to "tame" the event-based nature of Rhea and I feel there's a missed opportunity here.

---
In reply to: [209328451](https://github.com/Azure/azure-iot-sdk-node/pull/344#discussion_r209328451) [](ancestors = 209328451,209129265,209128149)

pierreca:
Submitted https://github.com/anthonyvercolano/azure-iot-sdk-node/pull/1 to update this PR with the SASL changes. please merge it and update this PR before final merge.

PR #346, opened 2018-08-14
Edge E2E wrappers

BertKleewein:
As part of my Jenkins scripting, I'm moving the E2E wrappers into the node tree.  This makes the build easier and is more intuitive.  The scripts to build and use the test containers are still in the internals repo.


edge-e2e/run.cmd
```diff
@@ -0,0 +1,7 @@
+@REM Copyright (c) Microsoft. All rights reserved.
+@REM Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+@setlocal
+@echo off
+
+jshint "%~dp0\wrapper\nodejs-server-server\Service" && node "%~dp0\wrapper\nodejs-server-server\index.js"
```
pierreca: >jshint "%~dp0\wrapper\nodejs-server-server\Service" && [](start = 0, length = 54)

do you want jshint at this point? should that be a build step instead?

edge-e2e/wrapper/nodejs-server-server/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "description": "REST API definition for End-to-end testing of the Azure IoT SDKs.  All SDK APIs that are tested by our E2E tests need to be defined in this file.  This file takes some liberties with the API definitions.  In particular, response schemas are undefined, and error responses are also undefined.",
+  "main": "index.js",
+  "scripts": {
+    "prestart": "npm install",
+    "start": "node index.js"
+  },
+  "keywords": [
+    "swagger"
+  ],
+  "license": "Unlicense",
```
pierreca: >Unlicense [](start = 14, length = 9)

needs to be MIT

edge-e2e/wrapper/nodejs-server-server/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "private": true,
+  "dependencies": {
+    "async": "^2.6.1",
+    "azure-event-hubs": "0.0.8",
+    "azure-iot-common": "1.7.1",
+    "azure-iot-device": "1.7.1",
+    "azure-iot-device-amqp": "1.7.1",
+    "azure-iot-device-http": "1.7.1",
+    "azure-iot-device-mqtt": "1.7.1",
+    "azure-iothub": "1.7.1",
```
pierreca: needs to be caret references to latest SDK packages

edge-e2e/wrapper/nodejs-server-server/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "bluebird": "^3.3.0",
```
pierreca: >"bluebird": "^3.3.0", [](start = 4, length = 21)

since we control the environment we should be able to assume it'll have a global Promise object and not need external libraries

edge-e2e/wrapper/nodejs-server-server/package.json
BertKleewein: OK, but we don't do that anywhere else.  Is this something new that we're able to do because we have Lerna?

---
In reply to: [210696479](https://github.com/Azure/azure-iot-sdk-node/pull/346#discussion_r210696479) [](ancestors = 210696479)

edge-e2e/wrapper/nodejs-server-server/package.json
BertKleewein: You'll have to explain this a little better.  It's also worth mentioning that some of the code that relies on Bluebird is auto-generated, but I'm not sure how that affects your comment.

---
In reply to: [210696646](https://github.com/Azure/azure-iot-sdk-node/pull/346#discussion_r210696646) [](ancestors = 210696646)

edge-e2e/wrapper/nodejs-server-server/package.json
pierreca: no, the only reason we don't do it is because packages are linked together - as a user of our SDK i should be able to rely on using the latest version, that's why i'm suggesting this: because users will do that and therefore this may catch bugs.
if you end up linking everything with lerna then I don't think it matters


---
In reply to: [212857354](https://github.com/Azure/azure-iot-sdk-node/pull/346#discussion_r212857354) [](ancestors = 212857354,210696479)

edge-e2e/wrapper/nodejs-server-server/package.json
pierreca: if the autogenerated code relies on bluebird then bluebird is probabyl a dependency of swagger-tools or one of the other packages. the point is that we don't use bluebird in our SDK anymore: we rely on the fact that node comes with a global `Promise` object and that it's the one we should be using.

---
In reply to: [212857482](https://github.com/Azure/azure-iot-sdk-node/pull/346#discussion_r212857482) [](ancestors = 212857482,210696646)

edge-e2e/wrapper/nodejs-server-server/glue/deviceGlue.js
```diff
@@ -0,0 +1,127 @@
...
+ **/
+exports.deviceConnectionIdRoundtripMethodCallMethodNamePUT = function(connectionId,methodName,requestAndResponse) {
+  debug(`deviceConnectionIdRoundtripMethodCallMethodNamePUT called with ${connectionId}, ${methodName}`);
+  debug(JSON.stringify(requestAndResponse, null, 2));
+  return glueUtils.makePromise('deviceConnectionIdRoundtripMethodCallMethodNamePUT', function(callback) {
+    var client = objectCache.getObject(connectionId);
+    client.onDeviceMethod(methodName, function(request, response) {
+      debug(`function ${methodName} invoked from service`);
+      debug(JSON.stringify(request, null, 2));
+      if (JSON.stringify(request.payload) !== JSON.stringify(requestAndResponse.requestPayload.payload)) {
```
pierreca: >JSON.stringify(request.payload) [](start = 10, length = 31)

nit: don't have to go through serialization twice if you put this in a temp variable

PR #349, opened 2018-08-22
Update /lib -> /src

theangellmethod:
/lib folder name no longer exists.

# Description of the problem

Previous description had /lib as a folder but it is now /src in the structure.

# Description of the solution

Update /lib to /src


pierreca:
Thanks @theangellmethod !!

PR #354, opened 2018-08-30
Added amqp module support.

anthonyvercolano:
Added support for modules over amqp.  This entailed 3 pieces.

1) Support correctly sending down CA certs to tls.  Needed so edge modules can validate the edgehub.
2) Set the appropriate application properties on output messages.
3) Set the appropriate endpoint and annotations for input messages


common/core/src/endpoint.ts
```diff
@@ -47,7 +47,7 @@ export function moduleEventPath(deviceId: string, moduleId: string): string {
 }
 
 export function moduleMessagePath(deviceId: string, moduleId: string): string {
-  return modulePath(deviceId, moduleId) + '/messages/devicebound';
```
BertKleewein: >devicebound [](start = 53, length = 11)

you'll need to update _endpoint_test.js

device/transport/amqp/src/amqp.ts
```diff
@@ -283,7 +287,12 @@ export class Amqp extends EventEmitter implements DeviceTransport {
                   /*Codes_SRS_NODE_DEVICE_AMQP_13_002: [ The connect method shall set the CA cert on the options object when calling the underlying connection object's connect method if it was supplied. ]*/
                   if (this._options && this._options.ca) {
                     config.sslOptions = config.sslOptions || {};
-                    config.sslOptions.caFile = this._options.ca;
+                    /*Codes_SRS_NODE_DEVICE_AMQP_06_012: [The `connect` method shall first test if the `ca` property is the name of an already existent file.  If so, it will attempt to read that file as a pem into a string value and pass the string to config object `ca` property.  Otherwise, it is assumed to be a pem string.] */
+                    if (fs.existsSync(this._options.ca)) {
```
BertKleewein: >if (fs.existsSync(this._options.ca)) { [](start = 20, length = 38)

It seems dangerous to pass an arbitrary string to a filesystem function.  Would it make more sense to check if the string begins with the -----BEGIN... thing?

device/transport/amqp/src/amqp.ts
```diff
@@ -283,7 +287,12 @@ export class Amqp extends EventEmitter implements DeviceTransport {
...
+                      config.sslOptions.ca = fs.readFileSync(this._options.ca, 'utf8');
```
BertKleewein: >config.sslOptions.ca = fs.readFileSync(this._options.ca, 'utf8'); [](start = 22, length = 65)

And, for what it's worth, I'm not sure where this requirement came from?  I don't think MQTT or HTTP supports the readFile functionality?  I think they only accept strings.  (Of course, I've been wrong about stuff like this plenty of times before....)

device/transport/amqp/src/amqp.ts
pierreca: that's a good point. didn't think about that when I suggested this. there are security implications with even checking if the file exists too. Maybe we should run this by a security person.  

---
In reply to: [214414255](https://github.com/Azure/azure-iot-sdk-node/pull/354#discussion_r214414255) [](ancestors = 214414255)

device/transport/amqp/src/amqp.ts
pierreca: history of the amqp10 library

---
In reply to: [214414741](https://github.com/Azure/azure-iot-sdk-node/pull/354#discussion_r214414741) [](ancestors = 214414741)

device/transport/amqp/test/_amqp_test.js
```diff
@@ -384,7 +395,18 @@ describe('Amqp', function () {
...
           assert(fakeBaseClient.connect.called);
-          assert.strictEqual(fakeBaseClient.connect.firstCall.args[0].sslOptions.caFile, 'ca cert');
+          assert.strictEqual(fakeBaseClient.connect.firstCall.args[0].sslOptions.ca, 'ca cert');
+          testCallback();
+        });
+      });
+
+      /*Tests_SRS_NODE_DEVICE_AMQP_06_012: [The `connect` method shall first test if the `ca` property is the name of an already existent file.  If so, it will attempt to read that file as a pem into a string value and pass the string to config object `ca` property.  Otherwise, it is assumed to be a pem string.] */
+      it('sets CA cert with contents of file if provided', function (testCallback) {
+        transport.setOptions({ ca: '.\\device\\transport\\amqp\\test\\_fake_pem.pem' });
```
pierreca: >.\\device\\transport\\amqp\\test\\_fake_pem.pem [](start = 36, length = 47)

please don't check in the file: rather create and delete it as part of the test setup/teardown.

device/transport/amqp/src/amqp.ts
anthonyvercolano: Yup, it's the way things were with the previous incantation.  This really seems like the least bad solution.

---
In reply to: [214525410](https://github.com/Azure/azure-iot-sdk-node/pull/354#discussion_r214525410) [](ancestors = 214525410,214414741)

device/transport/amqp/src/amqp.ts
anthonyvercolano: I actually thought about this quite a bit.  My assumption was that the node file api would be sufficiently careful about this.  I was actually made even more comfortable about this when I looked to see whether the api would throw.  It doesn't.

---
In reply to: [214525406](https://github.com/Azure/azure-iot-sdk-node/pull/354#discussion_r214525406) [](ancestors = 214525406,214414255)

device/transport/amqp/test/_amqp_test.js
anthonyvercolano: I thought you might say this.  Please let me know the objection.

---
In reply to: [214525511](https://github.com/Azure/azure-iot-sdk-node/pull/354#discussion_r214525511) [](ancestors = 214525511)

common/core/src/endpoint.ts
anthonyvercolano: Doh!  How did I miss that?  I thought I ran the tests too!

---
In reply to: [214413537](https://github.com/Azure/azure-iot-sdk-node/pull/354#discussion_r214413537) [](ancestors = 214413537)

device/transport/amqp/test/_amqp_test.js
anthonyvercolano: Done

---
In reply to: [214986317](https://github.com/Azure/azure-iot-sdk-node/pull/354#discussion_r214986317) [](ancestors = 214986317,214525511)

PR #363, opened 2018-09-07
Enable e2e tests for modules over MQTT

pierreca:
# Description of the problem
Some modules features were only tested over AMQP - adding MQTT for completeness.

# Description of the solution
Found a couple of things in the tests:
1. the way the originated device/module was discriminated was using an AMQP-specific property, switched it to message annotations
2. in MQTT the subscription for twin events sometimes took longer and tests were hitting a race condition. introduced a small timeout to counteract this (not a great solution so we'll see how stable this is over time: it's a test framework issue though, not an SDK API issue)


pierreca:
replaced by #364

PR #367, opened 2018-09-12
Sample code for a downstream device that connects to an IoT Edge device

mrohera:
# Checklist
- [X ] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-node/blob/master/.github/CONTRIBUTING.md).
- [ X] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).

# Description of the problem

Providing a sample to make it easier for users to connect downstream devices to an Azure IoT Edge device,

# Description of the solution

Added a sample that highlights how a developer should put together a connection string for the Edge and install the relevant CA certificate required to authenticate the Azure IoT Edge server certificate. The sample was based off simple_sample_device.js with the required pieces to connect to an Edge.


device/samples/edge_downstream_device.js
```diff
@@ -0,0 +1,91 @@
...
+      client.removeAllListeners();
+      client.open(connectCallback);
+    });
+  }
+};
+
+// Provide the Azure IoT device client via setOptions with the X509
+// Edge root CA certificate that was used to setup the Edge runtime
+var options = {
+  ca : fs.readFileSync(edge_ca_cert_path, 'utf-8').toString(),
```
anthonyvercolano: >.toString() [](start = 50, length = 11)

The string SHOULD be unnecessary.  I believe I don't use it in my code.

anthonyvercolano:
> # Samples for the Azure IoT device SDK for Node.js

Azure has a sample repo.  Let me find out if we need to move this now.

---
Refers to: device/samples/readme.md:1 in bc0b1fe. [](commit_id = bc0b1fe9a2df7de37741d4f08aa91f6fc2809b67, deletion_comment = False)

anthonyvercolano:
Why is the canary failing?

device/samples/edge_downstream_device.js
mrohera: I looked at this for reference: https://github.com/Azure/azure-iot-sdk-node/blob/c1ae87bf7c769a7193d3add07213a8b57c579540/device/samples/simple_sample_device_x509.js#L66

I can try without and re-test.

mrohera:
I am not sure why the canary is failing. When I click the details link I get a 401 unauthorized so I am in the dark here.

device/samples/edge_downstream_device.js
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+
+var fs = require('fs');
+var Protocol = require('azure-iot-device-mqtt').Mqtt;
+// Uncomment one of these transports and then change it in fromConnectionString to test other transports
+// var Protocol = require('azure-iot-device-amqp').AmqpWs;
+// var Protocol = require('azure-iot-device-http').Http;
```
BertKleewein: I would remove the Http, AmqpWs, and MqttWs lines.  Even though they're commented out, they're invalid for an edge downstream device so we shouldn't imply that they could be used.

BertKleewein:
I’ve added my signoff with one minor comment.  Please let us know when you’re ready for us to accept it.

device/samples/edge_downstream_device.js
mrohera: Makes sense, will remove and update this PR.

PR #369, opened 2018-09-18
update package-lock.json to fix problem in canary verification on vsts

BertKleewein:
Hesh was seeing a verification failure in a step with the name "canary" in it, but it had nothing to do with canary.  The package-lock.json for edge E2E tests was missing jshint, so I'm hoping that a package-lock.json update will fix this.


pierreca:
merged with #370

PR #382, opened 2018-09-30
test for failure of client service.

anthonyvercolano:
Just using for test run.

The check will take several hours to run.  Kill it if you have to.

DO NOT MERGE.


anthonyvercolano:
Not suitable.

PR #386, opened 2018-10-02
(fix) Pin @types/node to 8.9.4 to work around a breakage.

anthonyvercolano:
An incompatibility occurred in the most currently distributed node types module. 
We force those modules where this showed up version 8.9.4.
This gets them back on their feet.


anthonyvercolano:
No longer needed.

PR #420, opened 2018-11-08
Fix TypeScript compile-time issues

kpajdzik:
# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/device/core/src/blob_upload/blob_upload_client.ts
+++ b/device/core/src/blob_upload/blob_upload_client.ts
@@ -4,10 +4,10 @@
 'use strict';
 
 import { Stream } from 'stream';
-import { AuthenticationProvider, errorCallbackToPromise, ErrorCallback } from 'azure-iot-common';
+import { AuthenticationProvider, errorCallbackToPromise, ErrorCallback, TripleValueCallback, Callback } from 'azure-iot-common';
 
 import { BlobUploadResult } from './blob_upload_result';
-import { BlobUploader as DefaultBlobUploader } from './blob_uploader';
+import { BlobUploader as DefaultBlobUploader, BlobResponse } from './blob_uploader';
 import { FileUploadApi as DefaultFileUploadApi } from './file_upload_api';
 
 import * as errors from './blob_upload_errors';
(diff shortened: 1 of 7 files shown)


device/core/src/blob_upload/blob_uploader.ts
```diff
@@ -38,7 +38,7 @@ export class BlobUploader implements BlobUploaderInterface {
     }
   }
 
-  uploadToBlob(blobInfo: UploadParams, stream: Stream, streamLength: number, done: TripleValueCallback<any, { statusCode: number, body: string }>): Promise<any> | void {
+  uploadToBlob(blobInfo: UploadParams, stream: Stream, streamLength: number, done?: TripleValueCallback<any, { statusCode: number, body: string }>): Promise<{ body: any, result: { statusCode: number, body: string } }> | void {
```
pierreca: >{ body: any, result: { statusCode: number, body: string } } [](start = 157, length = 59)

looks like this would benefit from having its own type

device/core/src/blob_upload/blob_upload_client.ts
```diff
@@ -27,15 +27,15 @@ export interface UploadParams {
...
+  getBlobSharedAccessSignature(blobName: string, done?: Callback<UploadParams>): void;
   notifyUploadComplete(correlationId: string, uploadResult: BlobUploadResult, done: (err?: Error) => void): void;
 }
 
 /**
  * @private
  */
 export interface BlobUploader {
-  uploadToBlob(uploadParams: UploadParams, stream: Stream, streamLength: number, done: (err: Error, body?: any, result?: { statusCode: number, body: string }) => void): void;
+  uploadToBlob(uploadParams: UploadParams, stream: Stream, streamLength: number, done?: TripleValueCallback<any, { statusCode: number, body: string }>): void;
```
pierreca: >{ statusCode: number, body: string } [](start = 113, length = 36)

needs to be BlobResponse now right?

pierreca:
@kpajdzik 

As-is, the PR doesn't build. please run `lerna run build` to see the issues.
```
src/provisioningserviceclient.ts(495,9): error TS2349: Cannot invoke an expression whose type lacks a call signature. Type 'TripleValueCallback<IndividualEnrollment, any> | TripleValueCallback<DeviceRegistrationState, any...' has no compatible call signatures.
src/provisioningserviceclient.ts(497,9): error TS2349: Cannot invoke an expression whose type lacks a call signature. Type 'TripleValueCallback<IndividualEnrollment, any> | TripleValueCallback<DeviceRegistrationState, any...' has no compatible call signatures.
```

Also, please rebase and squash the commits :)

kpajdzik:
I think I didn’t catch it because I was updating TS locally and it seems like a bug in TS (https://github.com/Microsoft/TypeScript/issues/4612). Thanks for catching. Fixed and squashed.

anthonyvercolano:
Merged via pull of this branch.

Issue #416, opened 2018-11-01
TypeError: Callback has to be a Function (azure-iothub@ > 1.7.4)


beneuto:
Code working fine using azure-iothub v1.7.4 but breaks in v1.7.5 and above with the error: **TypeError: Callback has to be a Function.**

Similar problem with azure-iot-device.

Minimal code example:

```
var connectionString = 'HostName=myhub.azure-devices.net;SharedAccessKeyName=test;SharedAccessKey=xxxxxxxxxxxxxxxxxxxxxxxxx';

var device =  {   deviceId: "xyz"   };
 
var registry = iothub.Registry.fromConnectionString(connectionString);

registry.get(device.deviceId, function(err, deviceInfo, res) {    });
```

anthonyvercolano:
I'm at a bit of a loss.  I've taken the above supplied code.  Used my own device id and hub connection string and ran it on a 1.7.5 installation.  It worked fine.

beneuto:
It may be because I am running through node-red function node?   

Node-RED version: v0.19.4
Node.js  version: v8.11.1

anthonyvercolano:
Likely.  Will be discussing this with others who might have ideas.

anthonyvercolano:
Which node-red module are you using?

beneuto:
The Function node.

anthonyvercolano:
Thank you.  Repro'd and investigating further.

anthonyvercolano:
Should be available in the next release.

az-iot-builder-01:
@beneuto, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #421, opened 2018-11-10
(fix) Issue #416. Use typeof rather than instanceof Function.

anthonyvercolano:
instanceof Function was broken in the context of node-red

Changed the code to do typeof object === 'function'


pierreca:
nit: unnecessary parenthesis everywhere undermine readability.

PR #430, opened 2018-11-16
(feature) add retry on transient errors.

anthonyvercolano:
(feature) provisioning - persevere with transient errors.

### DIFF
--- a/provisioning/transport/amqp/src/amqp.ts
+++ b/provisioning/transport/amqp/src/amqp.ts
@@ -12,8 +12,9 @@ const debug = dbg('azure-iot-provisioning-device-amqp:Amqp');
 
 import { X509, errors } from 'azure-iot-common';
 import { ProvisioningTransportOptions, X509ProvisioningTransport, TpmProvisioningTransport, RegistrationRequest, RegistrationResult, ProvisioningDeviceConstants } from 'azure-iot-provisioning-device';
-import { Amqp as Base, SenderLink, ReceiverLink, AmqpMessage, AmqpBaseTransportConfig } from 'azure-iot-amqp-base';
+import { Amqp as Base, SenderLink, ReceiverLink, AmqpMessage, AmqpBaseTransportConfig, translateError, AmqpTransportError } from 'azure-iot-amqp-base';
 import { GetSasTokenCallback, SaslTpm } from './sasl_tpm';
+
 
 /**
  * @private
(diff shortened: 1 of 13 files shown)


provisioning/transport/amqp/src/amqp.ts
```diff
@@ -312,8 +319,14 @@ export class Amqp extends EventEmitter implements X509ProvisioningTransport, Tpm
...
-                /*Codes_SRS_NODE_PROVISIONING_AMQP_16_021: [The `queryOperationStatus` method shall call its callback with an error if the transport fails to send the request message.]*/
-                callback(err);
+                const translatedError = translateError('registration failure', err);
+                if ((translatedError instanceof errors.InternalServerError) || ((translatedError as AmqpTransportError) instanceof errors.ThrottlingError)) {
+                  /*Codes_SRS_NODE_PROVISIONING_AMQP_06_002: [ The `queryOperationStatus` disposition is an `instanceof` common errors InternalServerError or ThrottlingError, the result.status shall be set to 'assigning', and the callback invoked with *no* error. ] */
+                  callback(null, {status: 'assigning'});
+                } else {
+                  /*Codes_SRS_NODE_PROVISIONING_AMQP_16_021: [The `queryOperationStatus` method shall call its callback with an error if the transport fails to send the request message.]*/
+                  callback(err);
+                  }
```
pierreca: >   } [](start = 15, length = 4)

nit: indentation

provisioning/transport/http/src/http.ts
```diff
@@ -226,6 +230,10 @@ export class Http extends EventEmitter implements X509ProvisioningTransport, Tpm
         if (response.statusCode < 300) {
           /* Codes_SRS_NODE_PROVISIONING_HTTP_18_039: [ If the Http request succeeds, `queryOperationStatus` shall call `callback`, passing a `null` error along with the `result` and `response` objects. ] */
           callback(null, result, response, this._config.pollingInterval);
+        } else if (response.statusCode >= 429) {
+          /*Codes_SRS_NODE_PROVISIONING_HTTP_06_007: [ If the `queryOperationStatus` response contains a status code >= 429, the result.status value will be set with `assigning` and the callback will be invoked with *no* error object. ] */
+          result.status = 'assigning';
+          callback(null, result, response, this._config.pollingInterval);
```
pierreca: >result [](start = 25, length = 6)

what is this in that case? i believe it's an error. the polling state machine is going to emit an operationStatus event with that error instead of a result. is this OK?

pierreca: looks like no one is using those events from the polling state machine.... maybe we should stop emitting them if they have no value.

---
In reply to: [234755384](https://github.com/Azure/azure-iot-sdk-node/pull/430#discussion_r234755384) [](ancestors = 234755384)

provisioning/transport/amqp/src/amqp.ts
anthonyvercolano: Fixed.

---
In reply to: [234754603](https://github.com/Azure/azure-iot-sdk-node/pull/430#discussion_r234754603) [](ancestors = 234754603)

anthonyvercolano:
Being reworked elsewhere.

PR #434, opened 2018-11-22
Add overloads to functions returning void or Promise

Loghorn:
# Checklist

# Description of the problem
A lot of function in the SDK can return either void or a Promise<...>, depending on the passing of a callback or not. The problem is that TypeScript cannot know what the return type will be without proper overload function declaration.

# Description of the solution
I've added the needed overloads.
For example a function like `function foo(cbk?: Callback<string>): Promise<string> | void;` is now declared as:
```
function foo(cbk: Callback<string>): void;
function foo(): Promise<string>;
function foo(cbk?: Callback<string>): Promise<string> | void {
...
}
``

### CODE REVIEW

**Comment by Loghorn** [comment]
Unfortunately for people in Italy today is a regular working day 😃 
The overloads allow to write code like
```
const iotClient = Client.fromConnectionString('....');
iotClient.open().then(...).catch(...)
```
without TypeScript complaining that `Property 'then' does not exists on type void | Promise<Connected>`.
It also avoid the needs of casts:
```
// with the overloads
const twin = await iotClient.getTwin();
// without
const twin = (await iotClient.getTwin()) as Twin;
```

**Comment by pierreca** [comment]
@Loghorn thanks for your explanations and what looks like another excellent PR :) we've had a few issues related to typings changes since we added the Promisification layer so we'd like to take the "safe" if not "boring" approach of re-enabling some steps in our build process that would prevent API breaks on this side (all our e2e tests are still pure javascript and don't protect us against this). It'll be a day or two before we can merge this in, but it definitely looks like a desirable change - and as soon as we're done with these builds steps, we'll merge this (unless it breaks the build ;) and we'll release new packages.

**Comment by Loghorn** [comment]
@pierreca Thank you. Take the time you need to make all the changes and let me know if something is failing so I can fix it.

File: provisioning/service/src/provisioningserviceclient.ts [hunk 1]
```diff
@@ -60,6 +62,10 @@ export class ProvisioningServiceClient {
    * @param {function}          [deleteCallback] Invoked upon completion of the operation.
    * @returns {Promise<void> | void} Promise if no callback function was passed, void otherwise.
    */
+  public deleteIndividualEnrollment(enrollmentOrId: string | IndividualEnrollment, etag: string, deleteCallback: ErrorCallback): void;
+  public deleteIndividualEnrollment(enrollmentOrId: string | IndividualEnrollment, deleteCallback: ErrorCallback): void;
+  public deleteIndividualEnrollment(enrollmentOrId: string | IndividualEnrollment, etagOr: string): Promise<void>;
```
pierreca: >etagOr [](start = 83, length = 6)

found a typo :) I'll take care of it in the carry-over PR.

provisioning/service/src/provisioningserviceclient.ts
```diff
@@ -232,6 +252,10 @@ export class ProvisioningServiceClient {
    * @param {function}          [deleteCallback]      Invoked upon completion of the operation.
    * @returns {Promise<void> | void} Promise if no callback function was passed, void otherwise.
    */
+  public deleteDeviceRegistrationState(idOrRegistrationState: string | DeviceRegistrationState, etag: string, deleteCallback: ErrorCallback): void;
+  public deleteDeviceRegistrationState(idOrRegistrationState: string | DeviceRegistrationState, deleteCallback: ErrorCallback): void;
+  public deleteDeviceRegistrationState(idOrRegistrationState: string | DeviceRegistrationState, etagOrCallback: string): Promise<void>;
```
pierreca: >etagOrCallback [](start = 96, length = 14)

probably just etag (I'll take care of it)

service/src/job_client.ts
```diff
@@ -253,6 +263,12 @@ export class JobClient {
    * @throws {ReferenceError}   If one or more of the jobId, queryCondition or patch arguments are falsy.
    * @throws {TypeError}        If the callback is not the last parameter
    */
+  scheduleTwinUpdate(jobId: string | number, queryCondition: string, patch: any, jobStartTime: Date, maxExecutionTimeInSeconds: number, done: TripleValueCallback<any, any>): void;
+  scheduleTwinUpdate(jobId: string | number, queryCondition: string, patch: any, jobStartTime: Date, done: TripleValueCallback<any, any>): void;
+  scheduleTwinUpdate(jobId: string | number, queryCondition: string, patch: any, done: TripleValueCallback<any, any>): void;
+  scheduleTwinUpdate(jobId: string | number, queryCondition: string, patch: any, jobStartTime: Date, maxExecutionTimeInSeconds?: number | TripleValueCallback<any, any>): Promise<JobStatusResponse>;
```
pierreca: > | TripleValueCallback<any, any> [](start = 135, length = 32)

not necessary (I'll take care of it)

common/core/src/promise_utils.ts
```diff
@@ -69,6 +69,8 @@ export type HttpResponseCallback<TResult> = TripleValueCallback<TResult, any>;
  * const param = 42;
  * callbackToPromise((_callback) => foo(param, _callback)).then(result => { console.log(result); }, error => { console.error(error); });
  */
+export function callbackToPromise<TResult>(callBackOperation: (callback: Callback<TResult>) => void, userCallback: Callback<TResult>): void;
+export function callbackToPromise<TResult>(callBackOperation: (callback: Callback<TResult>) => void): Promise<TResult>;
```
kpajdzik: Not sure if here and in `errorCallbackToPromise` the overload is necessary. Those methods should not be called by customers and the callback parameter should determine if callback or Promise should be used. Does lack of overloads here break anything?

pierreca:
released with 1.8.2

Loghorn:
Thank you and sorry for the few typos 😃

Issue #352, opened 2018-08-24
Messages not delivered and no exceptions thrown (azure-iot-device@1.4.0)


wmmihaa:
# Context
- **OS and version used:** Ubuntu.Core 16.04
- **Node.js version:** 8.11
- **npm version:** 6.04

# Description of the issue:
To begin with, we have many IoT Hubs and we've only seen this issue on one of our devices on one hub. This is a new project, so we only have one device running on this particular hub. So this issue might be related to the IoT Hub and not the device SDK.

All devices are running **azure-iot-common@1.5.0,azure-iot-device@1.4.0 and azure-iot-device-mqtt@1.4.0** with **NoRetry** policy. At the particular device with this issue messages gets transmitted without any exceptions but no messages seems to get received by the IoT Hub. 

Question 1: Has you seen this behavor before? 
Question 2: As we're considering moving to 1.7, what changes has been made (I know there is a page somewhere, but I couldn't find it ;) ) 

![image](https://user-images.githubusercontent.com/3214367/44572363-e3405a80-a783-11e8-937a-225b4af9bbfc.png)

pierreca:
@wmmihaa version 1.7.x introduces modules and edge features.
The only case that I could think of for this is:
- on linux
- the device connects successfully
- then the connection is dropped (linux doesn't let the SDK know that it happened)
- the messages keep getting enqueued and you never get a callback
- disconnect is never fired because we are never notified by the socket

we are missing a timeout logic that is separate from the retry logic, especially in the case of the mqtt library where things just get enqueued. I'm guessing the callback passed to sendEvent never gets called?

wmmihaa:
That is the weird thing, I do get the sendEvent callback

pierreca:
In that case it seems as though the SDK "thinks" the message was put on the wire.
You could use the azure portal to open a support ticket - the support engineers can then grab your hub name and device id, a timestamp for the issue, and look at the SLA for your hub during that time and see if anything happened on our side.

bragma:
Hi @pierreca,
we have exactly the same problem reported above:

- on linux
- the device connects successfully
- after some time, something happens (maybe the client is disconnected)
- the messages keep getting enqueued and you never get a callback
- disconnect is never fired because we are never notified by the socket

If I try to send a direct method to the device, I get the correct reply so the device is connected / has reconnected, but still no messages are sent with sendEvent. What's the suggested fix for this? Shouldn't heartbeats solve this?

[...]

PR #451, opened 2018-12-18
Updated rhea dependancy to 0.3.6 related to #352

anthonyvercolano:
Carry over #450 to put through the gate.

# Checklist
- [X] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).

*Messages not delivered and no exceptions thrown* https://github.com/Azure/azure-iot-sdk-node/issues/352
# Description of the problem
In summary, there is a *size* field which is initially set to 2048 when the device or twin is created. The value of the field is decreasing as events are sent and increased upon acknowledgement from the server. -If device twin is used, this value is not increased, and when it reaches 0 the device fails to submit messages (but is still connected, so no errors are raised). If device twin is not used, there are not issues.

# Description of the solution
An update has been provided from the rhea repo (https://github.com/amqp/rhea/issues/171), and updates to the dependencies have been made.


pierreca:
Windows build will fail if the package-lock.json file isn't updated as well

PR #456, opened 2019-01-04
Feature/client certificate engine

sNAtttu:
# Checklist

# Description of the problem

In our iot edge solution we are using TPM based encryption. We have developed a native plugin which is used as an engine that provides the client certificate information.

The device code is using the AMQP protocol to communicate with the IoTHub and I'm passing the ssl engine in the x509Options for the authentication provider.

Example below:

```
authenticationProvider = authenticationProvider.fromX509Options(
  deviceId,
  iotHubHostName,
  {
    clientCertEngine: sslEngine
  }
);
```

I've found out that I'm successfully authenticated against the IoT Hub and everything else seems to be functioning normally but when I try to upload files through to the IoT Hub I get following error message:

**ErrorCode: IotHubUnauthorizedAccess**

In the device code we are using following function to do the file upload:

```
deviceClient.uploadToBlob(toFile, stream, size, cb);
```

This is the function which can be found from the device_client.ts. After some more research I noticed that when we call this function (uploadToBlob) it tries to generate SAS token in the SDK which ends up being unauthorized.

# Description of the solution

By recording network traffic we noticed that the unauthorized message was because of missing certificate information. This discovery led us to dig a little bit deeper into the codebase and it seemed that in the **azure-iot-sdk/common/transport/http.ts** the code checks that if the options passed are indeed x509Options. There wasn't check for the clientCertEngine. Maybe this was because the types for the node was using the major version 7 but the clientCertEngine TLS option was introduced in Node 9.3.0, so there wasn't such option available. I chose the minimum node version from the history of [tls.createSecureContext](https://nodejs.org/api/tls.html#tls_tls_createsecurecontext_options).

My solution for this was to update the node types from ^7.0.12 to ^9.3.0 in the dev dependencies, add the check for the clientCertEngine and then propagate the option to the x509Options.

Examples below:

**Changes to the isX509Options**

**Before:**
```
  isX509Options(options: any): boolean {
    return !!options && typeof(options) === 'object' &&
      (options.cert || options.key || options.passphrase ||
        options.certFile || options.keyFile);
  }
```
**After**
```
  isX509Options(options: any): boolean {
    return !!options && typeof(options) === 'object' &&
      (options.cert || options.key || options.passphrase ||
        options.certFile || options.keyFile || options.clientCertEngine);
  }
```

**Changes to the build request function**

**Before**
```
    /*Codes_SRS_NODE_HTTP_16_001: [If `options` has x509 properties, the certificate, key and passphrase in the structure shall be used to authenticate the connection.]*/
    if (x509Options) {
      httpOptions.cert = (x509Options as X509).cert;
      httpOptions.key = (x509Options as X509).key;
      httpOptions.passphrase = (x509Options as X509).passphrase;
    }
```

**After**
```
    /*Codes_SRS_NODE_HTTP_16_001: [If `options` has x509 properties, the certificate, key and passphrase in the structure shall be used to authenticate the connection.]*/
    if (x509Options) {
      httpOptions.cert = (x509Options as X509).cert;
      httpOptions.key = (x509Options as X509).key;
      httpOptions.passphrase = (x509Options as X509).passphrase;
      httpOptions.clientCertEngine = (x509Options as X509).clientCertEngine;
    }
```

These changes were enough for us to get successful response from the SAS Token request and after that the file upload functionality was working as expected.

### DIFF
--- a/common/core/src/authorization.ts
+++ b/common/core/src/authorization.ts
@@ -99,8 +99,12 @@ export interface X509 {
    * Passphrase used to decode the key associated with the X509 certificate.
    */
   passphrase?: string;
+
+  /**
+   * Name of an OpenSSL engine which can provide the client certificate.
+   */
   clientCertEngine?: string;
-  secureProtocol?: string;
+
   /**
    * @private
    */
--- a/common/core/src/authorization.ts
+++ b/common/core/src/authorization.ts
@@ -99,7 +99,8 @@ export interface X509 {
    * Passphrase used to decode the key associated with the X509 certificate.
    */
   passphrase?: string;
-
+  clientCertEngine?: string;
+  secureProtocol?: string;
   /**
    * @private
    */
--- a/common/transport/http/package-lock.json
+++ b/common/transport/http/package-lock.json
@@ -1,29 +1,35 @@
 {
-  "requires": true,
+  "name": "azure-iot-http-base",
+  "version": "1.8.3",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
     "@sinonjs/formatio": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/@sinonjs/formatio/-/formatio-2.0.0.tgz",
       "integrity": "sha1-hNt+nrVTHfGKjF4L+25EnlXmVLI=",
+      "dev": true,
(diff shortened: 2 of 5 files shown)


anthonyvercolano:
Merged by carry over #460

PR #486, opened 2019-02-09
Stabilize method_disconnect end-to-end tests

pierreca:
- add logs
- add timeouts
- manually open device client

I've had pretty good success with this (passed 90 times in a row, with logging enabled and disabled)

### DIFF
--- a/e2etests/test/method_disconnect.js
+++ b/e2etests/test/method_disconnect.js
@@ -147,14 +147,14 @@ protocolAndTermination.forEach( function (testConfiguration) {
     var methodResult = 200;
 
     var setMethodHandler = function(methodName, testPayload) {
-      debug('---------- new method test ------------');
+      debug('Setting up new method handler for: ' + methodName);
       // setup device to handle the method call
       deviceClient.onDeviceMethod(methodName, function(request, response) {
         // validate request
         assert.isNotNull(request);
         assert.strictEqual(request.methodName, methodName);
         assert.deepEqual(request.payload, testPayload);
-        debug('device: received method request');
+        debug('device: received method request: ' + request.methodName);
         debug(JSON.stringify(request, null, 2));
         // validate response
         assert.isNotNull(response);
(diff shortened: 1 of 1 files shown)


e2etests/test/method_disconnect.js
```diff
@@ -171,30 +171,27 @@ protocolAndTermination.forEach( function (testConfiguration) {
...
-                  sendMethodCallback(err);
-                });
-      }, 1000);
+          provisionedDevice.deviceId,
+          methodParams,
+          function(err, result) {
+            if(!err) {
+              debug('got method results');
+              debug(JSON.stringify(result, null, 2));
+              assert.strictEqual(result.status, methodResult);
```
anthonyvercolano: >result.status, methodResult [](start = 33, length = 27)

I would like to add a 'policy' to e2e tests as much as possible.  I want to see the device id printed out with every failure.  This way we can examine the logs and have the device id as one of the search parameters.

e2etests/test/method_disconnect.js
```diff
@@ -211,20 +208,35 @@ protocolAndTermination.forEach( function (testConfiguration) {
...
+      deviceClient.open(function (err) {
+        if (err) {
+          debug('error connecting the device client: ' + err.toString());
           testCallback(err);
+        } else {
+          debug('setting up method handler for ' + methodName1);
+          setMethodHandler(methodName1, firstPayload);
+          debug('waiting 2 seconds before invoking ' + methodName1);
+          setTimeout(function () {
+            sendMethodCall(serviceClient, methodName1, firstPayload, function(err) {
```
anthonyvercolano: >sendMethodCall(serviceClient, methodName1, firstPayload, function(err) { [](start = 12, length = 72)

You might want to use the rendezvous helper on these.   I've certainly seen the response from the injection come in quite late.  You might think of this as having three paths here.  The injection path, the client method handler path, and the service sending the method path.  Just a thought.

e2etests/test/method_disconnect.js
pierreca: that's a good idea. let's start doing that.

e2etests/test/method_disconnect.js
pierreca: I'll add that if we see failures.

PR #499, opened 2019-02-28
(fix) retry connection shutdown in the case of no response within tim…

anthonyvercolano:
Lacks an e2e test put I'd like to get a quick review of what I have.

### DIFF
--- a/common/transport/amqp/src/amqp.ts
+++ b/common/transport/amqp/src/amqp.ts
@@ -578,33 +578,67 @@ export class Amqp {
               }
             };
 
-            detachLink(this._cbs, () => {
-              let remainingLinks = [];
-              for (let senderEndpoint in this._senders) {
-                if (this._senders.hasOwnProperty(senderEndpoint)) {
-                  remainingLinks.push(this._senders[senderEndpoint]);
-                  delete this._senders[senderEndpoint];
-                }
+            let remainingLinks = [];
+            for (let senderEndpoint in this._senders) {
+              if (this._senders.hasOwnProperty(senderEndpoint)) {
+                remainingLinks.push(this._senders[senderEndpoint]);
+                delete this._senders[senderEndpoint];
               }
+            }
 
-              for (let receiverEndpoint in this._receivers) {
-                if (this._receivers.hasOwnProperty(receiverEndpoint)) {
-                  remainingLinks.push(this._receivers[receiverEndpoint]);
-                    delete this._receivers[receiverEndpoint];
-                }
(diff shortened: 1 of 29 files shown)


common/transport/amqp/test/_amqp_test.js
```diff
@@ -610,6 +610,59 @@ describe('Amqp', function () {
...
+        amqp._receivers.fake_receiver_endpoint = fakeReceiver;
+        amqp.disconnect(function(err) {
+          assert(fakeSender.detach.calledOnce);
+          assert(fakeSender.forceDetach.calledOnce);
+          assert(fakeReceiver.detach.calledOnce);
+          assert(fakeReceiver.forceDetach.calledOnce);
+          if (err) testCallback(err);
+          else {
+            testCallback();
+          }
```
pierreca: nit: linter would insist on having curly braces everywhere

common/transport/amqp/src/amqp.ts
```diff
@@ -578,33 +578,63 @@ export class Amqp {
...
+            //
+            // If the initial function invocation actually doesn't hang due to lack of communication, it's
+            // last action will be to clear the timeout and consequently the timeout invocation call to the
+            // rundown code will NOT occur.
+            //
+            let rerunWithForceTimer: any = undefined;
+            const rundownConnection = (force: boolean) => {
+              if (force) {
+                this._disconnectionOccurred = true;
+              }
```
pierreca: nit: this._disconnectionOccurred = force :)

common/transport/amqp/test/_amqp_test.js
anthonyvercolano: I love git blame.

Will fix.

---
In reply to: [261444583](https://github.com/Azure/azure-iot-sdk-node/pull/499#discussion_r261444583) [](ancestors = 261444583)

common/transport/amqp/src/amqp.ts
anthonyvercolano: Actually, after I left this afternoon I recalled about why I wrote it this way.

this._disconnectionOccurred might already be true when we enter this function.

Thus if we simply set it to the value of force we, on  the first call (the call NOT in the setTimer), will stomp a true to a false.

Not what we want.

In a sense this is why the force parameter name is good.

I will add a comment to call this out.

---
In reply to: [261444657](https://github.com/Azure/azure-iot-sdk-node/pull/499#discussion_r261444657) [](ancestors = 261444657)

PR #505, opened 2019-03-06
update github creds, yaml file, and github ref for horton gate

BertKleewein:
Normalizing the Horton gate YAML across SDKs.


BertKleewein:
This is the pipeline definition which links to the horton repo and runs the horton tests when triggered.  This is only a build definition.  For now, the results will be informational.  This job won't gate commits until the SDK team decides that it should.

PR #510, opened 2019-03-12
Import early set of scripts to improve automation of releases

fratster:
# Description of the problem
Internal release process is cumbersome.

# Description of the solution
Add bump-version code and dependent yaml file to support creation of release builds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,6 +68,9 @@ ipch/
 *.vsp
 *.vspx
 
+# Visual Studio user folder
+.vs
+
 # Guidance Automation Toolkit
 *.gpState
 
--- a/scripts/git-branch.ps1
+++ b/scripts/git-branch.ps1
@@ -1,0 +1,136 @@
+
+# create a new release branch with annoated tag and push to git origin
+
+[CmdletBinding()]
+param
+(
+    [Parameter(Mandatory = $true)]
+    [string]$Name,
+    
+    [Parameter(Mandatory = $true)]
+    [string]$Email
+)
+
+function TagExists($TagName, [switch]$Remote) {
+    
+    # look for the name in local refs/tags
+
+    if ($Remote -eq $true) {
+
+        foreach ($tag in $(git ls-remote --tags)) {
+
+            $name = Split-Path $tag -Leaf
+    
+            if ($name -eq $TagName) {
+                return $true
+            }
+        }
+    }
+    else {
+
+        foreach ($tag in $(git tag)) {
+
+            if ($tag -eq $TagName) {
+                return $true
+            }
+        }
+    }
+
+    return $false
+}
+
+function ExitOrContinue {
+
+    # stop only if prior git command failed
+
+    Write-Host "Last exit code is ($LASTEXITCODE)"
+
+    if ($LASTEXITCODE -ne 0) {
+        exit $LASTEXITCODE  # terminate
+    }
+}
+
+function CurrentBranch() {
+
+    # find current branch name from git output
+
+    foreach ($line in $(git branch)) {
+
+        # git branch output marks the current branch with a leading astrisk
+
+        if ($line.StartsWith('*')) {
+            $line = $line.Replace('* ', '')
+            return $line
+        }
+    }
+
+    throw "Current branch not found"
+}
+
+
+
+
+# start:
+
+$date = Get-Date -UFormat "%Y-%m-%d"
+$branch = "release_$date"
+$sources = $env:BUILD_SOURCESDIRECTORY
+
+Write-Output "Set current directory to '$sources'"
+Set-Location $sources
+
+git config user.email $Email
+git config user.name $Name
+
+$message = "Bump package versions for release $date"
+
+Write-Output "Create new branch '$($branch)'"
+
+git checkout -b $branch
+ExitOrContinue  # exit if we can't create the branch
+
+$lastCommitId = git rev-list --tags --max-count=1
+$lastTag = git describe $lastCommitId  # By default git describe only shows annotated tags.
+
+Write-Output "Print commits from $lastTag..$branch"
+
+git --no-pager log $lastTag..$branch --name-status
+
+Write-Output "Add changes to commit"
+
+git add .
+ExitOrContinue  # exit if we can't add the changes
+
+git commit -m $message
+ExitOrContinue  # exit if we can't commit the changes
+
+if ($(CurrentBranch) -eq 'master') {
+
+    Write-Output "Can't continue: current branch is 'master' but should be '$($branch)'."
+    exit -1
+}
+
+if ($(TagExists $date) -eq $true) {
+
+    Write-Host "Delete local tag '$date'"
+    git tag --delete $date
+}
+
+if ($(TagExists $date -Remote) -eq $true) {
+
+    Write-Host "Delete remote tag '$date'"
+    git push --delete origin $date
+}
+
+Write-Output "Create new tag '$($date)'"
+
+git tag -a $date -m "SDK release on $date"
+ExitOrContinue  # exit if we can't create a new tag
+    
+Write-Output "Push changes to origin"
+
+git push --tags -u origin $branch  # push version changes
+    
+Write-Host "Git push finished with exit code ($($LASTEXITCODE))"
+
+exit $LASTEXITCODE
--- a/vsts/node-release.yaml
+++ b/vsts/node-release.yaml
@@ -1,0 +1,55 @@
+name: Release_$(BuildID)_$(Date:yyyyMMdd)$(Rev:.r)
+resources:
+- repo: self
+  clean: true
+
+phases:
+
+- phase: Phase_1
+  displayName: Create Release Branch
+
+  condition: succeededOrFailed()
+  queue:
+    name: Hosted Ubuntu 1604
+
+  steps:
+
+  - checkout: self  # self represents the repo where the initial Pipelines YAML file was found
+    persistCredentials: true  # set to 'true' to leave the OAuth token in the Git config after the initial fetch
+    clean: true
+
+  - task: NodeTool@0
+    inputs:
+      versionSpec: '10.x'
+
+  - script: |
+       npm install --global node-gyp
+       npm install --global lerna
+    displayName: 'Install Dependencies'
+
+  - task: Npm@1
+    inputs:
+      command: 'install'
+      workingDir: $(Build.SourcesDirectory)/scripts/version
+
+  - powershell: $(Build.SourcesDirectory)/scripts/bump-versions.ps1
+    displayName: Bump Versions
+
+  - task: PowerShell@2
+    inputs:
+      filePath: $(Build.SourcesDirectory)/scripts/git-branch.ps1
+      arguments: -Name '$(GitHubName)' -Email '$(GitHubEmail)'
+    displayName: Create Release Branch
+
+  - script: |
+        lerna bootstrap
+        lerna run build
+    displayName: 'Bootstrap & Build'
+
+  - powershell: $(Build.SourcesDirectory)/scripts/npm-pack.ps1
+    displayName: NPM Pack
+
+  - task: PublishBuildArtifacts@1
+    inputs:
+      pathtoPublish: $(Agent.RootDirectory)/tgz
+      artifactName: drop
--- a/scripts/bump-versions.ps1
+++ b/scripts/bump-versions.ps1
@@ -1,0 +1,16 @@
+
+# run the node.js app that bumps file versions
+
+$sources = $env:BUILD_SOURCESDIRECTORY
+$configFile = "$($sources)/scripts/version/config.json"
+$app = "$($sources)/scripts/version/app.js"
+
+if ($(Test-Path -Path $configFile -PathType Leaf) -ne $true) {
+    throw "File not found: $($configFile)"
+}
+
+if ($(Test-Path -Path $app -PathType Leaf) -ne $true) {
+    throw "File not found: $($app)"
+}
+
+node $app $configFile $sources
--- a/scripts/npm-pack.ps1
+++ b/scripts/npm-pack.ps1
@@ -1,0 +1,57 @@
+
+# call npm pack on each sdk folder path to produce releasable outputs
+
+$folders = @(
+    'common/core',
+    'common/transport/amqp',
+    'common/transport/http',
+    'common/transport/mqtt',
+    'device/core',
+    'device/transport/amqp',
+    'device/transport/http',
+    'device/transport/mqtt',
+    'device/node-red',
+    'service',
+    'provisioning/service',
+    'provisioning/device',
+    'provisioning/transport/amqp',
+    'provisioning/transport/mqtt',
+    'provisioning/transport/http',
+    'security/x509',
+    'security/tpm',
+    'security/symmetric')
+
+
+$sources = $env:BUILD_SOURCESDIRECTORY
+$agent = $env:AGENT_ROOTDIRECTORY
+
+$tgzFolder = "$agent\tgz"  # tgzFolder folder path 
+New-Item $tgzFolder -ItemType Directory  # make new tgzFolder folder
+
+foreach ($folder in $folders) {
+
+    $path = Join-Path $sources -ChildPath $folder  # archive file package root folder
+    
+    if ($(Test-Path $path -PathType Container) -ne $true) {
+        throw "Folder not found: $($path)"
+    }
+    
+    $packageFile = Join-Path $path -ChildPath 'package.json'  # archive file manifest
+        
+    if ($(Test-Path $packageFile -PathType Leaf) -ne $true) {
+        throw "File not found: $($packageFile)"
+    }
+    
+    Set-Location $path  # set working folder to project root\
+    
+    $archiveName = npm pack  # create archive file
+    $archivePath = Join-Path $path -ChildPath $archiveName
+    
+    if ($(Test-Path $archivePath -PathType Leaf) -ne $true) {
+        throw "File not found: $($archivePath)"
+    }
+        
+    $destinationPath = Join-Path $tgzFolder -ChildPath $archiveName  # path to .tgz collection folder
+    
+    Move-Item $archivePath -Destination $destinationPath  # move .tgz to collection folder for publish
+}
--- a/scripts/version/app.js
+++ b/scripts/version/app.js
@@ -1,0 +1,265 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+
+var fs = require('fs');
+var path = require('path');
+var util = require('util');
+var async = require('async');
+var xpath = require('xpath');
+var DOMParser = require('xmldom').DOMParser;
+var XMLSerializer = require('xmldom').XMLSerializer;
+var colors = require('colors');
+
+if(process.argv.length === 4){
+    var configJSON = require(process.argv[2]);
+    var versions = require(configJSON.versionsFile);
+    var inputs = require(configJSON.inputsFile);
+    var basePath = process.argv[3];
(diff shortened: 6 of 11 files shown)


.gitignore
```diff
@@ -240,3 +243,6 @@ lerna-debug.log
 
 # mocha
 **/test-results.*.xml
+
+# release scripts
+.scripts/
```
pierreca: I don't think this folder should be ignored by git - after all, you're checking stuff in right now that goes into it.

.scripts/version/inputs.js
```diff
@@ -0,0 +1,119 @@
...
+        "filePath": "common/transport/mqtt/package.json"
+    },
+    {
+        "taskType": "regexReplaceTask",
+        "filePath": "device/readme.md",
+        "search": "(http\\:\\/\\/azure.github.io\\/azure-iot-sdks\\/node\\/api_reference\\/azure-iot-device\\/).*(\\/index.html)",
+        "replaceString": function(versions) {
+            return '$1' + versions.device + '$2';
+        }
+    },
```
pierreca: I think we can remove this it seems to point to an old docs folder.

.scripts/version/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "name": "bump_version",
+  "version": "1.0.0",
+  "description": "Version stamps all SDKs.",
+  "main": "app.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/Azure/azure-iot-sdks"
```
pierreca: >"https://github.com/Azure/azure-iot-sdks" [](start = 11, length = 41)

this URIs are wrong now

.scripts/version/versions.json
```diff
@@ -0,0 +1,22 @@
...
+  "azure-iot-provisioning-device": "1.5.3",
+  "azure-iot-provisioning-device-http": "1.5.3",
+  "azure-iot-provisioning-device-amqp": "1.5.3",
+  "azure-iot-provisioning-device-mqtt": "1.5.3",
+  "azure-iot-provisioning-service": "1.5.3",
+  "azure-iot-provisioning-device-e2e": "1.5.3",
+  "azure-iot-security-x509": "1.5.3",
+  "azure-iot-security-tpm": "1.5.3",
+  "azure-iot-security-symmetric-key": "1.5.3"
+}
```
pierreca: I don't think these are the latest versions

vsts/node-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+    name: Hosted
+
+  steps:
+
+  - checkout: self  # self represents the repo where the initial Pipelines YAML file was found
+    persistCredentials: true  # set to 'true' to leave the OAuth token in the Git config after the initial fetch
+
+  - task: NodeTool@0
+    displayName: 'Use Node 6.x'
+    enabled: false
```
pierreca: >false [](start = 13, length = 5)

if the task is disabled it should be removed. if it's enable, it should probably be node 8 or 10, 6 is kindof old now.

vsts/node-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+  clean: true
+
+phases:
+
+- phase: Phase_1
+  displayName: Create Release Branch
+
+  condition: succeededOrFailed()
+  queue:
+    name: Hosted
```
pierreca: >Hosted [](start = 10, length = 6)

I'd be curious to test if this could run on Linux (using powershell core?) npm & node operations are notably faster on linux.

vsts/node-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+          throw "File not found: $($configFile)"
+      }
+
+      if ($(Test-Path -Path $app -PathType Leaf) -ne $true) {
+          throw "File not found: $($app)"
+      }
+
+      node $app $configFile "$(Build.SourcesDirectory)"
+      
+    displayName: Bump Version
```
pierreca: >Version [](start = 22, length = 7)

nit: versions (with an 's') there are multiple packages and associated versions

vsts/node-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+    enabled: true
+
+  - script: dir /b /s
```
pierreca: >- script: dir /b /s [](start = 2, length = 19)

is this for debuggability? do we need it now?

vsts/node-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+      git tag -a $($releaseTag) -m "SDK release on $($releaseTag)"
+      git push --tags -u origin $($releaseBranch)
+      git checkout master
+      git branch -d $releaseBranch
+      
+    displayName: Create Release Branch
+    enabled: true
+
+  - powershell: |
+       runas.exe /savecred /user:administrator
```
pierreca: >runas.exe /savecred /user:administrator [](start = 7, length = 39)

i'm surprised that we need to be admins - is that required?

vsts/node-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+
+  - script: dir /b /s
+
+  - powershell: |
+      $releaseTag = Get-Date -UFormat "%Y-%m-%d"
+      $releaseBranch = "release_$($releaseTag)"
+
+      Set-Location "$(Agent.BuildDirectory)\s"
+
+      git config user.email $(GitHubEmail)
```
pierreca: >GitHubEmail [](start = 30, length = 11)

just double checking: those should be stored in a keyvault and linked from a variable group.

vsts/node-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+      git config user.name $(GitHubName) 
+      git checkout -b $releaseBranch
+      git add .
+      git commit -m "Bump package versions for release $($releaseTag)"
+      git tag -d $($releaseTag)
```
pierreca: >git tag -d $($releaseTag) [](start = 6, length = 25)

this deletes the tag locally but not on the remote. if this tag has already been created and pushed, it only delays the failure. 
if we want to build in robustness and the possibility to overwrite a previous release tentative, we need to:
1. detect if the tag already exists
2. delete it locally
3. detect if the tag has been pushed
4. if yes, delete it from the remote (using git push <remote> --delete <tag name>

vsts/node-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+      git tag -a $($releaseTag) -m "SDK release on $($releaseTag)"
+      git push --tags -u origin $($releaseBranch)
+      git checkout master
+      git branch -d $releaseBranch
```
pierreca: >git branch -d $releaseBranch [](start = 6, length = 28)

this shouldn't be necessary

vsts/node-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+      
+    displayName: Create Release Branch
+    enabled: true
+
+  - powershell: |
+       runas.exe /savecred /user:administrator
+       npm install --global node-gyp
+       npm install --global lerna
+    displayName: 'Install Dependencies'
+    enabled: false
```
pierreca: >false [](start = 13, length = 5)

Task is disabled?

vsts/node-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+
+      git config user.email $(GitHubEmail)
+      git config user.name $(GitHubName) 
+      git checkout -b $releaseBranch
+      git add .
+      git commit -m "Bump package versions for release $($releaseTag)"
+      git tag -d $($releaseTag)
+      git tag -a $($releaseTag) -m "SDK release on $($releaseTag)"
+      git push --tags -u origin $($releaseBranch)
+      git checkout master
```
pierreca: >git checkout master [](start = 6, length = 19)

careful if you checkout master now, all the files that have been modified and pushed to the release branch are now back to their "non-bumped" state. which will lead you to generate packages for the old version instead of the new one.

vsts/node-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+       npm install --global node-gyp
+       npm install --global lerna
+    displayName: 'Install Dependencies'
+    enabled: false
+
+  - script: |
+       call lerna bootstrap
+       call lerna run build
+    displayName: 'Bootstrap & Build'
+    enabled: false
```
pierreca: >false [](start = 13, length = 5)

probably needs to be enabled.

vsts/node-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+
+  - task: PublishTestResults@2
+    displayName: 'Publish Test Results | Mocha'
+    inputs:
+      testResultsFiles: 'e2etests/test-results.*.xml'
+      mergeTestResults: true
+      testRunTitle: 'E2E Tests - Windows'
+    condition: succeededOrFailed()
+    enabled: false
```
pierreca: I keep going back & forth on this - given we're going to build the release branch with a build triggered by the PR, I don't think we need to run the tests here. change my mind?

vsts/node-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+        'security/tpm',
+        'security/symmetric')
+
+      $repoParent = $(Get-Item "$(Agent.BuildDirectory)").Parent
+
+      $artifacts = '$(Agent.RootDirectory)\.tar'  # artifacts folder path 
+
+      if ($(Test-Path $artifacts -PathType Container) -eq $true) {
+          Remove-Item -Path $artifacts -Recurse  # delete old artifacts folder
+      }
```
pierreca: why should we do that if its' a new build? is this folder shared from one build to the next?

vsts/node-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+
+      New-Item -ItemType Directory -Path $artifacts  # make new artifacts folder
+
+      foreach ($folder in $folders) {
+
+        $path = Join-Path -Path "$(Agent.BuildDirectory)\s" -ChildPath $folder  # archive file project root folder
```
pierreca: >project [](start = 95, length = 7)

I think you mean "package" not "project"?

vsts/node-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+        
+        if ($(Test-Path -Path $path -PathType Container) -ne $true) {
+            throw "Folder not found: $($path)"
+        }
+        
+        $packageFile = Join-Path -Path $path -ChildPath 'package.json'  # archive file manifest
+            
+        if ($(Test-Path -Path $packageFile -PathType Leaf) -ne $true) {
+            throw "Package not found: $($packageFile)"
```
pierreca: >Package [](start = 19, length = 7)

nit: file

vsts/node-release.yaml
```diff
@@ -0,0 +1,185 @@
...
+        }
+        
+        Set-Location -Path $path  # set working folder to project root
+        
+        $json = Get-Content -Raw -Path $packageFile | ConvertFrom-Json  # read package.json to find output file name
+        $archiveName = "$($json.name)-$($json.version).tgz"
+        $archivePath = Join-Path -Path $path -ChildPath $archiveName
```
pierreca: note for later: we should be able to extract the archive name from the output of the NPM command rather than recreate it and hope they never change how they name packages.

.scripts/version/package.json
fratster: It's not clear how this URL is wrong.  What should we put instead?

---
In reply to: [265818897](https://github.com/Azure/azure-iot-sdk-node/pull/510#discussion_r265818897) [](ancestors = 265818897)

vsts/node-release.yaml
fratster: Yes they are

---
In reply to: [265819724](https://github.com/Azure/azure-iot-sdk-node/pull/510#discussion_r265819724) [](ancestors = 265819724)

vsts/node-release.yaml
fratster: I added logic to detect if the tag name already exists in local refs/tags and if not, the new tag is created

---
In reply to: [265820080](https://github.com/Azure/azure-iot-sdk-node/pull/510#discussion_r265820080) [](ancestors = 265820080)

.gitignore
fratster: I added a more specific folder name to exclude the node_modules dependencies folder created by npm install in order to run bump-version.

---
In reply to: [265818667](https://github.com/Azure/azure-iot-sdk-node/pull/510#discussion_r265818667) [](ancestors = 265818667)

.scripts/version/package.json
fratster: We can amend at a later date per discussion

---
In reply to: [266112503](https://github.com/Azure/azure-iot-sdk-node/pull/510#discussion_r266112503) [](ancestors = 266112503,265818897)

.gitignore
```diff
@@ -240,3 +243,6 @@ lerna-debug.log
 
 # mocha
 **/test-results.*.xml
+
+# release script dependencies
+.scripts/version/node_modules/
```
pierreca: should not be necessary given line 193

.scripts/git-branch.ps1
```diff
@@ -0,0 +1,74 @@
...
+    git branch -d
+
+    throw "Failed to find last tagged commit ($LASTEXITCODE)"
+}
+
+Write-Output "Print commits from $lastTag..$branch"
+git log $lastTag..$branch --name-status
+
+git add .
+git commit -m  $message
```
pierreca: i'm not seeing any error checking in here. maybe let's not block to check this in but i'm worried about error piling on if one of the first commands fail and then we check in everything.

.scripts/git-branch.ps1
```diff
@@ -0,0 +1,74 @@
...
+
+git checkout -b $branch
+
+$lastCommitId = git rev-list --tags --max-count=1
+$lastTag = git describe $lastCommitId  # By default git describe only shows annotated tags.
+
+if ($LASTEXITCODE -ne 0) {
+
+    git checkout master
+    git branch -d
```
pierreca: >git branch -d [](start = 4, length = 13)

IIRC git branch takes a branch name as a parameter. if it can function with the current branch, then you're deleting master. in both cases i think this is dangerous. you probably want to delete the branch first (by specifying $branch), THEN checkout master

.scripts/git-branch.ps1
```diff
@@ -0,0 +1,74 @@
...
+git commit -m  $message
+
+if ($(TagExists -TagName $date) -eq $false) {
+
+    Write-Output "Create new tag '$($date)'"
+    git tag -a $date -m "SDK release on $date"
+}
+else {
+    
+    Write-Output "Tag '$($date)' already exists"
```
pierreca: >Write-Output "Tag '$($date)' already exists" [](start = 4, length = 44)

at that point is' probably best to fail and exit.

.scripts/git-branch.ps1
```diff
@@ -0,0 +1,74 @@
...
+}
+
+Write-Output "Push changes to origin"
+
+git push --tags -u origin $branch  # push version changes
```
pierreca: this is the critical step. maybe add 2 steps:
1. verify that $branch is not "master"
2. verify that there have been no previous errors.

.scripts/git-branch.ps1
fratster: ok but...why do we need to fail here?  I could be missing something but in testing this case works as expected

---
In reply to: [267187667](https://github.com/Azure/azure-iot-sdk-node/pull/510#discussion_r267187667) [](ancestors = 267187667)

fratster: added comment to explain this, pls correct me if I'm mistaken

---
In reply to: [267483066](https://github.com/Azure/azure-iot-sdk-node/pull/510#discussion_r267483066) [](ancestors = 267483066,267187667)

scripts/git-branch.ps1
```diff
@@ -0,0 +1,133 @@
...
+            if ($tag -eq $TagName) {
+                return $true
+            }
+        }
+    }
+
+    return $false
+}
+
+function ExitOrContinue($BranchName) {
```
pierreca: >$BranchName [](start = 24, length = 11)

this parameter isn't used.

scripts/git-branch.ps1
```diff
@@ -0,0 +1,133 @@
...
+function CurrentBranch() {
+
+    # find current branch name from git output
+
+    git branch | % { $_.Trim() } | % {
+        
+        if ($_.StartsWith('*')) {
+            $name = $_.Replace('*', '').Trim()
+            return $name
+        }
```
pierreca: could you add a comment explaining why we may expect * there and why it's bad?

vsts/node-release.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+  - task: Npm@1
+    inputs:
+      command: 'install'
+      workingDir: $(Build.SourcesDirectory)/scripts/version
+
+  - powershell: $(Build.SourcesDirectory)/scripts/bump-versions.ps1
+    displayName: Bump Versions
+
+  - task: PowerShell@2
```
pierreca: >- task: PowerShell@2 [](start = 2, length = 20)

you're missing a mandatory step before this.
after bumping versions you still need to run `lerna bootstrap` and `lerna run build` to generate the javascript.

my comment about removing other (disabled) steps were about `lerna run ci` and `lerna run e2e`. these are the ones we don't need. we still need `lerna bootstrap` and `lerna run build`

vsts/node-release.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+  - script: |
+       npm install --global node-gyp
+       npm install --global lerna
+    displayName: 'Install Dependencies'
+    
+  - script: |
+        lerna bootstrap
+        lerna run build
+    displayName: 'Bootstrap & Build'
```
pierreca: i believe you need to run this after bump version otherwise the package-lock files aren't going to be good (that's why my comment was strategically placed after bump version but before npm pack). In other words:

1. install dependencies
2. bump version
3. lerna bootstrap
4. lerna run build
5. npm pack
6. publish artifacts

scripts/git-branch.ps1
```diff
@@ -0,0 +1,135 @@
+
+# create a new release branch with annoated tag and push to git origin
+
+[CmdletBinding()]
+param
+(
+    [Parameter(Mandatory = $false)]
```
pierreca: >$false [](start = 27, length = 6)

aren't those parameters mandatory?

scripts/git-branch.ps1
```diff
@@ -0,0 +1,135 @@
...
+
+    # find current branch name from git output
+
+    foreach ($line in $(git branch)) {
+
+        # git branch output marks the current branch with a leading astrisk
+
+        if ($line.StartsWith('*')) {
+            return $line
+        }
```
pierreca: nit: I kinda liked having the trimming of the spaces and * here, makes the code below cleaner. no big deal but if it's easy to bring back... i think it's worth it.

scripts/git-branch.ps1
```diff
@@ -0,0 +1,135 @@
...
+
+Write-Output "Add changes to commit"
+
+git add .
+ExitOrContinue  # exit if we can't add the changes
+
+git commit -m $message
+ExitOrContinue  # exit if we can't commit the changes
+
+if ($(CurrentBranch) -eq '* master') {
```
pierreca: >*  [](start = 26, length = 2)

to echo the previous comment i think that's a little weird. I can live with it but if we can bring back the code that removed the * and trims the spaces i think it'll look better.

pierreca:
Looking good :) please rebase on the latest version of master and squash the commits, then let's use this for our next release :)

fratster:
I have one commit ready to merge

---
In reply to: [475830942](https://github.com/Azure/azure-iot-sdk-node/pull/510#issuecomment-475830942) [](ancestors = 475830942)

anthonyvercolano:
Carried over.

PR #515, opened 2019-03-21
style(comment): add missing invokeDeviceMethod object key params

louistiti:
# Description of the problem

The comments does not match the real interface properties.

# Description of the solution

I added the `connectTimeoutInSeconds` property in the comment and renamed the `timeoutInSeconds` comment property to `responseTimeoutInSeconds`.


anthonyvercolano:
Thank you very much for providing this fix.  We very much appreciate it.  Will carry over your pull request to push it through our gate.

anthonyvercolano:
Merged with the carry over.

PR #519, opened 2019-03-27
Adjust the error test for the bad server.

anthonyvercolano:
mqtt web sockets on ubuntu vm is returning a bare error.  Adjusting  the error check on the disconnect to be more forgiving.


anthonyvercolano:
Do over because of the node-release.

PR #521, opened 2019-03-29
update package.json versions to be consistent in order to support `le…

BertKleewein:
…rna bootstrap --hoist`

This is an optimization for Horton, but it can benefit everyone.  `lerna boostrap --hoist` adds an extra layer of savings by installing NPM modules into a common directory and linking all packages to the common NPM modules.  This way, for instance, the mocha package that almost everyone uses is only at one spot on your drive (azure-iot-sdk-node/node_modules), and all the packages symlink to this copy.

On a dev machine (full lerna bootstrap) this takes the NPM overhead from about 2.5 GB down to about 1/2 GB.  
On a horton image (partial lerna bootstrap), it takes the overhead from about 1.2 GB down to 200 MB or so.  THIS IS HUGE because EVERY horton build for every language pulls down a node image for the friend module.  

Some of the size advantage is lost if packages don't all use the same version of a sub-package because the sub-packages might need a private version of the sub-package instead of the shared version.  This is meh, but no a big deal.

However, the big reason for this change is that `lerna bootstrap --hoist` can break if different packages use different typescript compilers because an older typescript compiler with a newer types/foo package (from the common folder) can fail the `npm build` step.

The one change in http.ts was caught by a newer types/node package which identified that the Date() constructor does not accept a Date object as a parameter.


common/core/src/retry_operation.ts
```diff
@@ -37,6 +37,9 @@ export class RetryOperation {
       this._retryCount++;
       /*Codes_SRS_NODE_COMMON_RETRY_OPERATION_16_001: [The `operation` function should be called at every retry.]*/
       operation((err, result, response) => {
+        if (this._currentTimeout) {
+            this._currentTimeout = 0;
+        }
```
anthonyvercolano: Why is this here?

provisioning/transport/amqp/package.json
```diff
@@ -11,21 +11,21 @@
     "azure-iot-amqp-base": "2.2.4",
     "azure-iot-common": "1.9.4",
     "azure-iot-provisioning-device": "1.5.4",
-    "bluebird": "^3.5.0",
+    "bluebird": "^3.5.2",
```
anthonyvercolano: >bluebird": "^3.5.2" [](start = 5, length = 19)

Why is bluebird still here?

ts-e2e/package.json
```diff
@@ -14,22 +14,22 @@
...
-    "@types/node": "^7.0.8",
-    "@types/uuid": "^2.0.29",
+    "@types/node": "^9.3.0",
+    "@types/uuid": "^3.4.3",
     "azure-iot-common": "^1.9.0-preview.0",
     "azure-iot-device": "^1.9.0-preview.0",
     "azure-iot-device-amqp": "^1.9.0-preview.0",
     "azure-iot-device-http": "^1.9.0-preview.0",
     "azure-iot-device-mqtt": "^1.9.0-preview.0",
     "azure-iothub": "^1.9.0-preview.0",
```
anthonyvercolano: Really out of sync.

device/ts-samples/package.json
```diff
@@ -19,13 +19,13 @@
   },
   "homepage": "https://github.com/azure/azure-iot-sdk-node#readme",
   "dependencies": {
-    "@types/node": "^7.0.14",
+    "@types/node": "^9.3.0",
     "azure-iot-common": "^1.1.6",
```
anthonyvercolano: >"azure-iot-common": "^1.1.6" [](start = 4, length = 28)

Out of sync with others - 1.9.4

anthonyvercolano:
>     "azure-iot-common": "^1.5.0",

Out of sync with others - 1.9.4

---
Refers to: longhaultests/package.json:31 in 78d8e6b. [](commit_id = 78d8e6be6b878085b2aabf8f8cbf683dc6ec0c11, deletion_comment = False)

anthonyvercolano:
>     "azure-iot-common": "^1.1.5",

Out of sync with others - 1.9.4

---
Refers to: network_e2e/tests/package.json:18 in 78d8e6b. [](commit_id = 78d8e6be6b878085b2aabf8f8cbf683dc6ec0c11, deletion_comment = False)

device/ts-samples/package.json
```diff
@@ -19,13 +19,13 @@
...
     "azure-iot-device": "^1.1.10",
     "azure-iot-device-amqp": "^1.1.10",
     "azure-iot-device-http": "^1.1.10",
     "azure-iot-device-mqtt": "^1.1.10",
```
anthonyvercolano: Out of sync.

network_e2e/tests/package.json
```diff
@@ -22,6 +22,6 @@
     "azure-iot-device-mqtt": "^1.1.7",
```
anthonyvercolano: Out of sync

build/tools/package.json
```diff
@@ -24,4 +24,4 @@
     "pem": "^1.14.2",
     "yargs": "^4.7.1"
```
anthonyvercolano: >"yargs": "^4.7.1 [](start = 4, length = 16)

Out of sync.

network_e2e/tests/package.json
```diff
@@ -22,6 +22,6 @@
...
     "network": "^0.3.2",
     "yargs": "^7.0.2",
```
anthonyvercolano: >"yargs": "^7.0.2" [](start = 4, length = 17)

Out of sync

ts-e2e/package.json
```diff
@@ -14,22 +14,22 @@
...
-    "chai": "^3.5.0",
+    "chai": "^4.2.0",
     "debug": "^4.1.1",
     "mocha": "^5.2.0",
-    "tslint": "^5.11.0",
+    "tslint": "^5.9.1",
     "typescript": "^3.1.6",
```
anthonyvercolano: >typescript": "^3.1.6" [](start = 5, length = 21)

Danger - 3.1.5

longhaultests/package.json
```diff
@@ -35,6 +35,6 @@
     "azure-iot-device-mqtt": "^1.4.0",
     "azure-iothub": "^1.3.0",
```
anthonyvercolano: Also out of date.  Sure, it's ok.

common/core/src/retry_operation.ts
BertKleewein: I was getting a linter error saying that this._currentTimeout is set but never read and this was easier than figuring out the suppression syntax.  
As it turns out, it wasn't a linter error, it used to be a linter error, but now it's a compiler error.  I needed to set noUnusedLocals to false in tsconfig.json to suppress this error.

---
In reply to: [270512971](https://github.com/Azure/azure-iot-sdk-node/pull/521#discussion_r270512971) [](ancestors = 270512971)

ts-e2e/package.json
BertKleewein: this is pierre's project and he's admitted that it's out of sync.  I'm not sure if I should be touching it at all.

---
In reply to: [270528223](https://github.com/Azure/azure-iot-sdk-node/pull/521#discussion_r270528223) [](ancestors = 270528223)

ts-e2e/package.json
BertKleewein: again, pierre's project.  I assume he wanted typescript 3.x for a reason.

---
In reply to: [270532046](https://github.com/Azure/azure-iot-sdk-node/pull/521#discussion_r270532046) [](ancestors = 270532046)

BertKleewein:
>     "azure-iot-common": "^1.1.5",

stale.  we should remove these -- I'll submit a PR.

---
In reply to: [478103051](https://github.com/Azure/azure-iot-sdk-node/pull/521#issuecomment-478103051) [](ancestors = 478103051)

---
Refers to: network_e2e/tests/package.json:18 in 78d8e6b. [](commit_id = 78d8e6be6b878085b2aabf8f8cbf683dc6ec0c11, deletion_comment = False)

common/core/src/retry_operation.ts
anthonyvercolano: Why not simply do clearTimeout(this._currentTimeout);

---
In reply to: [270613536](https://github.com/Azure/azure-iot-sdk-node/pull/521#discussion_r270613536) [](ancestors = 270613536,270512971)

PR #542, opened 2019-04-19
(feature) - deviceScope property.

anthonyvercolano:


e2etests/test/registry.js
```diff
@@ -234,6 +234,63 @@ describe('Registry', function () {
...
+            registry.delete(edgeCreateResult.deviceId, function() {
+              done(scopedDeviceCreateError);
+            });
+          } else {
+            debug('Created scoped device: ', scopedDeviceCreateResult.deviceId);
+            assert.equal(scopedDeviceCreateResult.deviceScope, scopedDevice.deviceScope, 'Created scoped device does not contain correct scope.');
+            scopedDeviceCreateResult.deviceScope = uuid.v4();
+            registry.update(scopedDeviceCreateResult, function(updateErr) {
+              assert(updateErr, 'Scoped device was incorrectly allowed to update its scope property.');
+              registry.delete(edgeCreateResult.deviceId, function () {});
```
pierreca: >function () {} [](start = 57, length = 14)

even if you don't want to fail we would benefit from logging here.
Failing to delete would probably expose an IoT Hub bug though... not sure why we'd ignore that.

PR #544, opened 2019-04-21
Update packages from npm audit.  Also policy fixes.

anthonyvercolano:
Standard NPM audit warnings.

Removed policy words.

Removed some unused variables.

the unused variable check is depreciated in tslint.  Added it to the tsconfig.json in the individual directories.


BertKleewein:
>     "compilerOptions": {

This is out of scope, but have you considered putting all of the ts defaults into a root file, say tsconfig.defaults.json and using the extends keyword to include that in all the child tsconfig files?

e.g. 
tsconfig.defaults.json in root:
{
    "compilerOptions": {
        "target":"es5",
        "sourceMap": true,
        "declaration": true,
        "noUnusedLocals": false,
        "lib": [ "es6" ]
    }
}
tsconfig.json here (and everywhere else):
{
 "extends": "../../tsconfig.defaults",
 "compilerOptions": {
     "outDir": "./lib"
 },
 "include": [
     "./src/**/*"
 ]
}

---
Refers to: common/core/tsconfig.json:2 in ab2ad71. [](commit_id = ab2ad7183578519f549586a877b9ec5b131c7a94, deletion_comment = False)

common/core/tsconfig.json
```diff
@@ -4,7 +4,8 @@
         "target":"es5",
         "sourceMap": true,
         "declaration": true,
-        "noUnusedLocals": false,
+        "noUnusedLocals": true,
+        "noUnusedParameters": true,
```
BertKleewein: For what it's worth, I've verified that these default to false.

anthonyvercolano: Me too.

anthonyvercolano:
>     "compilerOptions": {

Thank you.  I was not aware of this ability to extend.  I will likely use it in a future rev.

---
In reply to: [485496018](https://github.com/Azure/azure-iot-sdk-node/pull/544#issuecomment-485496018) [](ancestors = 485496018)

---
Refers to: common/core/tsconfig.json:2 in ab2ad71. [](commit_id = ab2ad7183578519f549586a877b9ec5b131c7a94, deletion_comment = False)

PR #548, opened 2019-04-25
(fix) Improperly assuming that rid is the entire query string.

anthonyvercolano:


provisioning/transport/mqtt/src/mqtt.ts
```diff
@@ -52,11 +53,12 @@ export class Mqtt extends EventEmitter implements X509ProvisioningTransport, Sym
 
       /* Codes_SRS_NODE_PROVISIONING_MQTT_18_010: [ When waiting for responses, `registrationRequest` shall watch for messages with a topic named $dps/registrations/res/<status>/?$rid=<rid>.] */
       /* Codes_SRS_NODE_PROVISIONING_MQTT_18_024: [ When waiting for responses, `queryOperationStatus` shall watch for messages with a topic named $dps/registrations/res/<status>/?$rid=<rid>.] */
-      let match = topic.match(/^\$dps\/registrations\/res\/(.*)\/\?\$rid=(.*)$/);
+      let match = topic.match(/^\$dps\/registrations\/res\/(.*)\/\?(.*)$/);
+      let queryParameters = queryString.parse(match[2]);
```
pierreca: >queryString.parse(match[2]); [](start = 28, length = 28)

probably need to try/catch this

provisioning/transport/mqtt/src/mqtt.ts
```diff
@@ -52,11 +53,12 @@ export class Mqtt extends EventEmitter implements X509ProvisioningTransport, Sym
...
 
-      if (!!match && match.length === 3) {
+      if (!!match && match.length === 3 && queryParameters['$rid']) {
```
pierreca: >queryParameters['$rid'] [](start = 43, length = 23)

accessing the $rid property like this causes linter errors

pierreca:
needs a test that verifies that if the query string has 2 parameters ($rid and $retry-after) it doesn't crash.

provisioning/transport/mqtt/src/mqtt.ts
anthonyvercolano: queryString.parse does NOT appear to throw.

provisioning/transport/mqtt/src/mqtt.ts
anthonyvercolano: Indeed it does.  Changing it to simply use .$rid

anthonyvercolano:
Being reworked elsewhere.

Issue #560, opened 2019-05-20
Bug using Device.Client#close() as a Promise


kartben:
in azure-iot-device 1.9.9, there is a bug/typo in `Device.Client#close()`'s call to `azure-iot-common#callbackToPromise` (wrong callback being passed), making it impossible to "await" a device client's `close()`

YoDaMa:
Release 2019-07-17 includes this bugfix.

az-iot-builder-01:
@YoDaMa, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #561, opened 2019-05-20
Fix for #560 - bug in Device.Client#close()

kartben:
Call super.close() with proper callback to fix the behavior when Device.Client#close() is called with no callback (e.g when callers expects to get a Promise)

# Checklist

Fixes issue #560 

# Description of the problem

# Description of the solution


anthonyvercolano:
Carried over by #561

PR #568, opened 2019-06-05
(fix) buffer call updated

YoDaMa:
# Description

This PR accomplishes two things:

1) Update `Buffer()` syntax to new API.
the old syntax for `new Buffer()` is being deprecated in favor of different ways to call Buffer. All this PR does is update the Buffer calls to follow the new format, per the [linked SO post](https://stackoverflow.com/questions/52165333/deprecationwarning-buffer-is-deprecated-due-to-security-and-usability-issues)

2) Bumping up the supported Node.js runtime version to v8 or greater. 
After a discussion within the team it was determined that moving forward we should follow the supported versions of Node.js released by the Node.js foundation.


YoDaMa:
Holding off on completing this PR while we wait to determine the resolution to Back Compat.

PR #572, opened 2019-06-19
(feature:device): custom user agent

YoDaMa:
## Description

This is an implementation of the Custom User Agent option. 

## Specifications

Below are the specifications decided upon for the Custom User Agent for IoT Hub:

HTTP user-agent string  

Store the entire string in the User-Agent HTTP header 

AMQP user-agent string  

Store entire string in the “:client-version” property during link creation 

MQTT user-agent string  

Append entire string to Username field in CONNECT packet

### DIFF
--- a/device/core/src/interfaces.ts
+++ b/device/core/src/interfaces.ts
@@ -16,7 +16,7 @@ import { Agent } from 'https';
  * @see {DeviceClientOptions}
  */
 export interface AmqpTransportOptions {
-    /**
+  /**
    * Optional [Agent]{@link https://nodejs.org/api/https.html#https_class_https_agent} object to use with AMQP-WS connections
    */
   webSocketAgent?: Agent;
--- a/device/core/src/interfaces.ts
+++ b/device/core/src/interfaces.ts
@@ -16,10 +16,6 @@ import { Agent } from 'https';
  * @see {DeviceClientOptions}
  */
 export interface AmqpTransportOptions {
-  /**
-   * Optional [Agent]{@link https://nodejs.org/api/https.html#https_class_https_agent} object to use with AMQP-WS connections
-   */
-  webSocketAgent?: Agent;
 }
 
 /**
@@ -118,17 +114,23 @@ export interface DeviceClientOptions extends X509 {
   ca?: string;
 
   /**
-   * Optional object with options specific to the Mqtt transport
+   * Custom user defined information to be appended to existing User Agent information. The User Agent Identification information
+   * is used predominantly by Microsoft internally for identifying metadata related to Device Client usage for Azure IoT.
    */
-  mqtt?: MqttTransportOptions;
+  productInfo?: string;
 
   /**
-   * Optional object with options specific to the Mqtt transport
+   * Optional object with options specific to the MQTT transport
    */
-  http?: HttpTransportOptions;
+   mqtt?: MqttTransportOptions;
 
   /**
-   * Optional object with options specific to the Amqp transport
+   * Optional object with options specific to the HTTP transport
+   */
+   http?: HttpTransportOptions;
+
+  /**
+   * Optional object with options specific to the AMQP transport
    */
   amqp?: AmqpTransportOptions;
 }
--- a/device/core/src/interfaces.ts
+++ b/device/core/src/interfaces.ts
@@ -16,6 +16,10 @@ import { Agent } from 'https';
  * @see {DeviceClientOptions}
  */
 export interface AmqpTransportOptions {
+    /**
+   * Optional [Agent]{@link https://nodejs.org/api/https.html#https_class_https_agent} object to use with AMQP-WS connections
+   */
+  webSocketAgent?: Agent;
 }
 
 /**
--- a/device/core/src/utils.ts
+++ b/device/core/src/utils.ts
@@ -11,12 +11,39 @@ const packageJson = require('../package.json');
 
 export function getUserAgentString(done: NoErrorCallback<string>): void;
 export function getUserAgentString(): Promise<string>;
-export function getUserAgentString(done?: NoErrorCallback<string>): Promise<string> | void {
+export function getUserAgentString(productInfo: string, done: NoErrorCallback<string>): void;
+export function getUserAgentString(productInfo: string): Promise<string>;
+export function getUserAgentString(productInfoOrDone?: string | NoErrorCallback<string>, doneOrNone?: NoErrorCallback<string>): Promise<string> | void {
+  let productInfo: string;
+  let done: NoErrorCallback<string>;
+
+  /*Codes_SRS_NODE_DEVICE_UTILS_41_001: [`getUserAgentString` shall not add any custom product Info if a `falsy` value is passed in as the first arg.]*/
+  if (!productInfoOrDone) {
+    productInfo = '';
+    done = doneOrNone;
+  } else {
+    switch (typeof(productInfoOrDone)) {
+      /*Codes_SRS_NODE_DEVICE_UTILS_41_002: [`getUserAgentString` shall accept productInfo as a `string` so that the callback is called with a string in the form 'azure-iot-device/<packageJson.version>(<platformString>)<productInfo>'.]*/
+      case 'string': {
+        productInfo = <string> productInfoOrDone;
+        done = doneOrNone;
+        break;
+      }
+      case 'function': {
+        productInfo = '';
+        done = <NoErrorCallback<string>> productInfoOrDone;
+        break;
+      }
+      /*Codes_SRS_NODE_DEVICE_UTILS_41_003: [`getUserAgentString` shall throw if the first arg is not `falsy`, or of type `string` or `function`.]*/
+      default:
+        throw new TypeError('Error: productInfo must be of type \'string\'');
+    }
+  }
   return noErrorCallbackToPromise((_callback) => {
     /*Codes_SRS_NODE_DEVICE_UTILS_18_001: [`getUserAgentString` shall call `getAgentPlatformString` to get the platform string.]*/
     getAgentPlatformString((platformString) => {
-      /*Codes_SRS_NODE_DEVICE_UTILS_18_002: [`getUserAgentString` shall call its `callback` with a string in the form 'azure-iot-device/<packageJson.version>(<platformString>)'.]*/
-      _callback(packageJson.name + '/' + packageJson.version + ' (' + platformString + ')');
+      /*Codes_SRS_NODE_DEVICE_UTILS_18_002: [`getUserAgentString` shall call its `callback` with a string in the form 'azure-iot-device/<packageJson.version>(<platformString>)<productInfo>'.]*/
+      _callback(packageJson.name + '/' + packageJson.version + ' (' + platformString + ')' + productInfo);
     });
   }, done);
 }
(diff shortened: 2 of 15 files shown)


device/core/src/interfaces.ts
```diff
@@ -112,19 +112,21 @@ export interface DeviceClientOptions extends X509 {
    * connecting to some other hub (e.g. an Edge Hub), then you may need to set this to the server cert that the hub uses for TLS.
    */
   ca?: string;
-
   /**
-   * Optional object with options specific to the Mqtt transport
+   * Custom User Agent Option 
+   */
```
pierreca: when we ship we'll find a more verbose, explanatory thing to put here :)

device/transport/amqp/src/amqp.ts
```diff
@@ -682,6 +686,14 @@ export class Amqp extends EventEmitter implements DeviceTransport {
       }
     }
 
+    if (options.hasOwnProperty('productInfo')) {
+      if (typeof(options.productInfo) === 'string') {
+        this._customUserAgentString = options.productInfo;
+      } else {
+        throw new errors.InvalidOperationError('productInfo must be set as a string');
+      }
```
pierreca: typescript guidelines say we should try (within reason) to coerce types. i'd be fine with not testing for string and not throwing.

device/transport/amqp/src/amqp.ts
```diff
@@ -682,6 +686,14 @@ export class Amqp extends EventEmitter implements DeviceTransport {
...
+    }
+
     /*Codes_SRS_NODE_DEVICE_AMQP_13_001: [ The setOptions method shall save the options passed in. ]*/
     this._options = options;
```
pierreca: >this._options = options; [](start = 4, length = 24)

here it looks like we keep a reference to the options in the instance. given this, do we need the _customUserAgentString property? could we just pull it from the _options object directly line 285 (checking that this_options existnce of course?)

common/transport/mqtt/src/mqtt_base.ts
```diff
@@ -229,6 +229,10 @@ export class MqttBase extends EventEmitter {
           agent: this._options.mqtt.webSocketAgent
         };
       }
+      if (this._options.productInfo) {
+        // According to CUA Spec: "Append entire string to Username field in CONNECT packet" 
+        options.username += this._options.productInfo
```
anthonyvercolano: Do we need to do anything to url encode this?

device/transport/amqp/src/amqp.ts
YoDaMa: It sounds like you're saying I don't need to check for correct types. Will remove these errors.

device/core/src/interfaces.ts
YoDaMa: Yes.

common/transport/mqtt/src/mqtt_base.ts
pierreca: really good question - given the username isn't used as a URL i would assume no... but it might be worth double-checking with someone in the service team

---
In reply to: [295432142](https://github.com/Azure/azure-iot-sdk-node/pull/572#discussion_r295432142) [](ancestors = 295432142)

YoDaMa: So... I've updated this. Background: I thought that I should appending the productInfo to the options.username.... Essentially tagging it to the mqtt username. BUT then I dug deeper and I think for the productInfo rather than adding it to the username I should just append it to however we already do the userAgent.

device/core/src/internal_client.ts
```diff
@@ -359,7 +359,7 @@ export abstract class InternalClient extends EventEmitter {
     }, (err) => {
       /*Codes_SRS_NODE_INTERNAL_CLIENT_16_043: [The `done` callback shall be invoked no parameters when it has successfully finished setting the client and/or transport options.]*/
       /*Codes_SRS_NODE_INTERNAL_CLIENT_16_044: [The `done` callback shall be invoked with a standard javascript `Error` object and no result object if the client could not be configured as requested.]*/
-      safeCallback(done, err);
+      safeCallback(done, err, new results.TransportConfigured());
```
pierreca: >new results.TransportConfigured() [](start = 30, length = 33)

if this is a bugfix the requirement (and probably test needs to change as well)

device/core/src/utils.ts
```diff
@@ -20,3 +20,21 @@ export function getUserAgentString(done?: NoErrorCallback<string>): Promise<stri
     });
   }, done);
 }
+
+export function getCustomUserAgentString(productInfo: string, done: NoErrorCallback<string>): void;
+export function getCustomUserAgentString(productInfo: string): Promise<string>;
+export function getCustomUserAgentString(productInfo: string, done?: NoErrorCallback<string>): Promise<string> | void {
```
pierreca: >getCustomUserAgentString [](start = 16, length = 24)

consider overloading getUserAgentString rather than creating a new API.

device/core/test/_internal_client_test.js
```diff
@@ -202,6 +202,25 @@ var ModuleClient = require('../lib/module_client').ModuleClient;
           testCallback();
         });
       });
+
+      it('productInfo properly sets string', function (done) {
+        var testOptions = { productInfo: 'test'};
+        var dummyTransport = new FakeTransport();
+        sinon.stub(dummyTransport, 'setOptions').callsFake(function (options, callback) {
+          assert.equal(options.productInfo, testOptions.productInfo);
```
pierreca: >equal [](start = 17, length = 5)

use strictEqual

device/core/test/_internal_client_test.js
```diff
@@ -202,6 +202,25 @@ var ModuleClient = require('../lib/module_client').ModuleClient;
...
+          callback(null, new results.TransportConfigured());
+        });
+
+        var client = new ClientCtor(dummyTransport);
+        client.setOptions(testOptions, function (err, results) {
+          if (err) {
+            done(err);
+          } else {
+            assert.equal(results.constructor.name, 'TransportConfigured');
+            done();
```
pierreca: unnecessary assert.

device/transport/amqp/src/amqp.ts
```diff
@@ -278,8 +278,11 @@ export class Amqp extends EventEmitter implements DeviceTransport {
                   this._d2cEndpoint = endpoint.deviceEventPath(credentials.deviceId);
                   this._messageEventName = 'message';
                 }
-
-                getUserAgentString((userAgentString) => {
+                let customInfo = '';
+                if (this.hasOwnProperty('_options') && this._options.hasOwnProperty('productInfo')) {
```
pierreca: >this.hasOwnProperty('_options') && this._options.hasOwnProperty('productInfo') [](start = 20, length = 78)

if (this._options && this._options.productInfo)

pierreca: >this.hasOwnProperty('_options') && this._options.hasOwnProperty('productInfo') [](start = 20, length = 78)

const customInfo = (this._options && this._options.productInfo) ? this._options._productInfo : undefined;

device/transport/amqp/test/_amqp_test.js
```diff
@@ -14,6 +14,8 @@ var Amqp = require('../lib/amqp.js').Amqp;
 var errors = require('azure-iot-common').errors;
 var results = require('azure-iot-common').results;
 var AuthenticationType = require('azure-iot-common').AuthenticationType;
+var getUserAgentString = require('azure-iot-device').getUserAgentString;
+
```
pierreca: is this necessary?

device/transport/http/src/http.ts
```diff
@@ -325,6 +326,11 @@ export class Http extends EventEmitter implements DeviceTransport {
       });
     }
 
+    if (options.hasOwnProperty('productInfo')) {
+      // According to Custom User Agent Spec: "Store the entire string in the User-Agent HTTP header"
+      this._productInfo = options.productInfo;
+    }
```
pierreca: use ternary operator

device/transport/http/src/http.ts
```diff
@@ -786,10 +792,14 @@ export class Http extends EventEmitter implements DeviceTransport {
     if (this._userAgentString) {
       done();
     } else {
-      getUserAgentString((agent) => {
+      if (!this.hasOwnProperty('_productInfo')) {
+        this._productInfo = '';
+      }
```
pierreca: not needed if using the ternary operator

device/transport/http/test/_http_test.js
```diff
@@ -349,6 +349,12 @@ describe('Http', function () {
         transport.setOptions({});
       });
     });
+
+    it('sets _productInfo if productInfo is provided', function() {
+      var transport = new Http(fakeAuthenticationProvider);
+      transport.setOptions({ productInfo: 'fakeProductInfoString' });
+      assert.exists(transport._productInfo);
+    });
```
pierreca: a better test would be to use a mock object as the second parameter of the Http constructor and assert that the common HTTP object is called with the User-Agent header that contains productInfo

device/transport/mqtt/src/mqtt.ts
```diff
@@ -929,7 +934,11 @@ export class Mqtt extends EventEmitter implements DeviceTransport {
     if (this._userAgentString) {
       done();
     } else {
-      getUserAgentString((agent) => {
+      let customInfo = '';
+      if (this.hasOwnProperty('_productInfo')) {
+        customInfo = this._productInfo;
```
pierreca: use ternary operator?

device/transport/mqtt/test/_mqtt_test.js
```diff
@@ -555,9 +558,21 @@ describe('Mqtt', function () {
...
+        testCallback();
+      });
+    });
+
+
   });
 
   describe('#connect', function() {
+    var fakeProductInfoOptions = {productInfo: 'fakeProductInfo'}
+
```
pierreca: remove and if this fails we have a backcompat' problem

device/transport/mqtt/test/_mqtt_test.js
```diff
@@ -555,9 +558,21 @@ describe('Mqtt', function () {
         testCallback();
       });
     });
+
+    it('sets options for productInfo', function(testCallback) {
+      var mqtt = new Mqtt(fakeX509AuthenticationProvider);
+      mqtt.setOptions(fakeProductInfoOptions, function (err) {
+        assert.strictEqual(mqtt._productInfo, fakeProductInfoOptions.productInfo);
+        testCallback();
```
pierreca: same as in HTTP - a better test would be to verify the username property on mqttbase.connect

pierreca:
watch out for the changes in e2etests, it seems you're reverting some fixes i made a couple of days ago

YoDaMa:
> watch out for the changes in e2etests, it seems you're reverting some fixes i made a couple of days ago. same thing in the service sdk folder.

noted and removed

device/core/src/utils.ts
```diff
@@ -11,12 +11,39 @@ const packageJson = require('../package.json');
 
 export function getUserAgentString(done: NoErrorCallback<string>): void;
 export function getUserAgentString(): Promise<string>;
-export function getUserAgentString(done?: NoErrorCallback<string>): Promise<string> | void {
+export function getUserAgentString(produtInfo: string, done: NoErrorCallback<string>): void;
```
pierreca: >produtInfo [](start = 35, length = 10)

typo

PR #575, opened 2019-06-28
Add logging and update timeouts for device_acknowledge_tests

pierreca:
We used to timeout before the service. Increasing timeouts so that the service times out before the tests do (and we capture the timeout instead of ending up in a bad state).

Also adding a bit of logging in the device_acknowledge tests and the AMQP transport of the service SDK


e2etests/test/device_acknowledge_tests.js
```diff
@@ -81,21 +81,28 @@ function device_acknowledgment_tests (deviceTransport, createDeviceMethod) {
...
                   } else {
+                    debug('+++Message abandonned');
                     assert.equal(result.constructor.name, 'MessageAbandoned');
                   }
                 });
               } else {
+                debug('+++completing the message');
                 deviceClient.complete(msg, function (err, res) {
                   if (err) {
+                    debug('+++error completing the message: ' + err.toString());
```
anthonyvercolano: >err.toString [](start = 64, length = 12)

Not sure that these string very well.

Issue #574, opened 2019-06-27
Unable to use AMQP over web socket with HTTPS proxy due to AmqpTransportOptions missing webSocketAgent property


dimiboi:
I am unable to use Azure IoT Hub Device SDK with an explicit corporate HTTPS proxy when using AMQP over web socket to connect to IoT Hub.

[`AmqpTransportOptions`](https://github.com/Azure/azure-iot-sdk-node/blob/734bd35f0a48b519ee4de233d1762e7ef2d4521c/device/core/src/interfaces.ts#L18) interface is missing `webSocketAgent` property, which is present in e.g. [`MqttTransportOptions`](https://github.com/Azure/azure-iot-sdk-node/blob/734bd35f0a48b519ee4de233d1762e7ef2d4521c/device/core/src/interfaces.ts#L79) interface. As a result, it is not possible to specify an HTTPS proxy agent for the web socket required to establish IoT Hub connection via HTTPS proxy.

Here is the [user voice idea](https://feedback.azure.com/forums/321918-azure-iot/suggestions/38023516-use-amqp-over-web-socket-connection-to-iot-hub-wit).

YoDaMa:
@dimiboi thanks for filing this request for an enhancement to the SDK. We'll take a look to assess where we have the resources to take action on this enhancement. Of course, we also accept outside PRs.

pierreca:
I believe this was adressed with https://github.com/Azure/azure-iot-sdk-node/releases/tag/2019-07-17
Closing. as resolved. Thanks for your contribution!

az-iot-builder-01:
@YoDaMa, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #579, opened 2019-07-04
HTTPS agent support added to AMQP over WebSocket transport

dimiboi:
# Checklist

#574 

# Description of the problem
Unable to use HTTPS agent with AMQP over WebSocket transport.

# Description of the solution
`webSocketAgent` property added to `AmqpTransportOptions` and passed to `setOptions` function to be used by `connect` function.

### DIFF
--- a/common/transport/amqp/src/amqp.ts
+++ b/common/transport/amqp/src/amqp.ts
@@ -707,7 +707,8 @@ export class Amqp {
     connectionParameters.host = parsedUrl.hostname;
     connectionParameters.reconnect = false;
     if (parsedUrl.protocol === 'wss:') {
-      if (config.sslOptions && this._options && this._options.amqp && this._options.amqp.webSocketAgent) {
+      if (this._options && this._options.amqp && this._options.amqp.webSocketAgent) {
+        config.sslOptions = config.sslOptions || { };
         config.sslOptions.agent = this._options.amqp.webSocketAgent;
       }
       let webSocket = require('ws');
--- a/common/transport/amqp/src/amqp.ts
+++ b/common/transport/amqp/src/amqp.ts
@@ -707,6 +707,7 @@ export class Amqp {
     connectionParameters.host = parsedUrl.hostname;
     connectionParameters.reconnect = false;
     if (parsedUrl.protocol === 'wss:') {
+      /*Codes_SRS_NODE_COMMON_AMQP_99_043: [The `connect` method shall set the `config.sslOptions.agent` option based on the `amqp.webSocketAgent` object passed in the `options` structure via the `setOptions` function.]*/
       if (this._options && this._options.amqp && this._options.amqp.webSocketAgent) {
         config.sslOptions = config.sslOptions || { };
         config.sslOptions.agent = this._options.amqp.webSocketAgent;
--- a/common/transport/amqp/src/amqp.ts
+++ b/common/transport/amqp/src/amqp.ts
@@ -116,6 +116,7 @@ export class Amqp {
   // Structure containing the connection information needed such as uri, ssl (cert) etc.
   //
   private _config: AmqpBaseTransportConfig;
+  private _options: any;
 
   /*Codes_SRS_NODE_COMMON_AMQP_16_001: [The Amqp constructor shall accept two parameters:
     A Boolean indicating whether the client should automatically settle messages:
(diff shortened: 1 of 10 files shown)


common/transport/amqp/test/_amqp_test.js
```diff
@@ -163,6 +163,46 @@ describe('Amqp', function () {
...
+      sinon.stub(amqp._rheaContainer, 'connect').callsFake(() => {
+        process.nextTick(() => {
+          amqp._rheaConnection.emit('connection_open', fakeConnectionContext);
+        });
+        return fakeConnection;
+      });
+      var fakeSession = new EventEmitter();
+      var fakeSessionContext = {session: fakeSession};
+      fakeConnection.create_session = sinon.stub().returns(fakeSession);
+      fakeSession.open = () => {};
```
YoDaMa: nit: is this line necessary when you're stubbing the open function in the next line? Or since you're creating this fake session could you not even stub it, just use this line to define the same behavior as the stub in the next line?

common/transport/amqp/src/amqp.ts
```diff
@@ -706,6 +707,11 @@ export class Amqp {
     connectionParameters.host = parsedUrl.hostname;
     connectionParameters.reconnect = false;
     if (parsedUrl.protocol === 'wss:') {
+      /*Codes_SRS_NODE_COMMON_AMQP_99_043: [The `connect` method shall set the `config.sslOptions.agent` option based on the `amqp.webSocketAgent` object passed in the `options` structure via the `setOptions` function.]*/
+      if (this._options && this._options.amqp && this._options.amqp.webSocketAgent) {
```
pierreca: >this._options.amqp.webSocketAgent [](start = 49, length = 33)

I think this would belong in the config object rather than in a separate _options object here. moving it to the config would have a couple of desirable effects:
- group it with the rest of sslOptions
- remove the need to add a setOptions method to the AmqpBase class

pierreca:
@dimiboi thanks for this PR. In addition to the comment i added in `common/transport/amqp/amqp.ts`, pls take a look at https://github.com/Azure/azure-iot-sdk-node/blob/master/device/transport/amqp/src/amqp.ts#L289

I think the websocket agent option would belong here rather than push the options structure through an additional setOptions API in the common amqp file.

dimiboi:
@pierreca, sure, I shall look into them.

common/transport/amqp/src/amqp.ts
dimiboi: @pierreca, the intention here is to keep the API consistent with MQTT transport implementation, which uses [setOptions](https://github.com/Azure/azure-iot-sdk-node/blob/026fa2bcef51f5ae8f8a000fa4cb42038bdf089c/common/transport/mqtt/src/mqtt_base.ts#L206) function, which sets a separate [_options](https://github.com/Azure/azure-iot-sdk-node/blob/026fa2bcef51f5ae8f8a000fa4cb42038bdf089c/common/transport/mqtt/src/mqtt_base.ts#L23) object and takes [MqttTransportOptions](https://github.com/Azure/azure-iot-sdk-node/blob/734bd35f0a48b519ee4de233d1762e7ef2d4521c/device/core/src/interfaces.ts#L79) interface instance as a parameter. There's a similar [HttpTransportOptions](https://github.com/Azure/azure-iot-sdk-node/blob/734bd35f0a48b519ee4de233d1762e7ef2d4521c/device/core/src/interfaces.ts#L62) interface used the same way.

pierreca: I agree with this. this comment complements what i'm suggesting in the main discussion thread but it's probably not very clear so i'll try to rephrase:
- keep the options structure the same, as well as the setOptions call. I agree the external API should be consistent.
- In  `device/transport/amqp/src/amqp.ts`, the caller of this connect function is creating the config object that already has an sslOptions property. Insert the websocketAgent property at this stage. ([https://github.com/Azure/azure-iot-sdk-node/blob/master/device/transport/amqp/src/amqp.ts#L289](https://github.com/Azure/azure-iot-sdk-node/blob/master/device/transport/amqp/src/amqp.ts#L289))
- remove all code in this class, the config.sslOptions object is already being passed to the ws constructor.

i think that would work, wouldn't it? and it would keep all sslOptions together internally, while providing a consistent API at the client level (ie all options object are consistent as you're suggesting)

---
In reply to: [301997067](https://github.com/Azure/azure-iot-sdk-node/pull/579#discussion_r301997067) [](ancestors = 301997067)

dimiboi: @pierreca, thank you, I get it now, I am on it.

common/transport/amqp/test/_amqp_test.js
dimiboi: No longer relevant.

common/transport/amqp/src/amqp.ts
dimiboi: @pierreca, done.

pierreca:
Squashed and merged using #584. For some reason github wouldn't let me merge directly from here even though all checked had passed.

@dimiboi thanks a lot for your contributions. we will be releasing this within the next couple of days.

PR #589, opened 2019-07-19
Add proxy support to blob file upload in Device SDK

dimiboi:
# Checklist
- [x] I have read the [contribution guidelines](https://github.com/Azure/azure-iot-sdk-node/blob/master/.github/CONTRIBUTING.md).

# Description of the problem
Currently, it is impossible to perform blob file upload behind a corporate HTTP proxy.

# Description of the solution
The PR enables an SDK user to pass an agent to the HTTP transport used in file upload API client and set the proxy used by file upload blob service.

### DIFF
--- a/build/tools/package-lock.json
+++ b/build/tools/package-lock.json
@@ -1,6 +1,8 @@
 {
-  "requires": true,
+  "name": "iothub-buildtools",
+  "version": "0.0.1",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
     "ansi-regex": {
       "version": "2.1.1",
--- a/common/transport/amqp/src/amqp_common_errors.ts
+++ b/common/transport/amqp/src/amqp_common_errors.ts
@@ -106,6 +106,10 @@ export function translateError(message: string, amqpError: Error): AmqpTransport
     }
   } else if ((<any>amqpError).code) {
     error = new errors.NotConnectedError(message);
+  } else if ((<any>amqpError).message && (<any>amqpError.message).message ) {
+    // In the case of a invalid Twin object, this will return the generic error message plus the specific error messaged provided from the server.
+    let errorString = message + '. ' + (<any>amqpError.message).message.toString();
+    error = new Error(errorString);
   } else {
     /*Codes_SRS_NODE_DEVICE_AMQP_COMMON_ERRORS_16_002: [If the AMQP error code is unknown, `translateError` should return a generic Javascript `Error` object.]*/
     error = new Error(message);
--- a/common/transport/amqp/test/_amqp_common_errors_test.js
+++ b/common/transport/amqp/test/_amqp_common_errors_test.js
@@ -42,6 +42,7 @@ describe('translateError', function() {
     { errorDescription: 'com.microsoft:device-already-exists', errorMessage: 'Device already exists', expectedErrorType: errors.DeviceAlreadyExistsError },
     { errorDescription: 'com.microsoft:device-container-throttled', errorMessage: 'Throttled', expectedErrorType: errors.ThrottlingError },
     { errorDescription: 'com.microsoft:iot-hub-suspended', errorMessage: 'IoT hub suspended', expectedErrorType: errors.IoTHubSuspendedError },
+    { errorDescription: 'com.microsoft:iot-hub-not-found-error', errorMessage: 'IoT hub not found', expectedErrorType: errors.IotHubNotFoundError },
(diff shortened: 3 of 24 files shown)


pierreca:
@dimiboi thanks for the PR!

not seeing a sample showing the usage for this, i'm assuming you're setting the http agent in the options, so that it gets passed to `FileUploadApi` class and configured with it. That's also the approach i would have taken.

Have you looked at passing the same agent to the storage sdk (used by `BlobUploader`)? if it's possible it's an approach that'd be a lot easier than carrying around a second option (fileUploadProxy) just for the sake of calling setProxy. If not passing the agent directly, maybe it'd be possible to extract the proxy URL from the agent and pass it to the storage sdk?

The bottom line is that i would really like to avoid having 2 options for the same feature (agent + fileUploadProxy).

If that's absolutely impossible, the other thing i've been considering for a while is to deprecate the "overarching" `uploadToBlob`  method and expose the FileUploadApi methods (`getBlobSharedAccessSignature` and `notifyUploadComplete`) directly on the client. This would be a little more code for the user (3 API calls instead of one, basically) but it would let the user control the storage sdk completely. In retrospect, it seems like much cleaner design compared to that big `uploadToBlob` function that is hampered by limitations (not being able to configure the storage SDK) and has a confusing behavior (if the blob upload fails but the end notification succeeds,, as it is now, the `uploadToBlob` function will appear to have succeeded)

Thoughts?

dimiboi:
@pierreca, I configured proxy the following way and verified with Fiddler that all the connections are established using the configured proxy:

```javascript
const proxy = 'http://localhost:8866';
const agent = new HttpsProxyAgent(proxy);

await client.setOptions({
  http: {
    agent: agent,
  },
  amqp: {
    webSocketAgent: agent,
  },
  fileUploadProxy: proxy,
});
```

I have looked into using an agent with Storage SDK. Unfortunately, [setProxy](https://github.com/Azure/azure-storage-node/blob/0557d02cd2116046db1a2d7fc61a74aa28c8b557/lib/common/services/storageserviceclient.js#L1227) function is the only way I could make it use a proxy server.

Unfortunately, exposing `FileUploadApi` won't help with addressing the issue that two separate options are required. The HTTP transport it uses needs an agent to be set, and its blob uploader needs a proxy address. This is a limitation imposed by Storage SDK.

pierreca:
Understood. To be honest, i'm not happy with how the SDK handles file upload in its current form:
1. pushing the dependency of azure-storage on all customers including those who don't do file upload is wrong
2. having to define the proxy twice in the same options structure is clunky
3. we hide too much of the azure-storage features under this "weak" uploadToBlob API
4. we have it written right now so that if the upload fails, but the notification to the hub succeeds, the failure to upload is signaled to the hub but not the device itself (device developer doesn't know and can't handle failure).
 
Here's the change I've had in mind for a while that this issue could precipitate (I could potentially have a team member work on that on the second half of August):

1. we surface getBlobSharedAccessSignature and notifyUploadComplete on the device client itself
2. we make these 2 APIs use existing proxy settings
3. we let users work out the file upload directly using the azure-storage SDK rather than hide it under our SDK

You'd be able to achieve what you want as well, we'd be having a better separation of concerns (ultimately deprecating the uploadToBlob function itself, which has other issues than just proxy settings) and would free azure-iot-device users from always carrying an azure-storage dependency: they would simply have to have it if they use the file upload feature, which a lot of users don't.

thoughts? would this approach work for you?

dimiboi:
It sounds good but we need a solution rather sooner than later. I have the following proposal.

I have studied the source code of Azure Storage SDK for Node.js and it turned out that it [supports getting proxy settings from environment variables](https://github.com/Azure/azure-storage-node/blob/0557d02cd2116046db1a2d7fc61a74aa28c8b557/lib/common/services/storageserviceclient.js#L1261). I have given it a try, and it indeed works. 

What this means is that we can use environment variables for file uploads to go via HTTPS proxy without adding `fileUploadProxy`. Such an approach should unblock my team already. However, we still need the functionality to propagate transport options to the HTTP transport used by File Upload API client to fetch upload token.

So how about I remove `fileUploadProxy` from transport options and `setProxy()` function call from the blob service yet keep the rest of the PR, which addresses the issue of the agent not being propagated to the HTTP transport?

pierreca:
@dimiboi 
>So how about I remove fileUploadProxy from transport options and setProxy() function call from the blob service yet keep the rest of the PR, which addresses the issue of the agent not being propagated to the HTTP transport?

Agreed, that works for me too.

Issue #587, opened 2019-06-24
Updating reported properties should show error message


flodan:
## Expected Behavior
When trying to update the reported properties with a JSON object, with a structure that is not supported, there should be an error message or the update should throw an error.

## Current Behavior
There is no message at all, instead it seems that the update went through but the reported properties are never updated.

## Steps to Reproduce

Code snippet to reproduce:
```
twin.on('properties.desired', async desiredProperties => {
  await moduleTwin.properties.reported.update(desiredProperties, error => {
    if (error) throw error // There should be an error here
    debug(`Updated module-twin reported properties with: ${JSON.stringify(desiredProperties)}`) // Instead this line is printed but the reported properties are not updated
  })
}
```

### Architecture
arm32

### Container Operating System
Linux containers

#### iotedged

iotedge 1.0.7.1

#### Edge Agent

1.0

#### Edge Hub

1.0

#### Docker

```
<Run `docker version` (`docker -H npipe:////./pipe/iotedge_moby_engine version` for Moby on Windows) and paste here>
3.0.5
```

darobs:
Hello @flodan, I transferred this over to the node.js SDK repo - we confirmed it happens outside of the IoT Edge context.

YoDaMa:
Including logs from my repro run in this thread. 
```
  azure-iot-common:RetryOperation Operation start time: 1563408660399 - Will stop retrying after: 1563408900399 +0ms
  azure-iot-device-mqtt:Mqtt disconnected -> connecting (disconnected.connect) +0ms
  azure-iot-device-mqtt:Mqtt topic publish: devices/alpha/messages/events/ +3ms
  azure-iot-device-mqtt:Mqtt message topic subscribe: devices/alpha/messages/devicebound/# +1ms
  azure-iot-mqtt-base:MqttBase disconnected -> connecting (disconnected.connect) +0ms
  azure-iot-mqtt-base:MqttBase ****?api-version=2019-03-30&DeviceClientType=azure-iot-device%2F1.10.0%20(node%20v10.15.3%3B%20Windows_NT%2010.0.17763%3B%20x64) +1ms
  azure-iot-mqtt-base:MqttBase uri:      **** +1ms
  azure-iot-mqtt-base:MqttBase Device is connected +8s
[...]

PR #590, opened 2019-07-19
improved error logging for twin failures

YoDaMa:
This PR Solves two issues. 

1) For #587 , this fixes the issue where in MQTT an erroneous Twin value can be set and an error will not be returned. That is the code in the mqtt twin client being added.

2) While reproing for the issue, I noticed that the AMQP error code for an erroneous Twin value was not very verbose. It would just print a generic error:
```
Error: could not update twin reported properties is not retriable
```
With this new code the Error logging is updated to print a full Error status including the message received from the server: 
```
Error: could not update twin reported properties: Provided json is not a valid 'tags' or 'properties' object: arrays are not supported
```

### DIFF
--- a/device/transport/mqtt/src/mqtt_twin_client.ts
+++ b/device/transport/mqtt/src/mqtt_twin_client.ts
@@ -257,6 +257,7 @@ export class MqttTwinClient extends EventEmitter {
         const parsedMessage = responseBody ? JSON.parse(responseBody) : undefined;
         /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_007: [When a message is received on the response topic with an `$rid` property in the query string of the topic matching the one that was sent on the request topic, the `callback` shall be called with a `null` error object and the parsed content of the response message.]*/
         /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_017: [When a message is received on the response topic with an `$rid` property in the query string of the topic matching the one that was sent on the request topic, the `callback` shall be called with a `null` error object.]*/
+        /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_41_001: [The `callback` shall be called with a Error object containing the error message if the `parsedMessage` contains an errorCode.]*/
         if (parsedMessage && parsedMessage.errorCode) {
           requestCallback(new Error(parsedMessage.message));
         } else {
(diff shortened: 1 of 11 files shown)


YoDaMa:
Note: I need to update some of the requirements docs before this PR goes in.

device/transport/mqtt/src/mqtt_twin_client.ts
```diff
@@ -254,9 +254,14 @@ export class MqttTwinClient extends EventEmitter {
...
         const responseBody = message.toString();
+        const parsedMessage = responseBody ? JSON.parse(responseBody) : undefined;
         /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_007: [When a message is received on the response topic with an `$rid` property in the query string of the topic matching the one that was sent on the request topic, the `callback` shall be called with a `null` error object and the parsed content of the response message.]*/
         /*Codes_SRS_NODE_DEVICE_MQTT_TWIN_CLIENT_16_017: [When a message is received on the response topic with an `$rid` property in the query string of the topic matching the one that was sent on the request topic, the `callback` shall be called with a `null` error object.]*/
-        requestCallback(null, responseBody ? JSON.parse(responseBody) : undefined);
+        if (parsedMessage && parsedMessage.errorCode) {
+          requestCallback(new Error(parsedMessage.message));
+        } else {
+          requestCallback(null, parsedMessage);
+        }
```
anthonyvercolano: Will be interesting to see how you test....

YoDaMa: thank you for the feedback.

device/transport/mqtt/test/_mqtt_twin_client_test.js
```diff
@@ -266,6 +266,32 @@ describe('MqttTwinClient', function () {
...
+      });
+
+      var twinClient = new MqttTwinClient(fakeMqttBase);
+      var fakeTwin = {
+        fake: 'patch',
+        errorCode: 'fakeErrorCode'
+      };
+      twinClient.updateTwinReportedProperties(fakeTwin, function (err, twin) {
+        assert.instanceOf(err, Error);
+        assert(!twin);
```
anthonyvercolano: >assert(!twin); [](start = 8, length = 14)

It has two other assert functions
assert.isNotOk
assert.isOk

YoDaMa: noted.

PR #594, opened 2019-08-07
node-nightly-df.yaml conditionally makes and copies release artifacts

fratster:
node-nightly-df.yaml conditionally makes and copies release artifacts. node-canary pipeline accepts configurable inputs with default values that preserve existing behaviour.

•	new variables.test to enable/disable tests for pipeline debugging
•	new variables.release to enable npm -pack artifact generation and copy
•	new tasks: DownloadBuildArtifacts@0, PublishBuildArtifacts@1, AzureFileCopy@2
•	new script: release-node.ps1


vsts/node-nightly-df.yaml
```diff
@@ -101,8 +106,46 @@ phases:
 
   - task: PublishTestResults@2
     displayName: 'Publish Test Results | Mocha'
+    condition: eq(variables.test, True)
```
anthonyvercolano: >condition: eq(variables.test, True) [](start = 4, length = 35)

Remove this variable.test conditions.

PR #610, opened 2019-08-21
Move file upload APIs to the client in order to give users the ability to use the storage SDK directly

YoDaMa:
Currently the API for UploadToBlob uses Storage SDK for Node.js v2 under the surface. That is very old, and also has limitations for customers who would like to use different aspects of the Storage SDK besides the UploadBlockBlobUsingSAS which we are currently using. 

This new API exposes FileUploadAPI which is already used under the surface to connect to IoT Hub over POST requests and retrieve a Storage Account SAS Token for the linked Storage Account and notify IoT Hub of the success or failure of the upload to blob execution. 

The two APIs proposed are this:

* `uploadToBlobV2GetStorageBlobSAS(blobName: string, callback?: Callback<UploadParams>): Promise<UploadParams> | void`
* `uploadToBlobV2NotifyBlobUploadComplete(uploadResponse: BlobUploadCommonResponseStub, callback?: ErrorCallback): Promise<void> | void`

This PR also includes a sample utilizing the new Storage SDK for Node.js v10. 

* _device/samples/upload_to_blob_v2.js_

There is a public preview of Storage SDK for Node.js v12 available, but since it is not GA, I am hesitant to implement a sample with it, since changes might happen. 

**NOTE**: Storage SDK is releasing v12 and I have decided to delay this PR until I have confirmation on the blobUploadResponse type. If it breaks from v10, then our API would break as well and we couldn't support v12. Since this is a new API, supporting the latest version is important.


device/core/src/device_client.ts
```diff
@@ -9,12 +9,14 @@ const debug = dbg('azure-iot-device:DeviceClient');
...
-import { BlobUploadClient } from './blob_upload';
+import { BlobUploadClient, UploadParams, BlobUploadResult2 } from './blob_upload';
 import { SharedAccessSignatureAuthenticationProvider } from './sas_authentication_provider';
 import { X509AuthenticationProvider } from './x509_authentication_provider';
 import { SharedAccessKeyAuthenticationProvider } from './sak_authentication_provider';
 import { DeviceMethodRequest, DeviceMethodResponse } from './device_method';
-import { DeviceClientOptions } from './interfaces';
+import { DeviceClientOptions, BlobUploadCommonResponseStub } from './interfaces';
+
+
```
pierreca: nit: watch out for unnecessary blank lines

device/core/src/interfaces.ts
```diff
@@ -5,6 +5,7 @@
 
 import { X509 } from 'azure-iot-common';
 import { Agent } from 'https';
+import { HttpOperationResponse } from '@azure/core-http';
```
pierreca: >@azure/core-http [](start = 39, length = 16)

that seems like a big dependency to take on for just a type

device/core/src/blob_upload/blob_upload_result_2.ts
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+
+import { BlobUploadCommonResponseStub } from './../interfaces';
+
+// import {
+//   BlobUploadCommonResponse
+// } from "@azure/storage-blob"; // Change to "@azure/storage-blob" in your package
```
pierreca: should be removed if not used.

device/core/src/blob_upload/blob_upload_result_2.ts
```diff
@@ -0,0 +1,44 @@
...
+
+
+export class BlobUploadResult2 {
```
pierreca: >BlobUploadResult2 [](start = 13, length = 17)

this seems to be the same class as BlobUploadResult so i don't see the need for this?

device/core/src/blob_upload/blob_upload_client.ts
```diff
@@ -55,26 +46,26 @@ export interface BlobUpload {
  * @private
  */
 export class BlobUploadClient implements BlobUpload {
+  public clientFileUploadApi: FileUploadInterface;
```
pierreca: >clientFileUploadApi [](start = 9, length = 19)

i don't think there's a need to make this public. you could create the FileUploadApi object in the device client and pass it to the blob client when you instantiate it. then use it directly in the client where it was instantiated.

device/core/src/interfaces.ts
YoDaMa: I will stub the HttpOperationResponse

device/core/src/device_client.ts
YoDaMa: noted.

device/core/src/blob_upload/blob_upload_client.ts
YoDaMa: fixed this.

device/core/package.json
```diff
@@ -7,6 +7,7 @@
   "main": "device.js",
   "typings": "device.d.ts",
   "dependencies": {
+    "@azure/storage-blob": "^12.0.0-preview.2",
```
pierreca: >@azure/storage-blob [](start = 5, length = 19)

if we're taking this on i'd assume we don't need azure-storage (below) anymore?

device/core/src/device_client.ts
```diff
@@ -171,13 +175,77 @@ export class Client extends InternalClient {
...
+  /**
+   * @description      The `uploadToBlobV2GetStorageBlobSAS` gets the linked storage account SAS Token from IoT Hub
+   *
+   * @param {String}    blobName                The name to use for the blob that will be created with the content of the stream.
+   * @param {Callback}  [callback]              Optional callback to call when the upload is complete.
+   * @returns {Promise<UploadParams> | void}    Promise if no callback function was passed, void otherwise.
+   *
+   * @throws {ReferenceException} If blobName is falsy.
+   */
+  uploadToBlobV2GetStorageBlobSAS(blobName: string, callback: Callback<UploadParams>): void;
```
pierreca: >uploadToBlobV2GetStorageBlobSAS [](start = 2, length = 31)

I don't think the "uploadToBlobV2" prefix adds clarity - in fact, it may even be confusing if we ship a v2 SDK. Also, using acronyms may make the name smaller but not necessarily easier to use.
how about simply "getBlobSharedAccessSignature and notifyBlobUploadComplete?

device/core/src/device_client.ts
```diff
@@ -298,6 +366,6 @@ export class Client extends InternalClient {
 
     /*Codes_SRS_NODE_DEVICE_CLIENT_16_090: [The `fromAuthenticationProvider` method shall pass the `authenticationProvider` object passed as argument to the transport constructor.]*/
     /*Codes_SRS_NODE_DEVICE_CLIENT_16_091: [The `fromAuthenticationProvider` method shall return a `Client` object configured with a new instance of a transport created using the `transportCtor` argument.]*/
-    return new Client(new transportCtor(authenticationProvider), null, new BlobUploadClient(authenticationProvider));
+    return new Client(new transportCtor(authenticationProvider), null, new BlobUploadClient(authenticationProvider), new DefaultFileUploadApi(authenticationProvider));
```
pierreca: >new BlobUploadClient(authenticationProvider), new DefaultFileUploadApi(authenticationProvider) [](start = 71, length = 94)

if you instantiate the FileUploadApi class but don't pass it to the BlobUploadClient it's going to instantiate one on its own which might waste resources. I think you should pass the fileuploadApi you create to the blob upload client.

device/core/src/blob_upload/blob_upload_result.ts
```diff
@@ -66,4 +67,24 @@ export class BlobUploadResult {
 
     return uploadResult;
   }
+
+  static fromAzureStorageCallbackArgs2(uploadResponse: BlobUploadCommonResponseStub): BlobUploadResult {
```
pierreca: >fromAzureStorageCallbackArgs2 [](start = 9, length = 29)

versioning APIs in names is bad practice. you can probably see for yourself a lot of duplicate code between this API and the original one. 
You should strive to minimize the amount of code you write because you'll have to maintain it.

the initial API isn't public. is there a way to modify it so that it works in both cases?

YoDaMa:
pinging @dimiboi about this PR. I would like to know how to validate/test that HTTP proxy support is not affected by these changes. Did you add a test with your PR to make sure that as I update to the new version of the Storage SDK, the http proxy support is not broken.

dimiboi:
@YoDaMa, as I have mentioned in #589, Storage SDK itself does not have HTTPS agent support, however, it respects `https_proxy` environment variable and uses the configured proxy server.

The other issue I was facing is that there was no way of using an agent to acquire a token via HTTPS. Therefore, I added [`setOptions`](https://github.com/dimiboi/azure-iot-sdk-node/blob/b02c2e1b8552a615a9579b599513f52d6d5fde3f/device/core/src/blob_upload/file_upload_api.ts#L40) function to `FileUploadApi` class, which sets the options of the HTTP transport the class uses. The options are propagated from the [device client](https://github.com/dimiboi/azure-iot-sdk-node/blob/b02c2e1b8552a615a9579b599513f52d6d5fde3f/device/core/src/device_client.ts#L116) via the [blob upload client](https://github.com/dimiboi/azure-iot-sdk-node/blob/b02c2e1b8552a615a9579b599513f52d6d5fde3f/device/core/src/blob_upload/blob_upload_client.ts#L74).

The flow is tested with the following tests:

[Here](https://github.com/dimiboi/azure-iot-sdk-node/blob/b02c2e1b8552a615a9579b599513f52d6d5fde3f/device/core/test/blob_upload/_file_upload_api_test.js#L56) is the test verifying that `FileUploadApi` class sets the underlying HTTP transport otions. 

[Here](https://github.com/dimiboi/azure-iot-sdk-node/blob/b02c2e1b8552a615a9579b599513f52d6d5fde3f/device/core/test/blob_upload/_blob_upload_client_test.js#L58) is the test verifying that `BlobUploadClient` sets the options of `FileUploadApi` class, which sets them on the underlying HTTP transport.

[Here](https://github.com/dimiboi/azure-iot-sdk-node/blob/b02c2e1b8552a615a9579b599513f52d6d5fde3f/device/core/test/_device_client_test.js#L337) is the test verifying that `DeviceClient` sets `BlobUploadClient` options, which sets `FileUploadApi` options, which in turn sets them on the underlying HTTP transport.

So when you expose `FileUploadApi` class, make sure it provides a function to set transport options, so that it is still possible to use HTTPS agent for proxy support.

device/core/package-lock.json
```diff
@@ -4,6 +4,101 @@
...
   "requires": true,
   "dependencies": {
+    "@azure/abort-controller": {
+      "version": "1.0.0-preview.1",
+      "resolved": "https://registry.npmjs.org/@azure/abort-controller/-/abort-controller-1.0.0-preview.1.tgz",
+      "integrity": "sha512-NnJqi6oHqt06Q2hz4nO1HO0QlyusBa3E/wezvn9flHEtl0IHYSmzGbtlb+MaAJ5GzxwqSevQ4q1+4B8fvVijOA==",
+      "requires": {
+        "tslib": "^1.9.3"
+      }
+    },
```
YoDaMa: remove this file from the PR

device/core/package.json
```diff
@@ -36,7 +36,7 @@
     "alltest": "istanbul cover ../../node_modules/mocha/bin/_mocha -- --reporter spec \"test/**/_*_test*.js\"",
     "ci": "npm -s run lint && npm -s run build && npm -s run alltest-min && npm -s run check-cover",
     "test": "npm -s run lint && npm -s run build && npm -s run unittest",
-    "check-cover": "istanbul check-coverage --statements 95 --branches 86 --lines 97 --functions 92"
```
YoDaMa: >97 [](start = 82, length = 2)

increase this back to 97

device/core/src/device_client.ts
```diff
@@ -171,13 +175,77 @@ export class Client extends InternalClient {
...
+        }
+        safeCallback(_callback, err, result);
+      });
+    }, callback);
+  }
+
+  /**
+   * @description      The `notifyBlobUploadStatus` method sends IoT Hub the result of a blob upload.
+   *
+   * @param {BlobUploadCommonResponseStub}   uploadResponse   The reponse received after a blob upload via the storage api is complete
```
YoDaMa: add the err to the comment

device/core/src/device_client.ts
```diff
@@ -171,13 +175,77 @@ export class Client extends InternalClient {
...
+   * @param {ErrorCallback}                  [callback]       Optional callback to call when the upload is complete.
+   * @returns {Promise<void> | void}                          Promise if no callback function was passed, void otherwise.
+   *
+   * @throws {ReferenceException} If uploadResponse is falsy.
+   */
+  notifyBlobUploadStatus(err: Error, uploadResponse: BlobUploadCommonResponseStub, callback: ErrorCallback): void;
+  notifyBlobUploadStatus(err: Error, uploadResponse: BlobUploadCommonResponseStub): Promise<void>;
+  notifyBlobUploadStatus(err: Error, uploadResponse: BlobUploadCommonResponseStub, callback?: ErrorCallback): Promise<void> | void {
+    return callbackToPromise((_callback) => {
+      /*Codes_SRS_NODE_DEVICE_CLIENT_41_XXX: [The `notifyBlobUploadStatus` method shall throw a `ReferenceError` if `uploadResponse` is falsy.]*/
```
YoDaMa: make sure these are numbered

YoDaMa: also see if these are reflected in tests

---
In reply to: [321486450](https://github.com/Azure/azure-iot-sdk-node/pull/610#discussion_r321486450) [](ancestors = 321486450)

device/core/src/interfaces.ts
```diff
@@ -138,3 +138,25 @@ export interface DeviceClientOptions extends X509 {
...
+ */
+export interface BlobUploadCommonResponseStub {
+  errorCode?: string;
+  _response?: {
+    status?: number;
+    bodyAsText?: string;
+  };
+}
+
+export interface RestErrorStub {
```
YoDaMa: add comment describing interface

device/core/src/blob_upload/blob_uploader.ts
```diff
@@ -23,11 +42,19 @@ export interface BlobService {
     createBlockBlobFromStream(containerName: string, blobName: string, stream: Stream, streamLength: number, done: (err: Error, body?: any, result?: BlobResponse) => void): void;
 }
 
+export interface Aborter {
```
YoDaMa: clean these up. Put private above them

device/core/src/blob_upload/blob_upload_client.ts
```diff
@@ -4,11 +4,11 @@
 'use strict';
 
 import { Stream } from 'stream';
-import { AuthenticationProvider, errorCallbackToPromise, ErrorCallback, TripleValueCallback, Callback } from 'azure-iot-common';
+import { AuthenticationProvider, errorCallbackToPromise, ErrorCallback } from 'azure-iot-common';
 
 import { BlobUploadResult } from './blob_upload_result';
-import { BlobUploader as DefaultBlobUploader, BlobResponse } from './blob_uploader';
-import { FileUploadApi as DefaultFileUploadApi } from './file_upload_api';
+import { BlobUploader as DefaultBlobUploader, BlobUploaderInterface as BlobUploader } from './blob_uploader';
```
YoDaMa: change this to IBlobUploader

device/core/src/blob_upload/blob_upload_client.ts
```diff
@@ -55,11 +39,11 @@ export interface BlobUpload {
  * @private
  */
 export class BlobUploadClient implements BlobUpload {
+  private _fileUploadApi: FileUploadInterface;
```
YoDaMa: IFileUpload

device/core/src/blob_upload/blob_upload_client.ts
```diff
@@ -55,11 +39,11 @@ export interface BlobUpload {
...
   private _authenticationProvider: AuthenticationProvider;
-  private _fileUploadApi: FileUpload;
   private _blobUploader: BlobUploader;
 
-  constructor(authenticationProvider: AuthenticationProvider, fileUploadApi?: FileUpload, blobUploader?: BlobUploader) {
+  constructor(authenticationProvider: AuthenticationProvider, fileUploadApi?: FileUploadInterface, blobUploader?: BlobUploader) {
```
YoDaMa: IFileUpload

device/core/src/blob_upload/blob_upload_result.ts
```diff
@@ -3,7 +3,7 @@
 
 'use strict';
 
-import { BlobResponse } from './blob_uploader';
+import { RestErrorStub } from './../interfaces';
```
YoDaMa: IRestErrorStub

device/core/test/blob_upload/fake_storage_api.js
```diff
@@ -3,19 +3,68 @@
...
+  });
+}
+
+
+
+// const blockBlobURL = new this.storageApi.BlockBlobURL(`https://${blobInfo.hostName}/${blobInfo.containerName}/${blobInfo.blobName}${blobInfo.sasToken}`, pipeline);
+// const uploadPromise = this.storageApi.uploadStreamToBlockBlob(this.storageApi.Aborter.timeout(30 * 60 * 1000), stream, blockBlobURL, streamLength, 20, { progress: ev => console.log(ev) });
+// uploadPromise
+
+// function createBlobServiceWithSas (host, sasToken) {
```
YoDaMa: delete this

device/core/test/blob_upload/fake_storage_api.js
```diff
@@ -3,19 +3,68 @@
...
+// StorageURL: any;
+// AnonymousCredential: any;
+// uploadStreamToBlockBlob: any;
+
+function AnonymousCredential() {};
+
+class Aborter {
+  constructor () {
+  }
+  static timeout() {
```
YoDaMa: make some sort of order?

device/core/test/blob_upload/_blob_upload_result_test.js
```diff
@@ -35,46 +35,41 @@ describe('BlobUploadResult', function() {
...
       assert.isTrue(result.isSuccess);
-      assert.equal(result.statusCode, testCode);
-      assert.equal(result.statusDescription, testBody);
+      assert.equal(result.statusCode, 201);
+      assert.equal(result.statusDescription, '');
+    });
+  
+    /*Tests_SRS_NODE_DEVICE_BLOB_UPLOAD_RESULT_16_007: [If the `err` argument is not `null` but body and response are `undefined` (non HTTP error), the `BlobUploadResult` error shall have the `statusCode` property set to -1.]*/
+    /*Tests_SRS_NODE_DEVICE_BLOB_UPLOAD_RESULT_16_008: [If the `err` argument is not `null` but body and response are `undefined` (non HTTP error), the `BlobUploadResult` error shall have the `statusDescription` property set to the error message.]*/
+    it('Creates a correct BlobUploadResult instance when response is a success', function() {
```
YoDaMa: change these requirement comments

device/core/src/blob_upload/blob_uploader.ts
```diff
@@ -5,9 +5,28 @@
...
+  sasToken?: string;
+  correlationId: string;
+}
+
+/**
+ * @private
+ */
+export interface BlobUploaderInterface {
+uploadToBlob(uploadParams: UploadParams, stream: Stream, streamLength: number, done: TripleValueCallback<any, BlobResponse>): void;
+uploadToBlob(uploadParams: UploadParams, stream: Stream, streamLength: number): Promise<any>;
```
pierreca: nit: indentation

device/core/src/blob_upload/blob_uploader.ts
```diff
@@ -23,11 +42,19 @@ export interface BlobService {
...
+}
+
 /**
  * @private
  */
 export interface StorageApi {
-    createBlobServiceWithSas(hostName: string, sasToken: string): BlobService;
+    BlockBlobURL: any;
+    Aborter: Aborter;
+    StorageURL: any;
```
YoDaMa: change these from any to something more specific

device/core/src/blob_upload/blob_upload_result.ts
```diff
@@ -40,30 +40,41 @@ export class BlobUploadResult {
...
+      const statusCode = err.hasOwnProperty('statusCode') ? err.statusCode : -1;
+      const statusDescription = err.hasOwnProperty('response') ? err.response : err.message;
       uploadResult = new BlobUploadResult(false, statusCode, statusDescription);
     } else {
-        if (response.statusCode >= 200 && response.statusCode < 300) {
-          uploadResult = new BlobUploadResult(true, response.statusCode, response.body);
-        } else {
-          uploadResult = new BlobUploadResult(false, response.statusCode, response.body);
-        }
+      if (!uploadResponse._response) throw new ReferenceError('if uploadResponse is provided, _response must be a member object');
```
pierreca: >throw [](start = 37, length = 5)

just want to make sure: we throw only when it's a developer error (ie a bug in our SDK or in the customer code). If it's something that could happen because a call to the service fails, either it shouldn't throw, or it should throw but we catch it and call the public api callback with the error.

device/core/src/blob_upload/blob_upload_result.ts
```diff
@@ -40,30 +40,41 @@ export class BlobUploadResult {
...
+  /*Codes_SRS_NODE_DEVICE_BLOB_UPLOAD_RESULT_16_009: [If `err` is null but `body` and `reponse` are provided, the `BlobUploadResult` error shall have the `isSuccess` property set to `true`.]*/
   /*Codes_SRS_NODE_DEVICE_BLOB_UPLOAD_RESULT_16_010: [If `err` is null but `body` and `reponse` are provided, the `BlobUploadResult` error shall have the `statusCode` property set to the HTTP status code of the blob upload response.]*/
   /*Codes_SRS_NODE_DEVICE_BLOB_UPLOAD_RESULT_16_011: [If `err` is null but `body` and `reponse` are provided, the `BlobUploadResult` error shall have the `statusDescription` property set to the HTTP response body of the blob upload response.]*/
   /*Codes_SRS_NODE_DEVICE_BLOB_UPLOAD_RESULT_16_012: [If `err` is null and `body` and/or `response` is/are falsy, `fromAzureStorageCallbackArgs` shall throw a `ReferenceError`.]*/
-  static fromAzureStorageCallbackArgs(err?: Error, body?: any, response?: BlobResponse): BlobUploadResult {
-    if (!err && (!body || !response)) throw new ReferenceError('if err is null, body and response must be supplied');
+  static fromAzureStorageCallbackArgs(err?: Error & RestErrorStub, uploadResponse?: any): BlobUploadResult {
```
pierreca: >any [](start = 84, length = 3)

should have a type

device/core/src/device_client.ts
```diff
@@ -39,10 +42,11 @@ export class Client extends InternalClient {
    * @param {string}  connStr           A connection string (optional: when not provided, updateSharedAccessSignature must be called to set the SharedAccessSignature token directly).
    * @param {Object}  blobUploadClient  An object that is capable of uploading a stream to a blob.
    */
```
anthonyvercolano: >*/ [](start = 3, length = 2)


You appear to be missing the description of fileUploadApi in the parameter description.

device/core/src/blob_upload/blob_upload_client.ts
```diff
@@ -124,4 +99,5 @@ export class BlobUploadClient implements BlobUpload {
       });
     }, done);
   }
+
```
anthonyvercolano: nit: There is no lint error from the blank line?

device/core/src/blob_upload/blob_upload_result.ts
```diff
@@ -40,30 +40,40 @@ export class BlobUploadResult {
...
     let uploadResult: BlobUploadResult;
+    if ((err && uploadResponse) || (!err && !uploadResponse)) throw new ReferenceError('either err or uploadResponse must be supplied, exclusively.');
```
anthonyvercolano: >either err or uploadResponse must be supplied, exclusively [](start = 88, length = 58)

[http://www.howtocreate.co.uk/xor.html](http://www.howtocreate.co.uk/xor.html)

device/core/src/blob_upload/blob_upload_result.ts
```diff
@@ -40,30 +40,40 @@ export class BlobUploadResult {
    *
    * @throws {ReferenceException} If err is null and either body or response are also falsy.
    */
+  // DISCUSS WITH PIERRE BECAUSE THESE BLOB REQUIREMENTS DONT MAKE SENSE AT ALL
```
anthonyvercolano: > DISCUSS WITH PIERRE BECAUSE THESE BLOB REQUIREMENTS DONT MAKE SENSE AT ALL [](start = 4, length = 75)

This ^^^^^^^^^^^^ really seems like something that should be done and deleted.

And as an aside, I like comments.  I would have written a bunch more comments in the below code.  If only for my own sanity while writing it.  It perhaps might have made more sense to put the requirements closer to the actual lines implementing them.

device/core/src/blob_upload/file_upload_api.ts
```diff
@@ -24,6 +24,21 @@ const packageJson = require('../../package.json');
  */
 /*Codes_SRS_NODE_FILE_UPLOAD_ENDPOINT_16_002: [`FileUploadApi` shall throw a `ReferenceError` if `deviceId` is falsy.]*/
 /*Codes_SRS_NODE_FILE_UPLOAD_ENDPOINT_16_003: [`FileUploadApi` shall throw a `ReferenceError` if `hostname` is falsy.]*/
+
+/**
+ * @private
+ */
+export interface FileUploadInterface {
```
anthonyvercolano: >export interface FileUploadInterface [](start = 0, length = 36)

Union types?

device/core/src/blob_upload/blob_upload_client.ts
YoDaMa: there is no lint error on this but it isn't pretty. removed it.

device/core/src/blob_upload/file_upload_api.ts
YoDaMa: What do you mean by Union types?

device/core/src/device_client.ts
```diff
@@ -171,13 +176,85 @@ export class Client extends InternalClient {
...
+    return callbackToPromise((_callback) => {
+      /*Codes_SRS_NODE_DEVICE_CLIENT_41_005: [The `notifyBlobUploadStatus` method shall throw a `ReferenceError` if `isSuccess` is falsy but not the boolean false.]*/
+      /*Codes_SRS_NODE_DEVICE_CLIENT_41_006: [The `notifyBlobUploadStatus` method shall throw a `ReferenceError` if `statusCode` is falsy but not the number 0.]*/
+      /*Codes_SRS_NODE_DEVICE_CLIENT_41_007: [The `notifyBlobUploadStatus` method shall throw a `ReferenceError` if `statusDescription` is falsy but not an empty string.]*/
+      if (!isSuccess && typeof(isSuccess) !== 'boolean' ) throw new ReferenceError('isSuccess cannot be \' ' + isSuccess + ' \'');
+      if (!statusCode && !(statusCode === 0)) throw new ReferenceError('statusCode cannot be \' ' + statusCode + ' \'');
+      if (!statusDescription && statusDescription !== '') throw new ReferenceError('statusDescription cannot be \' ' + statusDescription + ' \'.');
+      /*Codes_SRS_NODE_DEVICE_CLIENT_41_012: [The `notifyBlobUploadStatus` method shall throw a `ReferenceError` if `correlationId` is not set.]*/
+      if (!this._blobStorageUploadParams || !this._blobStorageUploadParams.correlationId) {
```
CIPop: See above: I think this should be a parameter. The app may need to manage multiple uploads: some larger (slower) and some smaller. It's ideal to allow the app to do these in parallel.

device/core/src/device_client.ts
```diff
@@ -171,13 +176,85 @@ export class Client extends InternalClient {
...
+      const retryOp = new RetryOperation(this._retryPolicy, this._maxOperationTimeout);
+      retryOp.retry((opCallback) => {
+        /*Codes_SRS_NODE_DEVICE_CLIENT_41_002: [The `getBlobSharedAccessSignature` method shall call the `getBlobSharedAccessSignature` method in the instantiated `_fileUploadApi` class and pass in `blobName` as a parameter.]*/
+        this._fileUploadApi.getBlobSharedAccessSignature(blobName, opCallback);
+      }, (err, result) => {
+        /*Codes_SRS_NODE_DEVICE_CLIENT_41_003: [The `getBlobSharedAccessSignature` method shall call the `_callback` callback with `err` and `result` from the call to `getBlobSharedAccessSignature`.]*/
+        if (!err) {
+          debug('got blob storage shared access signature.');
+          /*Codes_SRS_NODE_DEVICE_CLIENT_41_004: [The `getBlobSharedAccessSignature` method shall store the `result` value for use by the `notifyBlobUploadStatus` method.]*/
+          this._blobStorageUploadParams = result;
```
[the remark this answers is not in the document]
CIPop: Actually I think there's another bigger issue: Since it's overwritten, this won't work if the app calls `getBlobSharedAccessSignature` multiple times.
I think in that case it'll be able to notify only the very last one that won the overwrite.

device/core/test/_device_client_test.js
```diff
@@ -142,11 +142,163 @@ describe('Device Client', function () {
...
+      let FakeFileUploadApi = function (uploadResult) {
+        this.notifyUploadComplete = function(correlationId, uploadResult, callback) {
+          callback();
+        };
+      };
+      let client = new Client(new EventEmitter(), null, null, new FakeFileUploadApi());
+      client._blobStorageUploadParams = { correlationId: 'fakeCorrelationId' };
+      client.notifyBlobUploadStatus(isSuccess, statusCode, statusDescription, done);
+    });
+  });
```
CIPop: Please add two other tests: 
1. `it('calls getBlobSharedAccessSignature 10 times to upload 10 parallel files and succeeds notifying all 10 uploads in the same order')`
1. `it('calls getBlobSharedAccessSignature 10 times to upload 10 parallel files and succeeds notifying all 10 uploads in reverse order')`

PR #620, opened 2019-09-09
docs(provisioning): add reference to DPS docs

gjswalling:
point to docs for filling out the values '[provisioning host]' '[id scope]', '[registration id'], '[cert filename]', and '[key filename]'

# Checklist

# Description of the problem

# Description of the solution


provisioning/device/samples/readme.md
```diff
@@ -68,7 +68,7 @@ When you have a certificate change containing a root certificate and several opt
 
 At the very least, your root certificate needs to be uploaded to your Device Provisioning Service blade in the Azure portal, and it needs to be verified.
 
-Once the group is created and the certificates have been uploaded and verified, you can use [register_x509.js][register-x509] to register the device.  When running this sample, you'll need to replace the '[provisioning host]' '[id scope]', '[registration id'], '[cert filename]', and '[key filename]' values in the code with your specific values.
+Once the group is created and the certificates have been uploaded and verified, you can use [register_x509.js][register-x509] to register the device.  When running this sample, you'll need to replace the '[provisioning host]' '[id scope]', '[registration id'], '[cert filename]', and '[key filename]' values in the code with your specific values.  Please refer to guidance here for these values: https://docs.microsoft.com/en-us/azure/iot-dps/quick-create-simulated-device-x509-node#simulate-the-device
```
anthonyvercolano: >here for these values: https://docs.microsoft.com/en-us/azure/iot-dps/quick-create-simulated-device-x509-node#simulate-the-device [](start = 373, length = 129)

I think generally, the links are supposed to be down at the bottom of this file.  The links are the values of the [name or link "text"].  The [] thing goes here.  The hardlink below.

gjswalling: sounds good, just want to give some guidance about what values to use.  We could also copy the content here if you don't want to link to the simulated device section

YoDaMa: so I think what @anthonyvercolano is saying is you should replace your link. Instead of:
```
Please refer to guidance here for these values: https://docs.microsoft.com/en-us/azure/iot-dps/quick-create-simulated-device-x509-node#simulate-the-device
```

Do this:
```
For guidance on populating these values, refer [here][simulate-x509-device] https://docs.microsoft.com/en-us/azure/iot-dps/quick-create-simulated-device-x509-node#simulate-the-device
```
and then add the link to the bottom.

Appreciate you suggesting this! Love it when the community offers usability PRs

gjswalling: I tried moving the link to the bottom.  Please let me know if there's any further action on me here

Issue #615, opened 2019-09-03
[Digital Twins Device SDK] uncaught exception during registration


peolivei2:
Process crashes with the following exception when calling `register()`:

```
/app/node_modules/azure-iot-digitaltwins-device/dist/digital_twin_client.js:356
                throw new Error(interfaceInstanceName + ' is not registered');
                ^
 Error: urn_azureiot_Client_SDKInformation is not registered
    at /app/node_modules/azure-iot-digitaltwins-device/dist/digital_twin_client.js:356:23
    at Object.callbackToPromise (/app/node_modules/azure-iot-digitaltwins-device/node_modules/azure-iot-common/lib/promise_utils.js:33:16)
    at DigitalTwinClient._reportProperty (/app/node_modules/azure-iot-digitaltwins-device/dist/digital_twin_client.js:353:35)
    at Property.report (/app/node_modules/azure-iot-digitaltwins-device/dist/digital_twin_client.js:336:26)
    at /app/node_modules/azure-iot-digitaltwins-device/dist/digital_twin_client.js:490:49
    at /app/node_modules/azure-iot-device/node_modules/azure-iot-common/lib/retry_operation.js:57:21
    at /app/node_modules/azure-iot-device/lib/twin.js:76:25
    at MqttTwinClient._onResponseMessage (/app/node_modules/azure-iot-device-mqtt/lib/mqtt_twin_client.js:257:21)
    at MqttTwinClient._onMqttMessage (/app/node_modules/azure-iot-device-mqtt/lib/mqtt_twin_client.js:214:18)
    at emitTwo (events.js:131:20)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! simulation@1.0.2 start: `node --max_old_space_size=8192 ./dist/start.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the simulation@1.0.2 start script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
 npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2019-08-31T00_05_53_602Z-debug.log
```

Following the stack trace, the crash happens when the SDK tries to [report the SDK info](https://github.com/Azure/azure-iot-sdk-node/blob/716d6dec283e7367245e4858d742e678cfbcbdfa/digitaltwins/device/src/digital_twin_client.ts#L639).
[...]

PR #621, opened 2019-09-11
[digitaltwin] moving sdkInfo to client instance

peolivei2:
# Checklist

# Reference/Link to the issue solved with this PR
https://github.com/Azure/azure-iot-sdk-node/issues/615

# Description of the problem
`sdkInformation` was declared at module level but registered at the constructor, leading to a race condition whenever more than one client object was created.

# Description of the solution
Moved `sdkInformation` to a client instance field.


anthonyvercolano:
Carried over in #622

Issue #626, opened 2019-09-13
Nested Transport Obj on Promises


YoDaMa:
per APerrine's notes:

To repro
```javascript
require('dotenv').config()
const { Mqtt } = require('azure-iot-device-mqtt');
const { Client, Message } = require('azure-iot-device');

/* 
My least important question, but interesting...
why does the wrapping go transportObj : MessagEnqueued : transportObj??
(havent really investigated... maybe take out)
*/


const startMessageInterval = (client, messageDelay=3000) => setInterval( async() => {
    const asteroid = { size : Math.random()*100, timeTillImpact : Math.random()*100 } //leaving units to users imagination
    const msg = new Message( JSON.stringify(asteroid) )

    msg.properties.add('IMPACT_EMERGENCY', asteroid.timeTillImpact < 50 && asteroid.size > 50 ? 'true' : 'false')
    const transportObj = await client.sendEvent(msg).catch( err => console.error('Error sending message: ', err) )

    console.log(transportObj)

}, messageDelay)


const run = async() => {
    try {
        //create a client using the device connection string and the mqtt protocol
        const client = Client.fromConnectionString(process.env.DEVICE_CONN_STRING, Mqtt)
        const sendMessageInterval = startMessageInterval(client)


    } catch (err) {
        console.error('Error: ', err)
    }
}


run()
```

YoDaMa:
From my quick repro, this is the structure we get when setting a break point at `console.log(transportObj)`:
```typescript
MessageEnqueued
|---TransportObj 
      |---TransportObj
```

When I set a breakpoint internall in `mqtt.ts`:
```typescript
> MessageEnqueued {transportObj: Object}
[[StableObjectId]]:7
> transportObj:Object {cmd: "publish", topic: "devices/beta/messages/events/IMPACT_EMERGENCY=fals…", payload: "{"size":19.24501545677899,"timeTillImpact":90.1477…", …}
[[StableObjectId]]:9
cmd:"publish"
dup:false
messageId:8386
payload:"{"size":19.24501545677899,"timeTillImpact":90.14774673121259}"
qos:1
retain:false
topic:"devices/beta/messages/events/IMPACT_EMERGENCY=false"
```

az-iot-builder-01:
[...]

PR #627, opened 2019-09-14
[fix](azure-iot-device-mqtt): remove double nesting on sendEvent TransportObj

YoDaMa:
This is in reference to #626 

The wrapping of the response was 1) redundant and 2) out of line with how other MQTT calls are made. 

With this change an MQTT PUB and ACK look like this:
```
  azure-iot-mqtt-base:MqttBase CONNACK: {"cmd":"connack","retain":false,"qos":0,"dup":false,"length":2,"topic":null,"payload":null,"sessionPresent":false,"returnCode":0} +1ms
...
  azure-iot-device-mqtt:Mqtt PUBACK: {"cmd":"publish","topic":"devices/beta/messages/events/IMPACT_EMERGENCY=false","payload":"{\"size\":26.15078342292594,\"timeTillImpact\":22.72007917052823}","qos":1,"retain":false,"messageId":10181,"dup":false} +2s
```

Before (without this change):
```
  azure-iot-mqtt-base:MqttBase CONNACK: {"cmd":"connack","retain":false,"qos":0,"dup":false,"length":2,"topic":null,"payload":null,"sessionPresent":false,"returnCode":0} +1ms
...
  azure-iot-device-mqtt:Mqtt PUBACK: {"transportObj":{"cmd":"publish","topic":"devices/beta/messages/events/IMPACT_EMERGENCY=false","payload":"{\"size\":49.08919394567126,\"timeTillImpact\":57.049220217156815}","qos":1,"retain":false,"messageId":48866,"dup":false}} +6s
```

This resolves the unnecessary double nesting that occurred on message responses, which was a bug.

### DIFF
--- a/device/transport/mqtt/test/_mqtt_test.js
+++ b/device/transport/mqtt/test/_mqtt_test.js
@@ -295,6 +295,29 @@ describe('Mqtt', function () {
         fakemqtt.emit('connect', { connack: true });
       });
 
+      /*Tests_SRS_NODE_DEVICE_MQTT_41_004 [ The `sendEvent` method shall call its callback with a `MessageEnqueued` ]*/
+      /*Tests_SRS_NODE_DEVICE_MQTT_41_005 [ The `sendOutputEvent` method shall call its callback with a `MessageEnqueued` ]*/
+      it('calls the callback with the correctly structured response parameter', function (done) {
+        // There was a bug in the mqtt SendEvent code whereby the `response` would contain a doubly nested object.
+        // This was caused by creating a new MessagedEnqueued object, and then passing that in as a parameter
+        // to a callback that would use it as a parameter for creating a MessageEnqueued object, leading to a 
+        // odd double nesting. This does not seek to test the MessageEnqueued object itself, since that should
+        // be tested in the results unit testing, just to verify that the response is generated properly without
+        // double nesting.
+        fakeMqttBase.publish.callsArgWith(3, null, 'fakeResultObject');
+        var transport = new Mqtt(fakeAuthenticationProvider, fakeMqttBase);
+        transport.connect(function () {
+          testConfig.sendFunc(transport, new Message('message'), function(err, resp) {
+            assert.equal(resp.constructor.name, 'MessageEnqueued');
+            assert(resp.hasOwnProperty('transportObj'));
+            assert.equal(resp.transportObj,'fakeResultObject');
+            done();
+          });
+        });
+        fakemqtt.emit('connect', { connack: true });
+
+      });
+
       [
         /*Tests_SRS_NODE_COMMON_MQTT_BASE_16_011: [The `sendEvent` method shall serialize the `messageId` property of the message as a key-value pair on the topic with the key `$.mid`.]*/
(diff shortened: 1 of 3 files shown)


YoDaMa:
test needed.

device/transport/mqtt/test/_mqtt_test.js
```diff
@@ -295,6 +295,21 @@ describe('Mqtt', function () {
         fakemqtt.emit('connect', { connack: true });
       });
 
+      /*Tests_SRS_NODE_DEVICE_MQTT_41_004 [ The `sendEvent` method shall call its callback with a `MessageEnqueued` ]*/
+      /*Tests_SRS_NODE_DEVICE_MQTT_41_005 [ The `sendOutputEvent` method shall call its callback with a `MessageEnqueued` ]*/
+      it('calls the callback with the correctly structured response parameter', function (done) {
+        var transport = new Mqtt(fakeAuthenticationProvider, fakeMqttBase);
+        let foo = new results.MessageEnqueued();
```
anthonyvercolano: >let foo = new results.MessageEnqueued(); [](start = 8, length = 40)

This confuses me.  It doesn't appear to be used anywhere.  Further, don't use foo as a var name.

device/transport/mqtt/test/_mqtt_test.js
```diff
@@ -162,7 +162,7 @@ describe('Mqtt', function () {
       /*Tests_SRS_NODE_DEVICE_MQTT_18_045: [ The `sendOutputEvent` method shall connect the Mqtt connection if it is disconnected. ]*/
       it('connects the transport if currently disconnected', function (testCallback) {
         var transport = new Mqtt(fakeAuthenticationProvider, fakeMqttBase);
-        testConfig.sendFunc(transport, new Message('test'), function () {
+        testConfig.sendFunc(transport, new Message('test'), function (err, resp) {
```
anthonyvercolano: >err, resp [](start = 70, length = 9)

While this indeed is the function signature, you don't appear to be using either of these two arguments in the callback itself.

device/transport/mqtt/test/_mqtt_test.js
YoDaMa: good catch. left that in there.

device/transport/mqtt/test/_mqtt_test.js
```diff
@@ -295,6 +295,21 @@ describe('Mqtt', function () {
...
       });
 
+      /*Tests_SRS_NODE_DEVICE_MQTT_41_004 [ The `sendEvent` method shall call its callback with a `MessageEnqueued` ]*/
+      /*Tests_SRS_NODE_DEVICE_MQTT_41_005 [ The `sendOutputEvent` method shall call its callback with a `MessageEnqueued` ]*/
+      it('calls the callback with the correctly structured response parameter', function (done) {
+        var transport = new Mqtt(fakeAuthenticationProvider, fakeMqttBase);
+        let foo = new results.MessageEnqueued();
+        transport.connect(function () {
+          testConfig.sendFunc(transport, new Message('message'), function(err, resp) {
+            assert.equal(resp.__proto__.constructor.name, "MessageEnqueued");
```
anthonyvercolano: >assert.equal(resp.__proto__.constructor.name, "MessageEnqueued"); [](start = 12, length = 65)

Could this be better?   Is there another thing that could be compared here?  Do we  compare objects anywhere else in our unit tests and do it without simply checking the names?

PR #628, opened 2019-09-19
chore(multiplePackages) updated samples

Avsphere:
### What is different

Many of my samples are semantically similar, and I made sure to incorporate all of the methods used throughout.

The main difference is that the syntax / style / structure has been upgraded and samples like simple_sample_device_x509.js, simple_sample_device_with_sas.js, were compressed into device_connectionMethods.js as they only differed in a couple lines.

Some samples like dmpatterns_fwupdate_device I overhauled completely and incorporated its logic into the other samples.

I also added some new samples (sendFromDevice_pullFromEventHub and sendFromIotHub_listenOnDevice), these are little standalone scripts that demonstrate the concepts in action, I think they are nice but not realistic.

Happy for feedback, feel free to use or delete whatever

(Here is the original email chain that prompted this)
[RE_Updating_Azure_IoT_Node_SDK_Samples.zip](https://github.com/Azure/azure-iot-sdk-node/files/3633405/RE_Updating_Azure_IoT_Node_SDK_Samples.zip)


device/samples/package.json
```diff
@@ -1,22 +1,22 @@
...
+    "version": "0.0.1",
+    "private": true,
+    "description": "Simple samples demonstrating how to use the Azure IoT device SDK for Node.js",
+    "author": "Microsoft Corp.",
+    "license": "MIT",
+    "dependencies": {
+        "@azure/event-hubs": "^2.1.1",
+        "azure-iot-device": "^1.11.2",
+        "azure-iot-device-http": "^1.10.3",
+        "azure-iot-device-mqtt": "^1.10.2",
```
YoDaMa: we've gotta remove these carats from all the `azure-iot-*`. Because we have 'tight dependencies' on our published packages (i.e. we might break dependency between packages if they're not upgraded in lockstep), letting the minor or patch version float doesn't work.

anthonyvercolano:
At this point we've got to admit that we aren't going to do this.  Sorry.

PR #632, opened 2019-09-23
Normalizing Device Connection String envirnment variable between samples

jebrando:
# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/device/samples/simple_sample_device_x509.js
+++ b/device/samples/simple_sample_device_x509.js
@@ -4,10 +4,10 @@
 'use strict';
 
 var fs = require('fs');
-//var Protocol = require('azure-iot-device-mqtt').Mqtt;
+var Protocol = require('azure-iot-device-mqtt').Mqtt;
 // Uncomment one of these transports and then change it in fromConnectionString to test other transports
 // var Protocol = require('azure-iot-device-amqp').AmqpWs;
-var Protocol = require('azure-iot-device-http').Http;
+// var Protocol = require('azure-iot-device-http').Http;
 // var Protocol = require('azure-iot-device-amqp').Amqp;
 var Client = require('azure-iot-device').Client;
 var Message = require('azure-iot-device').Message;
--- a/device/samples/simple_sample_device_x509.js
+++ b/device/samples/simple_sample_device_x509.js
@@ -4,17 +4,17 @@
 'use strict';
 
 var fs = require('fs');
-var Protocol = require('azure-iot-device-mqtt').Mqtt;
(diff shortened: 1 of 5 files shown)


device/samples/simple_sample_device_x509.js
```diff
@@ -4,17 +4,17 @@
 'use strict';
 
 var fs = require('fs');
-var Protocol = require('azure-iot-device-mqtt').Mqtt;
+//var Protocol = require('azure-iot-device-mqtt').Mqtt;
 // Uncomment one of these transports and then change it in fromConnectionString to test other transports
 // var Protocol = require('azure-iot-device-amqp').AmqpWs;
-// var Protocol = require('azure-iot-device-http').Http;
+var Protocol = require('azure-iot-device-http').Http;
```
YoDaMa: Why are you setting the default to HTTP instead of MQTT?

device/samples/simple_sample_device_x509.js
```diff
@@ -4,17 +4,17 @@
...
 // var Protocol = require('azure-iot-device-amqp').Amqp;
 var Client = require('azure-iot-device').Client;
 var Message = require('azure-iot-device').Message;
 
 // String containing Hostname and Device Id in the following format:
 //  "HostName=<iothub_host_name>;DeviceId=<device_id>;x509=true"
-var deviceConnectionString = process.env.X509_DEVICE_CONNECTION_STRING; 
+var deviceConnectionString = process.env.DEVICE_CONNECTION_STRING;
```
YoDaMa: I'm wondering if normalizing this environment variable is actually the best idea, since the `;x509=true` postfix means that they are in effect not the same connection string (Device vs X509 Device).

device/samples/simple_sample_device_x509.js
jebrando: Because I was being sloppy.  I shall remove that and make it default to MQTT.

device/samples/simple_sample_device_x509.js
anthonyvercolano: I prefer one normalized name.  So removing the x509_ has my vote.

---
In reply to: [327191502](https://github.com/Azure/azure-iot-sdk-node/pull/632#discussion_r327191502) [](ancestors = 327191502)

jebrando: I see your point here, but since the device id needs to change between devices it is not realistic to have the same DEVICE_CONNECTION_STRING value between two devices.

PR #633, opened 2019-09-23
Add SeAsSecurityMessage interface and unit tests

kfbehar:
# Checklist

# Description of the problem

# Description of the solution


common/transport/amqp/test/_amqp_test.js
```diff
@@ -789,6 +789,35 @@ describe('Amqp', function () {
...
+      assert.doesNotThrow(() => {
+        amqp.connect({uri: 'uri'}, () => {
+          var fakeLink = new SenderLink('endpoint', undefined, fakeSession);
+          fakeLink.attach = this.stub().callsArg(0);
+          fakeLink.send = this.stub().callsArgWith(1, null, new results.MessageEnqueued());
+          this.stub(MockSenderLink, 'SenderLink').returns(fakeLink);
+          var msg = new Message('message');
+          msg.setAsSecurityMessage();
+          amqp.send(msg, 'endpoint', undefined, function() {
+            assert.equal(fakeLink.send.args[0][0]['message_annotations']['iothub-interface-id'], "urn:azureiot:Security:SecurityAgent:1");
```
anthonyvercolano: >"urn:azureiot:Security:SecurityAgent:1" [](start = 97, length = 39)

nit - single quotes.

device/transport/http/test/_http_test.js
```diff
@@ -140,6 +140,33 @@ describe('Http', function () {
...
+      msg.setAsSecurityMessage();
+
+      // act
+      transport.sendEvent(msg, function() {});
+
+      // assert
+      assert(spy.calledOnce);
+      assert.isOk(spy.args[0]);
+      assert.isOk(spy.args[0][2]);
+      assert.equal(spy.args[0][2]['iothub-interface-id'], "urn:azureiot:Security:SecurityAgent:1");
```
anthonyvercolano: >"urn:azureiot:Security:SecurityAgent:1" [](start = 58, length = 39)

nit - single rather than double.

device/transport/mqtt/src/mqtt.ts
```diff
@@ -916,7 +916,7 @@ export class Mqtt extends EventEmitter implements DeviceTransport {
     if (message.contentType) systemProperties['$.ct'] = <string>message.contentType;
     /*Codes_SRS_NODE_DEVICE_MQTT_16_083: [The `sendEvent` method shall serialize the `contentEncoding` property of the message as a key-value pair on the topic with the key `$.ce`.]*/
     if (message.contentEncoding) systemProperties['$.ce'] = <string>message.contentEncoding;
-
+    if (message.interfaceId) systemProperties['$.ifid'] = message.interfaceId;
```
anthonyvercolano: >['$.ifid'] [](start = 45, length = 10)

Does this clash with digitaltwins?

anthonyvercolano: There is a property sent in digitaltwins registration (which uses the telemetry path) that uses this as a property name.

---
In reply to: [327269387](https://github.com/Azure/azure-iot-sdk-node/pull/633#discussion_r327269387) [](ancestors = 327269387)

kfbehar: I was not aware of that, but we are already using that property in other SDKs, so I think it should be fine.
This is only in the context of a D2C message.

PR #656, opened 2019-11-19
chore (multiple): Adjust versions so hoisting is clean.

anthonyvercolano:
Component governance and make sure the the lerna --hoist is clean


longhaultests/package.json
```diff
@@ -22,7 +22,7 @@
   "homepage": "https://github.com/azure/azure-iot-sdk-node#readme",
   "devDependencies": {
     "@types/node": "^9.6.50",
-    "@types/uuid": "^3.4.5",
+    "@types/uuid": "^3.4.3",
```
jebrando: >"@types/uuid": "^3.4.3" [](start = 4, length = 23)

Why downgrade?

PR #661, opened 2019-11-23
test: move edge-e2e wrappers into Horton tree

BertKleewein:
Having the Horton wrappers inside the Horton tree is a much better model.  As part of a bigger effort to update the node support in Horton, I'm moving these back.


lerna.json
```diff
@@ -15,7 +15,6 @@
     "security/*",
     "longhaultests",
     "e2etests",
-    "edge-e2e/wrapper/nodejs-server-server",
```
BertKleewein: >"edge-e2e/wrapper/nodejs-server-server", [](start = 4, length = 40)

Question: do we also need to remove this from any version bump scripts?

anthonyvercolano: No.

PR #669, opened 2020-01-09
docs: Adding tls document describing what we support

jebrando:
# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/tls_protocol_version_and_ciphers.md
+++ b/tls_protocol_version_and_ciphers.md
@@ -1,0 +1,9 @@
+# IoT Node SDK support for TLS 1.2
+
+## TLS Version
+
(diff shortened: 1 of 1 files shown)


ericmitt:
typo: should be connection

ericmitt:
How a customer can force node to use tls1.2 ? Is it possible at 'node runtime' level like Java, or node just use what the OS provide? The doc should explain that.

jebrando:
This can not be done without a code change to the underlying TLS library.  So we are submitting a PR to the MQTT library that we're using.

---
In reply to: [572325484](https://github.com/Azure/azure-iot-sdk-node/pull/669#issuecomment-572325484) [](ancestors = 572325484)

PR #670, opened 2020-01-09
fix[azure-iot-device]: uploadToBlob large file support

YoDaMa:
This PR addresses a regression in the SDK that was introduced when the V2 Storage SDK was bumped to the V10 Storage SDK, for UploadToBlob. The calls have been updated to the V12 Storage SDK, which is the lastest version of the storage SDK, and have been validated to send beyond 100MB of data without error.


anthonyvercolano:
Consider adding a test that uploads (and deletes) a 256MB file.

PR #676, opened 2020-01-15
fix (azure-iothub): Added the property moduleContent #675.

anthonyvercolano:

Issue #675 reported that modulesContent was a mis-spelling.  Actually, what was necessary was for moduleContent needed to be there for module for devices.

### DIFF
--- a/service/src/configuration.ts
+++ b/service/src/configuration.ts
@@ -83,9 +83,14 @@ export interface ConfigurationMetrics {
 
 export interface ConfigurationContent {
   /**
-   * The configuration for all the modules.
+   * The configuration for edge modules.
    */
   modulesContent?: {[key: string]: Object};
+
+  /**
+   * The configuration for device modules
+   */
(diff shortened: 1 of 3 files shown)


service/src/configuration.ts
```diff
@@ -83,10 +83,15 @@ export interface ConfigurationMetrics {
...
   /**
-   * The configuration for all the modules.
+   * The configuration for edge modules.
    */
   modulesContent?: {[key: string]: Object};
 
+  /**
+   * The configuration for device modules
+   */
+  moduleContent?: {[key: string]: Object};
```
BertKleewein: >moduleContent [](start = 2, length = 13)

<eye roll> seriously?  for edge modules, it's plural, for iothub modules, it's singular?  But the plural name still points to a singular object?  (I know this wasn't your choice.  I'm just making sure the name is as painfully dumb as it appears).

PR #691, opened 2020-02-05
chore(azure-iot-device): Modify Edge keyword from comments

jebrando:
# Checklist

# Description of the problem

# Description of the solution

### DIFF
--- a/device/core/src/module_client.ts
+++ b/device/core/src/module_client.ts
@@ -190,8 +190,8 @@ export class ModuleClient extends InternalClient {
   }
 
   /**
-   * Invokes a method on a downstream device or on another module on the same Edge device. Please note that this feature only works when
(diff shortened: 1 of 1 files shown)


device/core/src/module_client.ts
```diff
@@ -190,8 +190,8 @@ export class ModuleClient extends InternalClient {
   }
 
   /**
-   * Invokes a method on a downstream device or on another module on the same Edge device. Please note that this feature only works when
-   * the module is being run as part of an Edge device.
+   * Invokes a method on a downstream device or on another module on the same IoTEdge device. Please note that this feature only works when
+   * the module is being run as part of an IoTEdge device.
```
anthonyvercolano: I like this fix.  If you search the tree for the " Edge" you might find a few more.

PR #694, opened 2020-02-06
chore(multiple): update to use nyc rather than istanbul

anthonyvercolano:
<


digitaltwins/device/package.json
```diff
@@ -20,21 +20,53 @@
     "chai": "^4.2.0",
     "es5-shim": "^4.5.12",
     "sinon": "^7.4.1",
-    "istanbul": "^0.4.5",
```
zolvarga: >istanbul [](start = 5, length = 8)

Just curious: You removed istambul, so no more unittest coverage?

Issue #657, opened 2019-11-19
[module client] ModuleClient.fromSharedAccessSignature() doesn't work


peolivei2:
For module client, the transport relies on the auth provider to get the moduleId and send messages to the correct module.
When using a connection string this works fine, as the connection string auth provider has support for moduleId.
However, it doesn't work when building the module client from a shared access signature, as the SAS auth provider does not support moduleId.

References:
[ModuleClient.fromSharedAccessSignature](https://github.com/Azure/azure-iot-sdk-node/blob/master/device/core/src/module_client.ts#L318)
[Code in SAS auth provider with no support for moduleId](https://github.com/Azure/azure-iot-sdk-node/blob/master/device/core/src/sas_authentication_provider.ts#L82)
[Code in MQTT transport that relies on moduleId from the auth provider](https://github.com/Azure/azure-iot-sdk-node/blob/master/device/transport/mqtt/src/mqtt.ts#L668)

anthonyvercolano:
Released.

az-iot-builder-01:
@peolivei2, thank you for your contribution to our open-sourced project! Please help us improve by filling out this 2-minute [customer satisfaction survey](https://forms.office.com/Pages/ResponsePage.aspx?id=v4j5cvGGr0GRqy180BHbRx8jpG79OudPqY-BojcKjkJUMFFBQjgxNkw5ODJIRjBWTzI0Nk85R0pUQiQlQCN0PWcu)

PR #703, opened 2020-02-12
fix(azure-iot-device): security provider should deal with module SAS

anthonyvercolano:
The SAS security provider should propagate through the moduleId if the SAS has one.

#657


device/core/src/sas_authentication_provider.ts
```diff
@@ -79,9 +79,12 @@ export class SharedAccessSignatureAuthenticationProvider extends EventEmitter im
     const sas: SharedAccessSignature = SharedAccessSignature.parse(sharedAccessSignature);
     const decodedUri = decodeURIComponent(sas.sr);
     const uriSegments = decodedUri.split('/');
+    const uriDeviceId = (uriSegments.length >= 2 && uriSegments[1] === 'devices') ? (uriSegments[2]) : null;
+    const uriModuleId = (uriSegments.length >= 4 && uriSegments[3] === 'modules') ? (uriSegments[4]) : null;
     const credentials: TransportConfig = {
```
jebrando: Test please.

PR #705, opened 2020-02-18
chore: update build script to use newer versions of node.  Same for l…

anthonyvercolano:
…ooper build script


YoDaMa:
>   displayName: Windows Build - Node 8

Node 12.

---
Refers to: vsts/node-nightly-df.yaml:61 in 5b4e51b. [](commit_id = 5b4e51b39074f76230fc835061c44eadca256944, deletion_comment = False)

anthonyvercolano:
Can't move windows for now because of the TPM support.  Will come back later.

PR #707, opened 2020-02-26
feat(multiple): add an option to set token renewal values for shared …

anthonyvercolano:
…access key authentication

Added the option at the device client level.  Adjusted the transports to validate the options.
Adjusted the SAK authentication provider to have a new method to set the renewal values.  If the
authentication provider is NOT already "running" then just set the values.  Otherwise set the values
and cause a new token event available to be emitted.

### DIFF
--- a/common/core/src/authentication_provider.ts
+++ b/common/core/src/authentication_provider.ts
@@ -26,4 +26,5 @@ export interface AuthenticationProvider {
   type: AuthenticationType;
   getDeviceCredentials(callback: Callback<TransportConfig>): void;
   getDeviceCredentials(): Promise<TransportConfig>;
-}
+  setTokenRenewalValues?(tokenValidTimeInSeconds: number, tokenRenewalMarginInSeconds: number): void;
+  }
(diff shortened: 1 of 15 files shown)


common/core/src/authentication_provider.ts
```diff
@@ -26,4 +26,5 @@ export interface AuthenticationProvider {
   type: AuthenticationType;
   getDeviceCredentials(callback: Callback<TransportConfig>): void;
   getDeviceCredentials(): Promise<TransportConfig>;
-}
+  setTokenRenewalValues?(tokenValidTimeInSeconds: number, tokenRenewalMarginInSeconds: number): void;
```
zolvarga: >tokenRenewalMarginInSeconds [](start = 58, length = 27)

Nitpick: I don't really like the name "Margin" for this. I think simple "tokenRenewalTimeInSeconds" would do it.

anthonyvercolano: It mimics the name used in the constructor.

PR #712, opened 2020-03-06
docs: update LTS lifetime and maintenance length

anthonyvercolano:
Make the maintenance length 12 months and then 6 months longer before removal.  Also note when the next LTS should be released.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -7,17 +7,17 @@
 
 # Getting Started
 
-### Why Azure IoT? 
+### Why Azure IoT?
 
-**[Click here][iot-dev-center]** to learn how Azure IoT can empower the digital transformation of your organization. 
+**[Click here][iot-dev-center]** to learn how Azure IoT can empower the digital transformation of your organization.
 
 ### New to Azure IoT Hub?
 
-**[Introduction to Azure IoT Hub:][iot-device-ecosystem]** Follow this guide to learn how to set up an IoT Hub and IoT Hub devices. 
+**[Introduction to Azure IoT Hub:][iot-device-ecosystem]** Follow this guide to learn how to set up an IoT Hub and IoT Hub devices.
 
-### New to the Azure IoT Node.JS SDK? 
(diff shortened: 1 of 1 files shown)


YoDaMa:
There's no statement in the readme that LTS will be 12 months of maintenance and then after that it will be removed in 6 months. We should make that explicit.

readme.md
```diff
@@ -107,10 +107,12 @@ Below is a table showing the mapping of the LTS branches to the package versions
 | NPM Package | Github Branch | LTS Status | LTS Start Date | Maintenance End Date | Removed Date |
 | :-----------: | :-----------: | :--------: | :------------: | :------------------: | :----------: |
 | 1.7.x         | [lts_09_2018](https://github.com/Azure/azure-iot-sdk-node/tree/lts_09_2018)   | Deprecated | 2018-09-16     | 2019-09-16           | 2020-03-09   |
-| 1.11.x        | [lts_09_2019](https://github.com/Azure/azure-iot-sdk-node/tree/lts_09_2019)   | Active     | 2019-09-09     | 2020-03-09           | 2020-09-19   |
+| 1.11.x        | [lts_09_2019](https://github.com/Azure/azure-iot-sdk-node/tree/lts_09_2019)   | Active     | 2019-09-09     | 2020-09-09           | 2021-03-09   |
 
 <sup>1</sup> All scheduled dates are subject to change by the Azure IoT SDK team.
 
+The next LTS branch will be released late March or early April 2020.
```
jebrando: If we have this line, it locks us into remembering to update this after every LTS release to talk about the next one.  I'm just worried that things will change and late March or early April will become late April.  Can you make this more generic and not put specific dates.

anthonyvercolano:
The table itself I thought made it clear.  But... Let me see what I can do.

---
In reply to: [595982046](https://github.com/Azure/azure-iot-sdk-node/pull/712#issuecomment-595982046) [](ancestors = 595982046)

PR #709, opened 2020-03-04
feat(azure-iot-service): add function to import/export by identity

jebrando:
# Checklist
- [ ] I have read the [contribution guidelines] (https://github.com/Azure/azure-iot-sdk-node/blob/master/.github/CONTRIBUTING.md).
- [ ] I added or modified the existing tests to cover the change (we do not allow our test coverage to go down).
- If this is a modification that impacts the behavior of a public API
  - [ ] I edited the corresponding document in the `devdoc` folder and added or modified requirements.

# Reference/Link to the issue solved with this PR (if any)

# Description of the problem

# Description of the solution

PR #684, opened 2020-01-31
chore: adding conventional commits enforcement

YoDaMa:
This utilizes a few tools:

https://github.com/commitizen/cz-cli
https://github.com/conventional-changelog/commitlint
https://github.com/typicode/husky#readme

PR #718, opened 2020-03-13
release(2020-03-13): bump package versions

anthonyvercolano:
- feat(azure-iot-service): add function to import/export by identity (#709)
 
   Enables new security functionality to specify how hub should authenticate when working with a storage account when running import/export devices jobs. Initially, this feature has limited availability, starting with 3 public cloud Azure regions. Since it is not yet widely available, this functionality must be explicitly enabled by the user.  To enable this functionality, this version of the SDK will require setting an environment variable of EnableStorageIdentity to 1. Once this feature is widely available, another SDK version will be released with this requirement removed.
 - docs: update LTS lifetime and maintenance length
 - feat(multiple): add an option to set token renewal values for shared access key authentication
 - chore: update debug looper to use node 8 for windows and 10 for ubuntu (#706)
 - fix(azure-iot-device): security provider should deal with module SAS #703
 - fix: patch unhandled promise
 - fix: remove pretty-print (#700)
 - chore(multiple): add script to perform a package-lock.json refresh
 - chore: remove the dependency for istanbul
 - chore: update dependencies
 - chore(azure-iot-device): Modify Edge keyword from comments
 - chore: istanbul to nyc (#685)
 - chore: adding conventional commits enforcement (#684)


YoDaMa:
Looks good. Conventional Commits makes the standardized release notes much easier to compile and understand 😃

PR #727, opened 2020-03-20
feat(azure-iothub): update service client and samples

zolvarga:
# Checklist

# Description of the problem

# Description of the solution


jebrando:
> // Copyright (c) Microsoft. All rights reserved.

This sample is interesting as far as should we have it and what should it do?  I believe we should get Elena involved just to tell her what we've done and she can guide us in what this sample should be.  Don't let this block the check in.

---
Refers to: digitaltwins/samples/service/digital_twin_full_sample.js:1 in 748b3f1. [](commit_id = 748b3f1dca0222e0a57a895527d193b46a175ae6, deletion_comment = False)

PR #781, opened 2020-05-04
refactor: rename deviceCapabilityModel to modelId

YoDaMa:


device/core/src/interfaces.ts
```diff
@@ -133,7 +133,7 @@ export interface DeviceClientOptions extends X509 {
    * !!Digital Twin Use Only!!
    * String used in MQTT username settting the Digital Twin Device Capability Model.
```
anthonyvercolano, line 134: > Digital Twin Device Capability Model [](start = 46, length = 37)

Probably want to change the comment too.

device/transport/mqtt/test/_mqtt_test.js
```diff
@@ -637,9 +637,9 @@ describe('Mqtt', function () {
     });
 
     /* Tests_SRS_NODE_DEVICE_MQTT_41_XXX: [For a Plug and Play Device the device capability model should be included as &digital-twin-model-id=<DEVICE’s DCM ID> after the api-version]*/
```
anthonyvercolano, line 639: >evice capability model [](start = 75, length = 22)

Change the comment.  Don't think this spec comment was actually in the .ts file.

digitaltwins/device/test/_digital_twin_client_test.js
```diff
@@ -51,9 +51,9 @@ describe('Digital Twin Client', () => {
       });
     });
 
-    /* Tests_SRS_NODE_DIGITAL_TWIN_DEVICE_41_002: [Will throw `ReferenceError` if the fromConnectionString method `capabilityModel` argument is falsy.] */
+    /* Tests_SRS_NODE_DIGITAL_TWIN_DEVICE_41_002: [Will throw `ReferenceError` if the fromConnectionString method `modelId` argument is falsy.] */
     [undefined, null, ''].forEach(function (falsyCapabilityModel) {
```
anthonyvercolano, line 55: >CapabilityMode [](start = 49, length = 14)

Change this.

PR #797, opened 2020-05-13
refactor(multipe packages): c2d states no longer using booleans

YoDaMa:
This PR removes the 'proxy' boolean that essentially guesses if the C2D link / subscription is set up or not. Instead of trying to guess every corner case possible, we offload the work to the transport, that will know if the C2D link is enabled or not because it controls that code. 

All the device client now checks is whether the user wants C2D or not. If the user wants C2D, then we enable it as a feature, otherwise it is disabled as a feature. If it is enabled as a feature, it will be attempted to be enabled using the disconnectHandler and when a newListener is added. In the event that multiple newListeners are added, they will all call to _enableC2D, however after the first enable the calls will just do nothing and return, with a debug log generated saying nothing happened.


device/transport/amqp/src/amqp.ts
```diff
@@ -414,26 +414,47 @@ export class Amqp extends EventEmitter implements DeviceTransport {
...
+            // let listenerCounts = [0, 0];
+            // if (this._c2dLink) {
+            //   listenerCounts[0] = this._c2dLink.listenerCount('message');
+            //   listenerCounts[1] = this._c2dLink.listenerCount('error');
+            // }
+            // if (listenerCounts[0] === 1 && listenerCounts[1] === 1) {
+            //   callback(true);
+            // } else {
+            //   callback(false);
+            // }
```
anthonyvercolano, line 427: Not sure what is going on with the commented code.

device/transport/mqtt/devdoc/mqtt_requirements.md
```diff
@@ -350,6 +358,8 @@ The `updateTwinReportedProperties` method is used to retrieve the device twin.
 
 **SRS_NODE_DEVICE_MQTT_16_046: [** `disableMethods` shall call its callback with an `Error` if an error is received while unsubscribing. **]**
 
+**SRS_NODE_DEVICE_MQTT_41_012: [** `disableMethods` shall unsubscribe from the topic for C2D messages only if it is currently subscribed. **]**
```
anthonyvercolano, line 361: >topic for C2D messages [](start = 79, length = 22)

Should this be topic for methods?

device/transport/mqtt/devdoc/mqtt_requirements.md
```diff
@@ -370,6 +380,8 @@ The `updateTwinReportedProperties` method is used to retrieve the device twin.
 
 **SRS_NODE_DEVICE_MQTT_18_067: [** `disableInputMessages` shall call its callback with an `Error` if an error is received while unsubscribing. **]**
 
+**SRS_NODE_DEVICE_MQTT_41_013: [** `disableInputMessages` shall unsubscribe from the topic for C2D messages only if it is currently subscribed. **]**
```
anthonyvercolano, line 383: >topic for C2D messages [](start = 85, length = 22)

Should this be topic input messages?

PR #803, opened 2020-05-14
Update patch to follow the Model

rido-min:
The update patch must use the same component names as the rest of the samples.


digitaltwins/samples/service/javascript/update_digital_twin.js
```diff
@@ -32,7 +32,7 @@ async function main() {
     //     { op: 'remove', path: '/newThermostat' },
     //     { op: 'replace', path: '/newThermostat', value: { tempSetpoint: 42 } },
```
anthonyvercolano, line 33: Update the comment.

PR #821, opened 2020-06-16
chore: node pnp sample

olivakar:
The PNP sample for node following a special client convention.

### DIFF
--- a/digitaltwins/samples/device/javascript/package.json
+++ b/digitaltwins/samples/device/javascript/package.json
@@ -21,7 +21,8 @@
     "azure-iot-device": "1.12.0-pnp-refresh.0",
     "azure-iot-common": "1.12.0-pnp-refresh.0",
     "azure-iot-device-mqtt": "1.12.0-pnp-refresh.0",
-    "azure-iot-digitaltwins-device": "1.0.0-pnp-refresh.0"
+    "azure-iot-digitaltwins-device": "1.0.0-pnp-refresh.0",
+    "util": "^0.12.3"
   },
   "devDependencies": {
     "eslint": "^6.8.0",
--- a/digitaltwins/samples/device/javascript/pnpSampleDevice.js
+++ b/digitaltwins/samples/device/javascript/pnpSampleDevice.js
@@ -57,7 +57,7 @@ const commandHandler = async (request, response) => {
     await sendCommandResponse(request, response, 200, 'turn Off response');
     break;
   }
-  default: // Is there no way to register for an unknown method?
+  default:
     await sendCommandResponse(request, response, 404, 'unknown method');
     break;
   }
@@ -159,7 +159,6 @@ async function main() {
     client.onDeviceMethod(commandNameBlink, commandHandler);
     client.onDeviceMethod(commandNameTurnOn, commandHandler);
     client.onDeviceMethod(commandNameTurnOff, commandHandler);
-    // Is there no way to register for an unknown method ?
 
     // Send Telemetry every 5.5 secs
     let index = 0;
--- a/digitaltwins/samples/device/javascript/pnpSampleDevice.js
+++ b/digitaltwins/samples/device/javascript/pnpSampleDevice.js
@@ -6,7 +6,6 @@
 const Protocol = require('azure-iot-device-mqtt').Mqtt;
 const Client = require('azure-iot-device').Client;
 const Message = require('azure-iot-device').Message;
-// const { promisify } = require('util')
 
 const packageJson = require('./package.json');
 
@@ -21,6 +20,7 @@ const deviceInfoComponentName = 'deviceInformation';
 const sdkInfoComponentName = 'sdkInformation';
 const componentPrefix = '$iotin:';
 const commandComponentCommandNameSeparator = '*';
+let intervalToken;
 
 const commandNameBlink = componentPrefix + sensorComponentName + commandComponentCommandNameSeparator + 'blink';
 const commandNameTurnOn = componentPrefix + sensorComponentName + commandComponentCommandNameSeparator + 'turnOn';
@@ -119,27 +119,42 @@ const helperAttachHandlerForDesiredPropertyPatches = (deviceTwin, componentName)
   });
 };
 
+const helperAttachExitListener = async (deviceClient) => {
+  const standardInput = process.stdin;
+  standardInput.setEncoding('utf-8');
 
-const sendTelemetryAtInterval = (deviceClient, componentName, secs) => {
-  let index = 0;
-  const interval = parseFloat(secs)*1000;
-  setInterval(() => {
-    console.log('Sending telemetry message %d...', index);
-    const data = JSON.stringify({ temp: 1 + (Math.random() * 90), humidity: 1 + (Math.random() * 99) });
-    const pnpMsg = new Message(data);
-    pnpMsg.properties.add(messageSubjectProperty, componentName);
-    pnpMsg.contentType = 'application/json';
-    pnpMsg.contentEncoding = 'utf-8';
-    deviceClient.sendEvent(pnpMsg);
-    index += 1;
-  }, interval);
+  // Prompt user to input data in console.
+  console.log('Please enter q or Q to exit sample.');
+  standardInput.on('data', (data) => {
+  // standard_input.on('data', async function (data) {
+    if (data === 'q\n' || data === 'Q\n') {
+      console.log('Clearing intervals and exiting sample.');
+      clearInterval(intervalToken);
+      // await deviceClient.close();
+      deviceClient.close();
+      process.exit();
+    } else {
+      console.log('User Input was : ' + data);
+      console.log('Please only enter q or Q to exit sample.');
+    }
+  });
 };
+async function sendTelemetry(deviceClient, componentName, index) {
+  console.log('Sending telemetry message %d...', index);
+  const data = JSON.stringify({ temp: 1 + (Math.random() * 90), humidity: 1 + (Math.random() * 99) });
+  const pnpMsg = new Message(data);
+  pnpMsg.properties.add(messageSubjectProperty, componentName);
+  pnpMsg.contentType = 'application/json';
+  pnpMsg.contentEncoding = 'utf-8';
+  await deviceClient.sendEvent(pnpMsg);
+}
 
 async function main() {
   // fromConnectionString must specify a transport, coming from any transport package.
   const client = Client.fromConnectionString(deviceConnectionString, Protocol);
   console.log('Connecting using connection string ' + deviceConnectionString);
   let resultTwin;
+
   try {
     // Add the modelId here
     await client.setOptions({ modelId: modelId });
@@ -151,7 +166,11 @@ async function main() {
     // Is there no way to register for an unknown method ?
 
     // Send Telemetry every 5.5 secs
-    await sendTelemetryAtInterval(client, sensorComponentName, 5.5);
+    let index = 0;
+    intervalToken = setInterval(() => {
+      sendTelemetry(client, sensorComponentName, index).catch((err) => console.log('error', err.toString()));
+      index += 1;
+    }, 5500);
 
     try {
       resultTwin = await client.getTwin();
@@ -183,11 +202,16 @@ async function main() {
       updateComponentReportedProperties(resultTwin, patchSDKInfo, sdkInfoComponentName);
       updateComponentReportedProperties(resultTwin, patchSensorInfo, sensorComponentName);
       helperAttachHandlerForDesiredPropertyPatches(resultTwin, sensorComponentName);
+
+      helperAttachExitListener(client);
     } catch (err) {
       console.error('could not retrieve twin or report twin properties\n' + err.toString());
     }
   } catch (err) {
     console.error('could not connect pnp client\n' + err.toString());
+  } finally {
+    console.log('entering and leaving the finally block');
+    // await client.close();
   }
 }
 
--- a/digitaltwins/samples/device/javascript/pnpSampleDevice.js
+++ b/digitaltwins/samples/device/javascript/pnpSampleDevice.js
@@ -6,6 +6,7 @@
 const Protocol = require('azure-iot-device-mqtt').Mqtt;
 const Client = require('azure-iot-device').Client;
 const Message = require('azure-iot-device').Message;
+// const { promisify } = require('util')
 
 const packageJson = require('./package.json');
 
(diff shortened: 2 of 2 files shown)


digitaltwins/samples/device/javascript/pnpSampleDevice.js
```diff
@@ -0,0 +1,202 @@
...
+      }
+    )
+  updateComponentReportedProperties(deviceTwin, patch, componentName);
+  console.log('updated the property');
+};
+
+
+const commandHandler = async (request, response) => {
+  switch (request.methodName) {
+  case commandNameBlink: {
```
jebrando, line 46: Add a console.log here just so users can see that blink happened.

digitaltwins/samples/device/javascript/pnpSampleDevice.js
```diff
@@ -0,0 +1,202 @@
...
+    await sendCommandResponse(request, response, 200, 'turn Off response')
+    break;
+  }
+  default: // Is there no way to register for an unknown method?
+    await sendCommandResponse(request, response, 404, 'unknown method')
+    break;
+  }
+};
+
+// TODO Not sure if this needs to be async or not coz we probably dont care if the response rached Hub.
```
jebrando, line 64: We need to figure this out and then remove the todo comment.

digitaltwins/samples/device/javascript/pnpSampleDevice.js
```diff
@@ -0,0 +1,202 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
anthonyvercolano, line 2: Shouldn't this file live in the master branch in the device/samples directory?

digitaltwins/samples/device/javascript/pnpSampleDevice.js
olivakar: the send method actually calls a helper which logs out the command.

digitaltwins/samples/device/javascript/pnpSampleDevice.js
```diff
@@ -0,0 +1,194 @@
...
+    await sendCommandResponse(request, response, 404, 'unknown method');
+    break;
+  }
+};
+
+// TODO Not sure if this needs to be async or not coz we probably dont care if the response rached Hub.
+const sendCommandResponse = async (request, response, status, payload) => {
+  helperLogCommandRequest(request);
+  try {
+    await response.send(status, payload);
```
anthonyvercolano, line 68: >response.send [](start = 10, length = 13)

This is written to be support callbacks or promises so it is perfectly reasonable for you to use await in this async function.

digitaltwins/samples/device/javascript/pnpSampleDevice.js
```diff
@@ -0,0 +1,194 @@
...
+  }
+  default: // Is there no way to register for an unknown method?
+    await sendCommandResponse(request, response, 404, 'unknown method');
+    break;
+  }
+};
+
+// TODO Not sure if this needs to be async or not coz we probably dont care if the response rached Hub.
+const sendCommandResponse = async (request, response, status, payload) => {
+  helperLogCommandRequest(request);
```
anthonyvercolano, line 66: >helperLogCommandRequest(request); [](start = 2, length = 33)

Nit: I would have put this in the command handler.  No biggy though.

digitaltwins/samples/device/javascript/pnpSampleDevice.js
```diff
@@ -0,0 +1,194 @@
...
+      propertyUpdateHandler(deviceTwin, componentName, propertyName, null, propertyValue.value, deviceTwin.properties.desired[versionProperty]);
+    });
+  });
+};
+
+
+const sendTelemetryAtInterval = async (deviceClient, componentName, secs) => {
+  let index = 0;
+  const interval = parseFloat(secs)*1000;
+  setInterval( async () => {
```
anthonyvercolano, line 126: >async () [](start = 15, length = 8)

This is what I think we're not supposed to do.  I don't believe that setInterval is built to handle a promise as its first argument.  The async means that this () => "returns" a promise.

digitaltwins/samples/device/javascript/pnpSampleDevice.js
```diff
@@ -0,0 +1,194 @@
...
+    console.log('Sending telemetry message %d...', index);
+    const data = JSON.stringify({ temp: 1 + (Math.random() * 90), humidity: 1 + (Math.random() * 99) });
+    const pnpMsg = new Message(data);
+    pnpMsg.properties.add(messageSubjectProperty, componentName);
+    pnpMsg.contentType = 'application/json';
+    pnpMsg.contentEncoding = 'utf-8';
+    await deviceClient.sendEvent(pnpMsg);
```
anthonyvercolano, line 133: >await deviceClient.sendEvent(pnpMsg) [](start = 4, length = 36)

However, here you can use a .then instead of the await.

digitaltwins/samples/device/javascript/pnpSampleDevice.js
```diff
@@ -0,0 +1,194 @@
...
+    index += 1;
```
anthonyvercolano, line 134: > index += 1 [](start = 3, length = 11)

You are not doing anything with this variable.

digitaltwins/samples/device/javascript/pnpSampleDevice.js
```diff
@@ -0,0 +1,194 @@
...
+        brightness: '10'
+      }, sensorComponentName);
+
+      // the below things can only happen once the twin is there
+      updateComponentReportedProperties(resultTwin, patchDeviceInfo, deviceInfoComponentName);
+      updateComponentReportedProperties(resultTwin, patchSDKInfo, sdkInfoComponentName);
+      updateComponentReportedProperties(resultTwin, patchSensorInfo, sensorComponentName);
+      helperAttachHandlerForDesiredPropertyPatches(resultTwin, sensorComponentName);
+    } catch (err) {
+      console.error('could not get twin\n' + err.toString());
```
anthonyvercolano, line 187: >console.error('could not get twin\n' + err.toString()); [](start = 6, length = 55)

Isn't this going to catch the other errors that and happen in the reporting of properties?

digitaltwins/samples/device/javascript/pnpSampleDevice.js
anthonyvercolano, line 126: >setInterval( [](start = 2, length = 12)

As an alternative to using setInterval you could use the promisified version of setTimeout that was used in the class example and simply perform the sending inside of a loop whose last statement is the await sleep (2000)?

digitaltwins/samples/device/javascript/pnpSampleDevice.js
```diff
@@ -0,0 +1,194 @@
...
+    await client.setOptions({ modelId: modelId });
+    await client.open();
+    console.log('Enabling the commands on the client');
+    client.onDeviceMethod(commandNameBlink, commandHandler);
+    client.onDeviceMethod(commandNameTurnOn, commandHandler);
+    client.onDeviceMethod(commandNameTurnOff, commandHandler);
+    // Is there no way to register for an unknown method ?
+
+    // Send Telemetry every 5.5 secs
+    await sendTelemetryAtInterval(client, sensorComponentName, 5.5);
```
anthonyvercolano, line 154: >await sendTelemetryAtInterval(client, sensorComponentName, 5.5); [](start = 4, length = 64)

I'm thinking this is not good.  Essentially when you are saying await, you are saying wait here unitl setTelementryAtInterval resolves.  But I don't think you wait to wait here.  You want to go on immediately and execute the following code.

digitaltwins/samples/device/javascript/pnpSampleDevice.js
```diff
@@ -0,0 +1,194 @@
...
+    console.log('Sending telemetry message %d...', index);
+    const data = JSON.stringify({ temp: 1 + (Math.random() * 90), humidity: 1 + (Math.random() * 99) });
+    const pnpMsg = new Message(data);
+    pnpMsg.properties.add(messageSubjectProperty, componentName);
+    pnpMsg.contentType = 'application/json';
+    pnpMsg.contentEncoding = 'utf-8';
+    await deviceClient.sendEvent(pnpMsg);
+    index += 1;
+  }, interval);
+};
```
anthonyvercolano, line 136: const sendTelemetryAtInterval = async (deviceClient, componentName, secs) => {
  const interval = parseFloat(secs)*1000;
  let index = 0;
  while (true) {
    console.log('Sending telemetry message %d...', index++);
    const data = JSON.stringify({ temp: 1 + (Math.random() * 90), humidity: 1 + (Math.random() * 99) });
    const pnpMsg = new Message(data);
    pnpMsg.properties.add(messageSubjectProperty, componentName);
    pnpMsg.contentType = 'application/json';
    pnpMsg.contentEncoding = 'utf-8';
    await deviceClient.sendEvent(pnpMsg);
    await sleep(interval);
  }
};

digitaltwins/samples/device/javascript/pnpSampleDevice.js
```diff
@@ -0,0 +1,194 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+
+const Protocol = require('azure-iot-device-mqtt').Mqtt;
+const Client = require('azure-iot-device').Client;
+const Message = require('azure-iot-device').Message;
```
anthonyvercolano, line 8: >; [](start = 51, length = 1)

add:
const { promisify }= require('util');

olivakar: do we need it ? i guess u added it for **promisifying** sleep

digitaltwins/samples/device/javascript/pnpSampleDevice.js
olivakar: i went back and forth with this...i have put it in the command handler now

digitaltwins/samples/device/javascript/pnpSampleDevice.js
olivakar: i have seen some examples of interval and timeout accepting async callbacks but from our class i believe we should not mix callbacks with async. have removed async/await and this will only run in the `check` phase of event loop now.

digitaltwins/samples/device/javascript/pnpSampleDevice.js
olivakar: i think the reason this will not work (unless we change the telemetry to be last thing that we do) is because we will keep on scheduling awaits and these will all be scheduled in the microtask queue which has a higher preference of getting processed immediately...so we will never be able to schedule the other things as we are in a while true loop...plz correct me if i am mistaken

digitaltwins/samples/device/javascript/pnpSampleDevice.js
olivakar: should be..changed the error message

digitaltwins/samples/device/javascript/pnpSampleDevice.js
anthonyvercolano: yup.

---
In reply to: [442561203](https://github.com/Azure/azure-iot-sdk-node/pull/821#discussion_r442561203) [](ancestors = 442561203)

digitaltwins/samples/device/javascript/pnpSampleDevice.js
anthonyvercolano: I thought I looked for examples of giving an async to setInterval or setTimer and didn't find one.  I will check again

---
In reply to: [442562459](https://github.com/Azure/azure-iot-sdk-node/pull/821#discussion_r442562459) [](ancestors = 442562459)

digitaltwins/samples/device/javascript/package.json
```diff
@@ -21,7 +21,8 @@
     "azure-iot-device": "1.12.0-pnp-refresh.0",
     "azure-iot-common": "1.12.0-pnp-refresh.0",
     "azure-iot-device-mqtt": "1.12.0-pnp-refresh.0",
```
anthonyvercolano, line 23: We should be able to turn this to the released packages.

digitaltwins/samples/device/javascript/pnpSampleDevice.js
```diff
@@ -0,0 +1,212 @@
...
+const helperAttachExitListener = async (deviceClient) => {
+  const standardInput = process.stdin;
+  standardInput.setEncoding('utf-8');
+  console.log('Please enter q or Q to exit sample.');
+  standardInput.on('data', (data) => {
+    if (data === 'q\n' || data === 'Q\n') {
+      console.log('Clearing intervals and exiting sample.');
+      clearInterval(intervalToken);
+      deviceClient.close();
+      process.exit();
```
anthonyvercolano, line 131: >process.exit(); [](start = 6, length = 15)

Don't know that you need this.

digitaltwins/samples/device/javascript/pnpSampleDevice.js
```diff
@@ -0,0 +1,212 @@
...
+
+  try {
+    // Add the modelId here
+    await client.setOptions({ modelId: modelId });
+    await client.open();
+    console.log('Enabling the commands on the client');
+    client.onDeviceMethod(commandNameBlink, commandHandler);
+    client.onDeviceMethod(commandNameTurnOn, commandHandler);
+    client.onDeviceMethod(commandNameTurnOff, commandHandler);
+    // Is there no way to register for an unknown method ?
```
anthonyvercolano, line 162: >// Is there no way to register for an unknown method ? [](start = 4, length = 54)

remove this comment.

digitaltwins/samples/device/javascript/pnpSampleDevice.js
```diff
@@ -0,0 +1,212 @@
...
+  }
+  case commandNameTurnOn: {
+    await sendCommandResponse(request, response, 200, 'turn On response');
+    break;
+  }
+  case commandNameTurnOff: {
+    await sendCommandResponse(request, response, 200, 'turn Off response');
+    break;
+  }
+  default: // Is there no way to register for an unknown method?
```
anthonyvercolano, line 60: >// Is there no way to register for an unknown method? [](start = 11, length = 53)

Remove this comment.

digitaltwins/samples/device/javascript/pnpSampleDevice.js
olivakar: if you are referring to the process.exit how do i quit application otherwise ?

PR #824, opened 2020-06-25
refactor(multiple): update to typescript 3.7.5 move to dist instead o…

anthonyvercolano:
…f lib parameter checking and suppression

Start changing files to build with 3.7.5 change the transpiling to output to dist, more checking (and suppressing) parameter usage.

Still quite a few directories to go


common/transport/amqp/.npmignore
```diff
@@ -3,4 +3,5 @@ devdoc/
 test/
 samples/
 src/
-lib/**/*.map
\ No newline at end of file
+dist/**/*.map
+.nyc_output
```
YoDaMa, line 7: is this supposed to be added?

common/transport/amqp/src/amqp.ts
```diff
@@ -213,11 +214,16 @@ export class Amqp {
             this._fsm.transition('connecting', connectionParameters, connectCallback);
           },
           disconnect: (callback) => callback(null, new results.Disconnected()),
+          // @ts-ignore TS6133 endpoint and linkOptions are not used
```
YoDaMa, line 217: would be in favor of not having these and instead using _

common/transport/amqp/src/sender_link.ts
```diff
@@ -118,6 +118,7 @@ export class SenderLink extends EventEmitter implements AmqpLink {
       this._fsm.handle('senderRejectedEvent', context);
     };
 
+    // @ts-ignore TS6133 context not used
```
YoDaMa, line 121: everywhere.

device/samples/simple_sample_device.js
```diff
@@ -61,13 +57,13 @@ function connectCallback () {
...
 
-};
+}
 
 // fromConnectionString must specify a transport constructor, coming from any transport package.
 let client = Client.fromConnectionString(deviceConnectionString, Protocol);
 
 client.on('connect', connectCallback);
-client.on('error', errorCallback);
+client.on('error', (err) => console.error(err.message));
```
YoDaMa, line 66: could you remove this change.

device/transport/http/src/http.ts
```diff
@@ -68,8 +68,8 @@ export class Http extends EventEmitter implements DeviceTransport {
   private _http: Base;
   private _opts: HttpReceiverOptions;
   private _cronObj: any;
-  private _intervalObj: number;
-  private _timeoutObj: number;
+  private _intervalObj: NodeJS.Timer;
```
YoDaMa, line 71: are these necessary for the Typescript upgrade? If not, I'd actually prefer them in their own PR.

anthonyvercolano: Yup, kinda are.  It started beefing about this.

common/transport/amqp/src/sender_link.ts
anthonyvercolano: Sure.

common/transport/amqp/src/amqp.ts
anthonyvercolano: Great idea.  Thank you.

common/transport/amqp/.npmignore
anthonyvercolano: It certainly seemed like the right think to do.  Consistency between all the directories.

device/samples/simple_sample_device.js
anthonyvercolano: Sure.

PR #829, opened 2020-06-30
chore: rename digital-twin-model-id to modelId in mqtt (Merge pending Deployment)

jebrando:
# Checklist

# Description of the problem

# Description of the solution


device/transport/mqtt/src/mqtt.ts
```diff
@@ -537,7 +537,7 @@ export class Mqtt extends EventEmitter implements DeviceTransport {
 
     /*Codes_SRS_NODE_DEVICE_MQTT_41_XXX: [The MQTT transport should ] */
```
anthonyvercolano, line 538: >[The MQTT transport should ] [](start = 41, length = 28)

Realize that you didn't make the mistake, but, could you either copy the full requirement text here or delete the requirement comment entirely?
Thank you.

PR #843, opened 2020-07-08
chore(azure-iot-device) : update the handler for twin

olivakar:


device/samples/pnp/pnpTemperatureController.js
```diff
@@ -97,30 +98,41 @@ const updateComponentReportedProperties = (deviceTwin, patch, componentName) =>
   });
 };
 
-const helperAttachHandlerForDesiredPropertyPatches = (deviceTwin) => {
+const helperAttachHandlerForDesiredPropertyPatches = (deviceTwin, componentNames) => {
   deviceTwin.on('properties.desired', (delta) => {
-    if (delta[thermostat1ComponentName] || delta[thermostat2ComponentName] || delta[deviceInfoComponentName]) {
+    console.log('Received an update for device with value: ' + JSON.stringify(delta));
+    Object.entries(delta).forEach(([key, values]) => {
```
anthonyvercolano, line 104: > Object.entries(delta).forEach(([key, values]) => { [](start = 3, length = 51)

Consider the case of the application being out of date with the model.   So that this device only knows about components a,b, and c.  However, suppose that there is a new component d.

Consider further that a changes occur inside a, b, and c, and a writeable property of d.  so d.foo is changed.

The code will to reports for a, b, c in the block starting at line 106.  When it gets to d, the code will fall through to the block starting at 122.  It appears that it will report a,b,c, and d all over again.

device/samples/pnp/pnpTemperatureController.js
```diff
@@ -97,30 +98,41 @@ const updateComponentReportedProperties = (deviceTwin, patch, componentName) =>
...
-          propertyContent.ad = 'Successfully executed patch';
-          propertyContent.av = version;
-          patch[componentName][propertyName] = propertyContent;
-        }
-      });
-      updateComponentReportedProperties(deviceTwin, patch, componentName);
-    }
+      else if  (key !== '$version') {
+        const patchForRoot = { };
+        Object.entries(delta).forEach(([propertyName, propertyValue]) => {
```
anthonyvercolano, line 123: >Object.entries(delta).forEach(([propertyName, propertyValue]) => { [](start = 8, length = 66)

In addition in reference to my above comment, I believe the reports for a,b, and c could be in error.

the desired looks like below (not showing d):
{
   a: {
         __t: 'c',
         bob: 1,
          carol: 2
       },
    b: {
        __t: 'c',
        ted: 3,
        alice: 4
     }

Could you end up reporting:
    a: {
          value: {__t: 'c', bob: 1, carol: 2},
          ac: 200,
          av: 2
        },
     b: {
          value: {__t: 'c', ted:  3, alice: 4},
          ac: 200,
           av: 2
        }

device/samples/pnp/pnpTemperatureController.js
olivakar: my assumption that user must know about the existence of all its components

device/samples/pnp/pnpTemperatureController.js
```diff
@@ -9,7 +9,8 @@ const Message = require('azure-iot-device').Message;
 
 // String containing Hostname, Device Id & Device Key in the following formats:
 //  'HostName=<iothub_host_name>;DeviceId=<device_id>;SharedAccessKey=<device_key>'
-const deviceConnectionString = process.env.DEVICE_CONNECTION_STRING;
+// const deviceConnectionString = process.env.DEVICE_CONNECTION_STRING;
+const deviceConnectionString = 'HostName=hubforsdkfolks.azure-devices.net;DeviceId=olkarnode;SharedAccessKey=3FC/uiphxm2+w00mHdR10ZBdSOKfyyQFF5qHNYx63HY=';
```
jebrando, line 13: Please remove conn string

PR #850, opened 2020-07-14
chore: updating package files to include the correct version

jebrando:
# Checklist

# Description of the problem

# Description of the solution


digitaltwins/samples/service/javascript/get_digital_twin.js
```diff
@@ -15,10 +15,6 @@ const { inspect } = require('util');
 async function main() {
   const deviceId = process.env.IOTHUB_DEVICE_ID;
 
-  // Create digital twin service client
-  const credentials = new IoTHubTokenCredentials(process.env.IOTHUB_CONNECTION_STRING);
```
anthonyvercolano, line 19: I think you need the hub connection string.

digitaltwins/samples/service/javascript/get_digital_twin.js
```diff
@@ -15,10 +15,6 @@ const { inspect } = require('util');
...
-  const digitalTwinServiceClient = new DigitalTwinServiceClient(credentials);
```
anthonyvercolano, line 20: I think you need to authenticate with something.  And that is the hub connection string.

PR #856, opened 2020-07-21
fix(azure-iot-device): invalid body received on a device method in…

anthonyvercolano:
…voke throws

If somehow an invalid body is pushed through on a device method invocation, the device client side
will throw when it trys to json parse that body.

### DIFF
--- a/device/core/src/internal_client.ts
+++ b/device/core/src/internal_client.ts
@@ -396,18 +396,25 @@ export abstract class InternalClient extends EventEmitter {
   private _addMethodCallback(methodName: string, callback: (request: DeviceMethodRequest, response: DeviceMethodResponse) => void): void {
     const self = this;
     this._transport.onDeviceMethod(methodName, (message) => {
-      // build the request object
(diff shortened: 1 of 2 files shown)


device/core/src/internal_client.ts
```diff
@@ -396,18 +396,25 @@ export abstract class InternalClient extends EventEmitter {
...
+          message.requestId,
+          message.methods.methodName,
+          message.body
+        );
+        // Codes_SRS_NODE_INTERNAL_CLIENT_13_001: [ The onDeviceMethod method shall cause the callback function to be invoked when a cloud-to-device method invocation signal is received from the IoT Hub service. ]
+        callback(request, response);
+      } catch (err) {
+        response.send(400, 'Invalid request format: ' + err.message, (err) => {
+          if (err) {
+            debug('Error sending invalid request response back to application');
```
jebrando, line 414: Nitpick: Can you change to something like 'Error sending invalid request response'

PR #864, opened 2020-07-24
chore: adding modelId to the twin object

jebrando:
# Checklist

# Description of the problem

# Description of the solution


anthonyvercolano:
Well, sorry about this, but it would be best if you updated the package.json files to use  -pnp-refresh.4

Also does this include the fix for common from 1.12.0 -> 1.12.4  ?

Also, why are there these humongous commit histories?

PR #865, opened 2020-07-27
docs(azure-iot-device): setOptions documentation

YoDaMa:
Part of this is adding clarity where the Device Client supports Proxy. It is also consolidating our documentation for setOptions into a file, a new "README" for the Device Client that is linked from the main readme. Maybe the better place should be the Wiki... Unclear. That can be discussed @elhorton.


doc/device_client.md
```diff
@@ -0,0 +1,32 @@
+# Azure IoT Device Client Library
+
+The Azure IoT Device Client Library provides an efficient and well tested API for connecting and using devices with the Azure IoT Hub.
+
+## client.setOptions(options, [callback])
```
elhorton, line 5: (fyi, I'm going to be asking naive questions like a customer with no knowledge of the SDK)-- in what circumstance would I add a callback to the setOptions call?

doc/device_client.md
```diff
@@ -0,0 +1,32 @@
...
+- `keepalive`: Keepalive interval in numeric format (seconds). This controls the keepalive ping for MQTT specifically. If you are using AMQP or HTTP, this will do nothing.
+- `productInfo`: Custom user defined information to be appended to existing User Agent information. The User Agent Identification information is used predominantly by Microsoft internally for identifying metadata related to Device Client usage for Azure IoT.
+- `modelId`: !!Digital Twin Use Only!! String used in MQTT username setting the Digital Twin modelId.
+- `tokenRenewal`: Optional object with token renewal values.  Only use with authentication that uses pre shared keys.
+- `mqtt`: Options specific to the MQTT transport
+    - `webSocketAgent`: [Agent]{@link https://nodejs.org/api/https.html#https_class_https_agent} object to use with MQTT-WS connections. Can be used for tunneling proxies.
```
elhorton, line 17: in the case of protocol specific options, do I need to specify the protocol? Or will a call to client.SetOptions(webSocketAgent=xxx) automatically work for whichever protocol I'm using (assuming amqp or mqtt in this case based on the parameter name)

doc/device_client.md
```diff
@@ -0,0 +1,32 @@
+# Azure IoT Device Client Library
+
+The Azure IoT Device Client Library provides an efficient and well tested API for connecting and using devices with the Azure IoT Hub.
+
+## client.setOptions(options, [callback])
+
+Used directly after creating the Device Client to set options on the client. These should not be dynamically altered during runtime of the device client.
+
+### Available Options
```
elhorton, line 9: Will these options also show up somewhere in API reference docs?

doc/device_client.md
```diff
@@ -0,0 +1,32 @@
...
+    - `agent`: [Agent]{@link https://nodejs.org/api/https.html#https_class_https_agent} object to use with HTTP connections. Can be used for tunneling proxies.
+    - `receivePolicy`: Configuration parameters to use for receive polling.
+        - `interval`: Interval **in seconds** at which the Azure IoT hub is going to be polled.
+        - `at`: Use this option to configure the receiver to receive only once at a specific time.
+        - `cron`: Use a cron-formatted string.
+        - `manualPolling`: Does not poll and instead rely on the user calling the `receive` method.
+        - `drain`: Boolean indicating whether only one message should be received all messages should be drained.
+- `amqp`: Options specific to the AMQP transport
+    - `webSocketAgent`: [Agent]{@link https://nodejs.org/api/https.html#https_class_https_agent} object to use with AMQP-WS connections. Can be used for tunneling proxies.
+- `cert`: X509 Certificate.
```
jebrando, line 28: Could you please expand this cert one to be

x.509 certificate used for authenticating the device to the IoTHub.

doc/device_client.md
```diff
@@ -0,0 +1,32 @@
...
+- `key`: Key associated with the X509 certificate.
+- `passphrase`: Passphrase used to decode the key associated with the X509 certificate.
+- `clientCertEngine`: Name of an OpenSSL engine which can provide the client certificate.
```
jebrando, line 31: We can use an OpenSSL Engine?  Wow, I had no idea, I wonder if this works.

doc/device_client.md
```diff
@@ -0,0 +1,32 @@
...
+        - `interval`: Interval **in seconds** at which the Azure IoT hub is going to be polled.
+        - `at`: Use this option to configure the receiver to receive only once at a specific time.
+        - `cron`: Use a cron-formatted string.
+        - `manualPolling`: Does not poll and instead rely on the user calling the `receive` method.
+        - `drain`: Boolean indicating whether only one message should be received all messages should be drained.
+- `amqp`: Options specific to the AMQP transport
+    - `webSocketAgent`: [Agent]{@link https://nodejs.org/api/https.html#https_class_https_agent} object to use with AMQP-WS connections. Can be used for tunneling proxies.
+- `cert`: X509 Certificate.
+- `key`: Key associated with the X509 certificate.
+- `passphrase`: Passphrase used to decode the key associated with the X509 certificate.
```
jebrando, line 30: Can you make the X509 to be X.509.  It looks fancier.

doc/device_client.md
```diff
@@ -0,0 +1,32 @@
...
+
+The Azure IoT Device Client Library provides an efficient and well tested API for connecting and using devices with the Azure IoT Hub.
+
+## client.setOptions(options, [callback])
+
+Used directly after creating the Device Client to set options on the client. These should not be dynamically altered during runtime of the device client.
+
+### Available Options
+
+- `ca`: Public certificate in PEM form for certificate authority being used by the Hub service.  This is the CA that the hub is using to secure TLS connections and the client validates the connection using this public cert in order to validate the identity of the hub.  If you are connecting to an Azure IoT Hub inside of an Azure data center, you do not need to set this.  If you are connecting to some other hub (e.g. an Edge Hub), then you may need to set this to the server cert that the hub uses for TLS.
```
jebrando, line 11: Can you please change "an Edge Hub" to "an EdgeHub" to satisfy poli-check.

doc/device_client.md
YoDaMa: This is incorrect. what's a good way to guide users that they need to put it in an object?

i.e.: `client.setOptions({mqtt: {websocketAgent: agent}}`

(also, I know the API looks kinda dumb, since all three can accept an agent, but,,, that's just how it is and we can't break it now).

doc/device_client.md
YoDaMa: callback is if you're using client.setOptions in the traditional node callback way, where once it completes the callback is used to move on to the next operation. In the Node Promises syntax you would not need a callback, and just call await on the setOptions: `await client.setOptions(options);`

doc/device_client.md
YoDaMa: ...no...

doc/device_client.md
YoDaMa: no clue tbh

doc/device_client.md
elhorton: ok, should they? seems to me like it would be helpful but if that's breaking API reference doc precedent that's fine

doc/device_client.md
elhorton: ok got it, as long as customers understand the use here--- as a javascript developer seems pretty straightforward

doc/device_client.md
elhorton: yeah it's fine, doesn't have to be perfect as long as it's usable-- in this case, probably best to just show an example. As in, before starting the protocol specific examples with `mqtt`, a line saying:

 **Protocol specific options**: to use options for a specific protocol, you must pass it as an object that specifies both the protocol used and the option you are setting. For example, `client.setOptions({mqtt: {websocketAgent: agent}})`.

doc/device_client.md
YoDaMa: wait sorry, these options do show up in the API somewhere, but not in this form. I pulled most of the definitions for the options from the docstrings.

Issue #903, opened 2020-10-05
Using `query.next()` with no parameters does not work as specified


khendry:
# Context

- **OS and version used:** macOS 10.15.7
- **Node.js version:** 10.20.1
- **npm version:** 6.14.4

## Description of the issue

I'm implementing an HTTP triggered Azure function to query my DPS for the list of enrolment group registration states using the `azure-iot-provisioning-server` module. I create the client, construct the query spec and when I trigger the query using `query.next()` the function throws and error. 

When I modify the execution of the query, to use the query callback and wrap that in a promise, I can work around this issue. But the API does not seem to work as written. 

## Code sample exhibiting the issue

``` TypeScript
    // Inside and `async` function
    const dpsClient = ProvisioningServiceClient.fromConnectionString(dpsConnStr)
    const query = dpsClient.createEnrollmentGroupDeviceRegistrationStateQuery({
        query: '*'
    }, enrollmentGroupId, 500)
    const { responseBody } = await query.next()
    // BOOM!
```

## Console log of the issue

``` Console
LanguageWorkerConsoleLog[error]
Worker e6a958c5-7cb7-4abb-8ec4-77a5ea194d12 uncaught exception (learn more: https://go.microsoft.com/fwlink/?linkid=2097909 ): TypeError: actualCallback is not a function
     at ~/functions/node_modules/azure-iot-provisioning-service/dist/query.js:37:21
	 at requestCallback (~/functions/node_modules/azure-iot-http-base/dist/rest_api_client.js:142:17)     
	 at IncomingMessage.<anonymous> (~/functions/node_modules/azure-iot-http-base/dist/http.js:118:17)     
	 at IncomingMessage.emit (events.js:203:15)     
	 at endReadableNT (_stream_readable.js:1145:12)     
	 at process._tickCallback (internal/process/next_tick.js:63:19)
Executed 'Functions.things' (Failed, Id=80a1aa0c-7cf6-4464-bcb4-359cbded357e, Duration=760ms)
System.Private.CoreLib: Exception while executing function: Functions.things. System.Private.CoreLib: node exited with code 1
 LanguageWorkerConsoleLog[error] Worker e6a958c5-7cb7-4abb-8ec4-77a5ea194d12 uncaught exception (learn more:
 https://go.microsoft.com/fwlink/?linkid=2097909 ):
 TypeError: actualCallback is not a function     
[...]

PR #912, opened 2020-11-02
fix (azure-iot-provisioning-server): #903 - Using  with no parameters does not work as specified

BertKleewein:
Also fixed provisioning service samples directory to be lerna'ed and lintable.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -15,6 +15,7 @@
     "provisioning/transport/*",
     "provisioning/device/samples",
     "provisioning/device/ts-samples",
+    "provisioning/service/samples",
     "security/*",
     "sdklab/longhaultests",
     "sdklab/mean_time_recovery",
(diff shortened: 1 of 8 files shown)


lerna.json
```diff
@@ -15,6 +15,7 @@
     "provisioning/transport/*",
     "provisioning/device/samples",
     "provisioning/device/ts-samples",
+    "provisioning/service/samples",
```
anthonyvercolano, line 18: Nice catch thank you!  We already had this in the version  bump.

anthonyvercolano:
More of a nit: any chance we could add a unit test?

anthonyvercolano (approved):
:shipit:

anthonyvercolano:
Thank you for catching and fixing the samples setup!

PR #914, opened 2020-11-20
switch to v2 of tempcontroller model

ericmitt:
Make tempcontroller sample compatible with Central. use now v2 of model


anthonyvercolano (approved):
:shipit:

PR #915, opened 2020-11-30
docs(azure-iothub): update the documentation for udpate and updateDev…

anthonyvercolano:
…ice designating maditory fields


YoDaMa (approved):

service/src/registry.ts
```diff
@@ -115,6 +115,13 @@ export class Registry {
    * @method            module:azure-iothub.Registry#update
    * @description       Updates an existing device identity on an IoT hub with
    *                    the given device information.
+   *                    The `deviceInfo` parameter must include all properties
+   *                    that are updatable.  For example, if updating the `status`
+   *                    property from `disabled` to `enabled`, the `deviceInfo`
+   *                    object should also have the statusReason, authentication
+   *                    (and its sub-properties), capabilities, and deviceScope
+   *                    set to the devices current values or those properties will
+   *                    be reset.
```
elhorton, line 124: Do you think it's self explanatory that the recommended developer workflow be _Get the device identity, then copy all of the fields into the update API_? Or should you explain that workflow?

service/src/registry.ts
```diff
@@ -307,7 +314,13 @@ export class Registry {
...
    * @method            module:azure-iothub.Registry#updateDevices
    * @description       Updates an array of devices.
-   *
+   *                    The individual elements of the `devices` parameter must
+   *                    include all properties that are updatable.  For example,
+   *                    if updating the `status` property from `disabled` to
+   *                    `enabled`, the object should also have the statusReason,
+   *                    authentication (and its sub-properties), capabilities,
+   *                    and deviceScope set to the devices current values or those
+   *                    properties will be reset.
```
elhorton, line 323: Thoughts on just including a link to the IoT Hub documentation as well so they know what properties are "updatable" ? [Device Identity properties](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-devguide-identity-registry#device-identity-properties)

elhorton: also, that documentation page doesn't necessarily describe what is "updatable"... do you know this? I'd like to add a column to that page to make this clear

elhorton (approved):

service/src/registry.ts
```diff
@@ -115,6 +115,13 @@ export class Registry {
...
    * @description       Updates an existing device identity on an IoT hub with
    *                    the given device information.
+   *                    The `deviceInfo` parameter must include all properties
+   *                    that are updatable.  For example, if updating the `status`
+   *                    property from `disabled` to `enabled`, the `deviceInfo`
+   *                    object should also have the statusReason, authentication
+   *                    (and its sub-properties), capabilities, and deviceScope
+   *                    set to the devices current values or those properties will
+   *                    be reset.
    * @param {Object}    deviceInfo  An object which must include a `deviceId`
```
anthonyvercolano, line 125: >param {Object} [](start = 6, length = 14)

Just a quick test.

anthonyvercolano:
>    * @param {Function}  [done]      The optional function to call when the operation is

A second test.

---
Refers to: service/src/registry.ts:128 in 597ed31. [](commit_id = 597ed312ab62ff484ac17ffd2995eb947e998666, deletion_comment = False)

service/src/registry.ts
anthonyvercolano: It would be good to add a column to the doc page.  I am hesitant to add a link.  It could easily become invalid.

service/src/registry.ts
anthonyvercolano: done.

Issue #925, opened 2021-02-03
getBlobSharedAccessSignature calls on device_client.js failing with connection timeout error behind a corporate HTTP proxy.


ksin63:
As an alternate to uploadToBlob method to save SAS token quota from IotHub. I am directly using **_getBlobSharedAccessSignature_** and **_notifyBlobUploadStatus_** method to control when to acquire token or reuse existing token in case of file upload failure/retry scenario. My workflow is described below

1. Use IoT SDK to acquire SAS token:
2. Use Azure Storage SDK to upload file:
3. Again use IoT SDK to notify IoT Hub about success or failure:

This works perfectly in normal environment with edge having direct internet connectivity, though it fails when the edge-proxy is running behind the corporate firewall with below error.
**_Encountered unrecoverable error: Error: connect ETIMEDOUT 13.79.172.43:443_** 

After a quick glance on the source code it seems the developer in this [PR-Add proxy support to blob file upload in Device SDK](https://github.com/Azure/azure-iot-sdk-node/pull/589) missed to apply setOptions for **_FileUploadApi_**. 

A quick fix should go as below
```
setOptions(options: DeviceClientOptions, done?: Callback<results.TransportConfigured>): Promise<results.TransportConfigured> | void {
    if (!options) throw new ReferenceError('options cannot be falsy.');
    if (this._blobUploadClient) {
      /*Codes_SRS_NODE_DEVICE_CLIENT_99_103: [The `setOptions` method shall set `blobUploadClient` options.]*/
      this._blobUploadClient.setOptions(options);
    }
    /* Implicit call of setOption for fileUploadApi here*/
    if (this._fileUploadApi) {
       /* [The `setOptions` method shall set `fileUploadApi` options.]*/
       this._fileUploadApi.setOptions(options);
    }
    return super.setOptions(options, done);
  }
```

abossard:
Would it be possible to backport the fix also to version `1.17.0-lts0720up1`?

YoDaMa:
@abossard we unfortunately only backport security issue patches to LTS releases. Bug fixes like this are the purpose of the main branch.

anthonyvercolano:
@abossard Is this blocking production?

PR #926, opened 2021-02-03
fix: #925 added setOption in case of fileUploadClient

abossard:
# Checklist

#925 

# Description of the problem
Pass the options also to the file upload client

# Description of the solution
Checked if there is a fileUploadClient in use, and if so, the options are added to it.


YoDaMa:
thanks for the contribution @abossard! can you please add a test for this condition? 

It would look very similar to the following test in `_blob_upload_client_test.js`, but it would go inside `_device_client_test.js`. It looks like the `describe` block for the setoptions in device client test is at line 463, so you could start there.

```js
  describe('#setOptions', function () {
    /*Tests_SRS_NODE_DEVICE_BLOB_UPLOAD_CLIENT_99_011: [`setOptions` shall set `fileUploadApi` options.]*/
    it('sets file upload API options', function () {
      var fakeOptions = '__FAKE_OPTIONS__';
      var fakeFileUpload = new FakeFileUploadApi();
      var client = new BlobUploadClient(fakeConfig, fakeFileUpload, null);
      client.setOptions(fakeOptions);
      assert.isTrue(fakeFileUpload.setOptions.calledWith(fakeOptions));
    });
  });
```

abossard:
Hi @YoDaMa  thanks for pointing it out. I added the required test.

anthonyvercolano (approved):
:shipit:

Issue #927, opened 2021-02-04
RegistrationClient interface does not expose promisified version of API


vishnureddy17:
https://github.com/Azure/azure-iot-sdk-node/blob/9bde57d79659a17affeb69002cb2ff270f945d09/provisioning/device/src/interfaces.ts#L192-L208

This `RegistrationClient` interface in `azure-iot-provisioning-device` does not exposed the promisified version of the API, despite `X509Registration` , `SymmetricKeyRegistration`, and `TpmRegistration` all having promisified implementations of `.register()` and `.cancel()`.

PR #928, opened 2021-02-04
fix(azure-iot-provisioning-device): change RegistrationClient to expose promisified version of API

vishnureddy17:
fixes issue #927


YoDaMa (approved):

YoDaMa (changes requested):
I think that this is a good change, can we verify that people will not be broken by this change?

vishnureddy17:
@YoDaMa The callback versions of the functions appear to be working as normal with the change.

YoDaMa (approved):

YoDaMa:
lgtm

PR #947, opened 2021-03-21
feat(node-red-contrib-azureiothubnode): add properties

dzeyelid:
# Checklist

# Description of the problem

I would like to use this Node-RED's node to send messages to IoT Hub. Currently the node sends messages with remain `contentEncoding` and `contentType` `undfined`. By that, the body of recorded data through message routing is encoded by base64. In some cases, it is bad because it's difficult to analyze and query.

# Description of the solution

I added the properties `Content Encoding` and `Content Type` on the node to configure them.

Then, this directory `device/node-red` doesn't have any test. So I didn't write tests.


YoDaMa (approved):
lgtm

PR #949, opened 2021-03-22
feat(azure-iot-common): log when a NoRetry policy is used

vishnureddy17:
This is an ask from CSS to quickly identify when a customer is using a `NoRetry` policy. 

.NET PR on the same issue: https://github.com/Azure/azure-iot-sdk-csharp/pull/1763


vishnureddy17:
Closing because GitHub won't update the PR after pushing a fixup. Will open a new PR.

PR #950, opened 2021-03-22
feat(azure-iot-common): log when a NoRetry policy is used

vishnureddy17:
This is an ask from CSS to quickly identify when a customer is using a `NoRetry` policy. 

.NET PR on the same issue: https://github.com/Azure/azure-iot-sdk-csharp/pull/1763

### DIFF
--- a/common/core/src/retry_operation.ts
+++ b/common/core/src/retry_operation.ts
@@ -24,6 +24,9 @@ export class RetryOperation {
    * @param {number} maxTimeout  The maximum timeout for this operation, after which no retry will be attempted.
    */
   constructor (policy: RetryPolicy, maxTimeout: number) {
+    if (policy && policy.constructor && policy.constructor.name === 'NoRetry') {
+      debug('A RetryOperation is being used with a NoRetry policy. The operation will not be retried on failure.');
+    }
     this._policy = policy;
     this._maxTimeout = maxTimeout;
   }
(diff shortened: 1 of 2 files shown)


common/core/src/retry_policy.ts
```diff
@@ -156,6 +159,10 @@ export class ExponentialBackOffWithJitter implements RetryPolicy {
  * @implements {RetryPolicy}
  */
 export class NoRetry implements RetryPolicy {
+  constructor() {
```
anthonyvercolano, line 162: Should this be in the constructor for RetryOperation?

anthonyvercolano: Someting like policy.constructor.name === 'NoRetry'

vishnureddy17: Wouldn't it cause a problem if the user creates a custom retry policy called `NoRetry`? Maybe this case is insignificant enough to do it the way you suggest.

anthonyvercolano: Well, what we are doing here can NOT affect the running of the application SDK code.  In the end all we are producing is a debug.

Second, if they hand us a retry policy whose constructor is named NoRetry, I'm going to assume that it doesnt retry.

The nice thing here is as opposed to simply instantiating something that is NoRetryish, but moving this to the RetryOperation, we can say the developer is actually doing utilizing something that is NoRetryish.

anthonyvercolano (approved):

PR #955, opened 2021-04-06
feat(azure-iothub): add AAD authentication

zolvarga:
# Checklist

# Description of the problem

# Description of the solution


common/transport/http/src/rest_api_client.ts
```diff
@@ -109,7 +113,60 @@
     let httpHeaders: any = headers || {};
     if (this._config.sharedAccessSignature) {
       httpHeaders.Authorization = (typeof(this._config.sharedAccessSignature) === 'string') ? this._config.sharedAccessSignature as string : (this._config.sharedAccessSignature as SharedAccessSignature).extend(anHourFromNow());
+      this.executeBody(requestBody, httpHeaders, headers, method, path, timeout, requestOptions, done);
+    } else if (this._config.tokenCredential) {
+      let accessToken = this.getToken();
+      Promise.resolve(accessToken).then((value) => {
```
anthonyvercolano, line 119: What happens if this rejects?

anthonyvercolano: I'm thinking if this rejects, the program will crash.  If not now, then under node 14?

common/transport/http/package.json
```diff
@@ -19,8 +19,9 @@
     "sinon": "^7.4.1",
     "source-map-support": "^0.5.16",
     "ts-node": "^8.6.2",
-    "tslint": "^6.1.3",
-    "typescript": "3.7.5"
+    "tslint": "^6.1.0",
```
anthonyvercolano, line 22: Why did we go backwards in the version of tslint?

service/src/client.ts
```diff
@@ -453,7 +455,8 @@ export class Client extends EventEmitter {
     const config: Client.TransportConfigOptions = {
       host: decodedUri,
       keyName: sas.skn,
-      sharedAccessSignature: sas.toString()
+      sharedAccessSignature: sas.toString(),
```
anthonyvercolano, line 458: Why isn't there a fromToken in here and jobClient?

anthonyvercolano (changes requested):
Let's wait on the job and other entries before we check this in,

service/src/client.ts
zolvarga: Added, fixed

common/transport/http/src/rest_api_client.ts
zolvarga: Error handling added

common/transport/http/package.json
zolvarga: There is no reason :) Rolled back

anthonyvercolano (approved):
Finish up testing?  Make sure we squash before we merge

PR #984, opened 2021-06-09
[DON'T REVIEW] fix: 🐛 Support US Government cloud for AAD authentication

vishnureddy17:
- Allow users to specify the token scope for AAD authentication
- Add constant values for the token scopes for public clouds and US Government clouds
- Fix error handling issues with `getToken()` in `RestApiClient` and `Amqp`
- Add missing unit tests for AAD authentication


vishnureddy17:
Closing because a different token scope is no longer required for US Government cloud. Keeping the branch around in case custom token scopes need to be added in the future.

PR #991, opened 2021-06-16
Fix getToken() error handling and add missing tests for AAD authentication

vishnureddy17:
- Fix error handling issues with `getToken()` in `RestApiClient` and `Amqp`
- Add missing unit tests for AAD authentication

### DIFF
--- a/common/transport/http/src/rest_api_client.ts
+++ b/common/transport/http/src/rest_api_client.ts
@@ -173,6 +173,9 @@ export class RestApiClient {
     if ((!this._accessToken) || this.isAccessTokenCloseToExpiry(this._accessToken)) {
       this._accessToken = await this._config.tokenCredential.getToken(this._iotHubPublicScope) as any;
     }
+    if (!this._accessToken) {
+      throw new Error('AccessToken creation failed');
+    }
     return this._BearerTokenPrefix + this._accessToken.token;
   }
 
--- a/common/transport/http/src/rest_api_client.ts
+++ b/common/transport/http/src/rest_api_client.ts
@@ -36,7 +36,7 @@ export interface HttpTransportError extends Error {
  * @throws {ArgumentError}   If the config argument is missing a host or sharedAccessSignature error
  */
 export class RestApiClient {
-  private _iotHubPublicScope: string[] = ['https://iothubs.azure.net/.default'];
+  private _iotHubPublicScope: string = 'https://iothubs.azure.net/.default';
   private _BearerTokenPrefix: string = 'Bearer ';
   private _MinutesBeforeProactiveRenewal: number = 9;
   private _MillisecsBeforeProactiveRenewal: number = this._MinutesBeforeProactiveRenewal * 60000;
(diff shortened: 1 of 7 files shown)


common/transport/http/src/rest_api_client.ts
```diff
@@ -36,7 +36,7 @@ export interface HttpTransportError extends Error {
  * @throws {ArgumentError}   If the config argument is missing a host or sharedAccessSignature error
  */
 export class RestApiClient {
-  private _iotHubPublicScope: string[] = ['https://iothubs.azure.net/.default'];
+  private _iotHubPublicScope: string = 'https://iothubs.azure.net/.default';
```
YoDaMa, line 39: does this affect the getToken credential call signature, since you're removing the array wrap?

common/transport/http/src/rest_api_client.ts
```diff
@@ -176,11 +173,7 @@ export class RestApiClient {
     if ((!this._accessToken) || this.isAccessTokenCloseToExpiry(this._accessToken)) {
       this._accessToken = await this._config.tokenCredential.getToken(this._iotHubPublicScope) as any;
```
YoDaMa, line 174: this._iotHubPublicScope is now a sring. does getToken() change?

service/src/amqp.ts
```diff
@@ -592,11 +589,7 @@ export class Amqp extends EventEmitter implements Client.Transport {
     if ((!this._accessToken) || this.isAccessTokenCloseToExpiry(this._accessToken)) {
       this._accessToken = await this._config.tokenCredential.getToken(this._iotHubPublicScope) as any;
     }
-    if (this._accessToken) {
-      return this._bearerTokenPrefix + this._accessToken.token;
-    } else {
-      return null;
-    }
+    return this._bearerTokenPrefix + this._accessToken.token;
```
YoDaMa, line 592: is this._accessToken guaranteed to exist?

YoDaMa (commented):
see comments

common/transport/http/src/rest_api_client.ts
vishnureddy17: It accepts a string or an array of strings. In the AMQP client, we were using a string, so I changed it here to be consistent.

service/src/amqp.ts
vishnureddy17: No, but the promise would get rejected if it doesn't exist

common/transport/http/src/rest_api_client.ts
vishnureddy17: See above comment

service/src/amqp.ts
vishnureddy17: Follow up based on a discussion with Daniel Rodriguez who works on `@azure/identity`:
It is safe to assume that for implementations of `TokenCredential` from `@azure/identity`, `getToken()` will always throw rather than returning null. But for custom implementations of `TokenCredential`, this may not be the case.

Based on this, I will change it back to what it was before.

vishnureddy17: Addressed this

YoDaMa (approved):
lgtm

PR #1000, opened 2021-07-19
Add notice about upcoming service TLS changes to the readme

timtay-microsoft:
I'm adding these notices to all or our repos so that users have plenty of time to add the new public certificate to their device's certificate store before this change happens

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -7,18 +7,18 @@
 
 ## Critical Upcoming Change Notice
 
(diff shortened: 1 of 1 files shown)


timtay-microsoft:
I'm not that familiar with the Node SDK, but I'm assuming here that it pulls its trusted certificates from the device's certificate store. Please correct me if I'm wrong here, though.

anthonyvercolano (approved):

PR #1001, opened 2021-07-20
feat: 🎸 allow modelId to be specified in device and module client fac…

vishnureddy17:


device/transport/mqtt/src/mqtt.ts
```diff
@@ -520,6 +520,17 @@ export class Mqtt extends EventEmitter implements DeviceTransport {
...
+  /**
+   * @private
+   * @method         module:azure-iot-device-mqtt:Mqtt#setModelId
+   * @description    This method sets the model ID so that it is included in the username on connect as per the Azure IoT Plug and Play requirements.
+   *
+   * @param {String}    modelId The model Id to set.
+   */
+  setModelId(modelId: string): void {
+    this._mid = '&model-id=' + modelId;
+  }
```
BertKleewein, line 532: There's already code in setOptions (line 550 below) to set model id.  Why add a second way do do this?

device/core/src/module_client.ts
```diff
@@ -361,13 +364,25 @@ export class ModuleClient extends InternalClient {
...
-  static fromEnvironment(transportCtor: any, callback: Callback<ModuleClient>): void;
   static fromEnvironment(transportCtor: any): Promise<ModuleClient>;
-  static fromEnvironment(transportCtor: any, callback?: Callback<ModuleClient>): Promise<ModuleClient> | void {
+  static fromEnvironment(transportCtor: any, modelId: string): Promise<ModuleClient>;
+  static fromEnvironment(transportCtor: any, callback: Callback<ModuleClient>): void;
+  static fromEnvironment(transportCtor: any, modelId: string, callback: Callback<ModuleClient>): void;
+  static fromEnvironment(transportCtor: any, callbackOrModelId?: Callback<ModuleClient> | string, callback?: Callback<ModuleClient>): Promise<ModuleClient> | void {
+    if (callbackOrModelId && typeof(callbackOrModelId) !== 'function' && typeof(callbackOrModelId) !== 'string') {
+      throw new TypeError('The second argument to fromEnvironment must be a callback or a modelId string.');
+    }
```
BertKleewein, line 379: I forgot how much I HATE HATE HATE function overloading in node.  Out of all the code in this review, this is the most fragile looking.  (No insult intended, this is just the way we do it, and it sucks.)  I recommend adding unit tests for this to test all of the various new valid and invalid cases this code adds.

BertKleewein (changes requested):
:clock1:

device/transport/mqtt/src/mqtt.ts
vishnureddy17: A couple of reasons:
1. `setOptions` will eventually call `setOptions` on our base MQTT client which will overwrite any options passed in previously. My concern was that the user passes in an MQTT transport with some options that they already set and then we come in and internally call `setOptions` again and overwrite their options. 
2. `setOptions` takes a callback despite setting the modelId being synchronous. In the case of setting the model id, the callback is actually useless. We really shouldn't have a function that is sometimes synchronous and sometimes asynchronous. 

On second thought, I'm actually leaning toward removing this and using setOptions because:
1. Users shouldn't be using our transport objects directly to begin with. And this wouldn't be a breaking change because we would only call `setOptions` when the modelId is specified.
2. While it is bad practice to have a function that is sometimes synchronous and sometimes asynchronous, we already do it all over the place in our SDK. And perhaps it's a lesser evil that having two ways of doing the same thing. Whenever we get around to Track 2, we should be careful to avoid situations like this in the future.

device/core/src/module_client.ts
vishnureddy17: Function overloading is made much worse in our situation because we have to retain callback functionality. IMO, it wouldn't be that bad if we got rid of the callbacks. Hopefully we don't have to deal with this in Track 2.

Yeah, there needs to be tests for `fromEnvironment`. Since testing `fromEnviroment` seems much more involved than the other factory methods, I was planning on doing it in a later PR (but before we merge the branch to master). I have it on a list of things I need to address along with writing E2E tests.

If you're not comfortable approving this PR without those tests, I can go ahead and work on them now.

BertKleewein (approved):
:shipit:

device/transport/mqtt/src/mqtt.ts
BertKleewein: Fair enough.  I rememer all the pain that setOptions has caused in the past.  I did get burned by the unnecessary callback, and I dd get burned by the fact that it overwrites old options down in the transport.  I think you'll be fine either way you resolve this, and I hope we can eventually do better.

device/core/src/module_client.ts
BertKleewein: Yeah, I'm OK without the tests.  And I agree with what you say.

device/transport/mqtt/src/mqtt.ts
vishnureddy17: Actually, I'll stick with the `setModelId` method but add some comments explaining the reason behind it. That'd be clearer, I think.

PR #1002, opened 2021-07-22
feat(azure-iot-device): 🎸 Implement PnP telemetry

vishnureddy17:


device/core/test/_internal_client_test.js
```diff
@@ -780,6 +780,82 @@ var ModuleClient = require('../dist/module_client').ModuleClient;
...
+          assert.strictEqual(message.contentType, 'application/json');
+          assert.strictEqual(message.data, JSON.stringify(payload));
+          assert.strictEqual(message.properties.getValue('$.sub'), component);
+          assert.strictEqual(callback, telemetryCallback);
+        }
+        client.sendTelemetry(payload, component, telemetryCallback);
+      });
+    });
   });
 });
```
anthonyvercolano, line 861: Could you add a test that you can actually await on the sendTelemetry? - would entail you creating fake sendEvent that returns a valid result.

anthonyvercolano (approved):

device/core/test/_internal_client_test.js
vishnureddy17: On it

vishnureddy17: Done

device/core/test/_internal_client_test.js
```diff
@@ -780,6 +780,120 @@ var ModuleClient = require('../dist/module_client').ModuleClient;
...
+            assert.isUndefined(callback);
+          } catch (err) {
+            return Promise.reject(err);
+          }
+          return Promise.resolve(new results.MessageEnqueued());
+        }
+        client.sendTelemetry(payload).then(
+          function () {
+            testCallback();
+          },
```
anthonyvercolano, line 821: Isn't the normal pattern to have a .catch in the error case?

vishnureddy17: .catch is just syntactic sugar for the two argument .then

PR #1003, opened 2021-07-28
feat(azure-iot-device): PnP Commands

vishnureddy17:


device/core/src/pnp/command_request.ts
```diff
@@ -0,0 +1,36 @@
...
+     */
+    payload: JSONValue;
+
+    constructor(methodRequest: DeviceMethodRequest) {
+        this.requestId = methodRequest.requestId;
+        this.payload = methodRequest.payload;
+        const splitMethod = methodRequest.methodName.split('*');
+        if (splitMethod.length > 1) {
+            this.componentName = splitMethod[0];
+            this.commandName = splitMethod.slice(1).join('*');
```
BertKleewein, line 31: >splitMethod.slice(1).join('*'); [](http://example.com/codeflow?start=31&length=31)

`*` was chosen as a separator beause it's not valid for componentName, commandName, or methodName values.  You can juse ues `splitMethod[1]` here.

device/core/src/pnp/command_response.ts
```diff
@@ -0,0 +1,28 @@
+import { JSONValue } from '.';
+import { DeviceMethodResponse } from '../device_method';
+import { ErrorCallback } from 'azure-iot-common';
+
+export class CommandResponse {
+    private methodSend: (status: number, payload?: any | ErrorCallback, done?: ErrorCallback) => Promise<void> | void;
+
+    constructor(methodResponse: DeviceMethodResponse) {
+        this.methodSend = methodResponse.send.bind(methodResponse);
```
BertKleewein, line 9: >methodResponse.send.bind(methodResponse); [](http://example.com/codeflow?start=26&length=41)

Is the `bind` required here?  I assume it is, but I'm wondering why.

BertKleewein (approved):
:shipit:

device/core/src/pnp/command_response.ts
vishnureddy17: When you assign a function value to a variable like this, it looses its context. Using `bind()` here will assigns `this` so that `methodResponse.send()` works correctly.

device/core/src/pnp/command_request.ts
vishnureddy17: Will the service fail the method/command request if an attempt is made to use `*` in the names? If not, I think we should either leave the code as it is now or throw when the user tries to register a listener with `*` in the name. Thoughts?

vishnureddy17:
Merged, but I can incorporate feedback into the next PR if anyone has any.

PR #1009, opened 2021-08-05
fix: Fix bugs with TokenCredential refresh in RestApiClient and service Amqp transport

vishnureddy17:
Apologies for the `package.json` and `package-lock.json` pollution. It was needed to update Sinon to a newer version to get a needed feature for a test. I put it in a separate commit to make this easier.


service/src/amqp.ts
```diff
@@ -209,26 +209,27 @@ export class Amqp extends EventEmitter implements Client.Transport {
...
                   this.getToken().then((accessToken) => {
-                    this._amqp.putToken(audience, accessToken, (err) => {
+                    const tokenValue = this._bearerTokenPrefix + accessToken.token;
+                    this._amqp.putToken(this._iotHubPublicScope, tokenValue, (err) => {
                       if (err) {
                         /*Codes_SRS_NODE_IOTHUB_SERVICE_AMQP_06_004: [** If `putToken` is not successful then the client will remain disconnected and the callback, if provided, will be invoked with an error object.]*/
                         this._fsm.transition('disconnecting', err, callback);
                       } else {
-                        this._fsm.transition('authenticated', accessToken, callback);
+                        this._renewalNumberOfMilliseconds = (2 / 3) * (accessToken.expiresOnTimestamp - Date.now());
```
anthonyvercolano, line 231: Do some serious thinking about this _renewalNumberOfMilliseconds.  Can you think of any reason for it to go negative OR zero?  I would hate to get into a tight renewal loop.

anthonyvercolano (approved):

service/src/amqp.ts
vishnureddy17: Done

Issue #578, opened 2019-07-03
Creating a reported property called `update` breaks the twin update function


peolivei2:
According to SDK samples. to update a reported property, you should call the `twin.properties.reported.update` function. However, if I create a reported property called `update`, this overwrites the function and twin updates no longer work. See repro below:

```javascript
const Mqtt = require('azure-iot-device-mqtt').Mqtt;
const Client = require('azure-iot-device').Client;
const setTimeout = require('timers').setTimeout;

const connectionStr = '<redacted>';

const client = Client.fromConnectionString(connectionStr, Mqtt);

client.open(err => {
    if (err) {
        console.log('[LOG] Open failed');
        return;
    }

    console.log('[LOG] Client open. Getting twin...');

    client.getTwin((twinErr, deviceTwin) => {
        console.log('[LOG] getTwin callback', twinErr, deviceTwin);

        deviceTwin.properties.reported.update({ update: 'abc' }, function(err) {
            if (err) {
                throw err;
            };

            console.log('[LOG] Twin state reported');

            deviceTwin.properties.reported.update({ someprop: 'abc' }, function(err) {
                if (err) {
                    console.log('[LOG] Error updating reported property');
                    throw err;
                };
                
                console.log('[LOG] Twin state reported');
            });
        });
    });
    
    setTimeout(() => {
        client.close(() => {
            console.log('[LOG] Client closed. Getting twin...');
        });
    }, 10000);
});
```

Which outputs:

```
[LOG] Client open. Getting twin...
[LOG] getTwin callback null Twin {
  properties:
   { reported: { update: [Function: update], '$version': 1 },
     desired: { fanSpeed: [Object], '$version': 1 } } }
[LOG] Twin state reported
C:\Users\peolivei\Downloads\sdktst\tst.js:27
            deviceTwin.properties.reported.update({ someprop: 'abc' }, function(err) {
                                           ^

TypeError: deviceTwin.properties.reported.update is not a function
    at C:\Users\peolivei\Downloads\sdktst\tst.js:27:44
    at C:\Users\peolivei\Downloads\sdktst\node_modules\azure-iot-common\lib\retry_operation.js:54:21
    at C:\Users\peolivei\Downloads\sdktst\node_modules\azure-iot-device\lib\twin.js:122:21
[...]

PR #1013, opened 2021-08-12
feat(azure-iot-device): PnP Properties

vishnureddy17:


device/core/src/internal_client.ts
```diff
@@ -337,6 +337,42 @@ export abstract class InternalClient extends EventEmitter {
...
+    }, done);
+  }
+
+  onWritablePropertyUpdateRequest(callback: (properties: ClientPropertyCollection) => void): void {
+    this.getTwin((err, twin) => {
+      if (err) {
+        this.emit('error', err);
+        return;
+      }
+      twin.on('properties.desired', (patch) => {
```
vishnureddy17, line 360: The way this is implemented, the user can only have a listener for entire desired property patches. In contrast, the legacy twin API lets the user listen for specific properties.

I think this is the approach we want, but would like to hear thoughts on this.

device/core/src/pnp/properties.ts
```diff
@@ -0,0 +1,81 @@
+import { JSONValue, JSONObject } from '.';
+import { Twin } from '../twin';
+
+export function generateWritablePropertyResponse(value: JSONValue, ackCode: number, ackDescription: string, ackVersion: number): {
```
vishnureddy17, line 4: We don't often have functions exported like this, should we move this somewhere else or does this seem fine?

device/core/src/pnp/properties.ts
```diff
@@ -0,0 +1,81 @@
...
+    getProperty(propertyName: string): JSONValue;
+    getProperty(componentName: string, propertyName: string): JSONValue;
+    getProperty(propertyNameOrComponentName: string, propertyName?: string): JSONValue {
+        return propertyName ?
+            this.backingObject[propertyNameOrComponentName]?.[propertyName] :
+            this.backingObject[propertyNameOrComponentName];
+    }
+}
+
+export class ClientProperties {
```
vishnureddy17, line 73: `ClientProperties` has two `ClientPropertyCollections`, but I would think from the naming that its the other way around, i.e., `ClientPropertyCollection` is the bigger thing that contains `ClientProperties`.

I named it this way because that's how we did it in Python, but is this the way we want it?

device/core/src/internal_client.ts
BertKleewein: This way is better. I think the way we did it for the legacy API was a mistake in this regard.  No other SDK allows for per-property events.

device/core/src/pnp/properties.ts
```diff
@@ -0,0 +1,81 @@
...
+import { Twin } from '../twin';
+
+export function generateWritablePropertyResponse(value: JSONValue, ackCode: number, ackDescription: string, ackVersion: number): {
+    value: JSONValue,
+    ac: number,
+    ad: string,
+    av: number
+} {
+    return {
+        value,
```
BertKleewein, line 11: >value [](http://example.com/codeflow?start=8&length=5)

the schema specifies that you tag it with the name `value`.  e.g. `return { value: value, `.  Is this name somehow implied in JS or TS syntax when you use it like this?

device/core/src/pnp/properties.ts
BertKleewein: Yes. The names are not great, but they give us cross-language parity

device/core/src/pnp/properties.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+export class ClientPropertyCollection {
+    /**
+     * The object representation of the properties.
+     */
+    backingObject: JSONObject = {};
+
+    constructor(backingObject?: JSONObject) {
+        this.backingObject = backingObject;
```
BertKleewein, line 25: >backingObject [](http://example.com/codeflow?start=29&length=13)

I assume this syntax leaves `this.backingObject == undefined` if you don't pass in an object.  Would `this.backingObject = backingObject or {}` be better?

vishnureddy17: Yes, I just noticed that as well and it's fixed now. Good catch.

device/core/src/pnp/properties.ts
vishnureddy17: This is a feature called Object Literal Shorthand Syntax:
https://alligator.io/js/object-property-shorthand-es6/

device/core/src/pnp/json.ts
```diff
@@ -5,3 +5,5 @@ export type JSONValue =
  | null
  | JSONValue[]
  | {[key: string]: JSONValue};
+
+export type JSONObject = {[key: string]: JSONValue};
```
BertKleewein, line 9: >JSONObject [](http://example.com/codeflow?start=12&length=10)

super nit-picky semantic comment.  Technically, These are not JSON anything.  JSON is the text representation of native objects.  You serialize objects of these types into JSON, but you could also serialize them into other forms.  I don't know what a better name would be.  Are they PnpCompatibleObjects and PnpCompatibleValues?  (I don't rememer if we can expose "Pnp" in type names or not.)

device/core/src/pnp/properties.ts
BertKleewein: Cool!  I assume that our typescript config correctly transpiles ES6/ES2015 (or doesn't need to transpile this because of our supported Node versions).

device/core/src/pnp/properties.ts
```diff
@@ -0,0 +1,81 @@
...
+     * Assigns the provided value to a specified property for a component.
+     * If no component is specified, the default component is assumed.
+     *
+     * @param {string} componentName The name of the component.
+     * @param {string} propertyName  The name of the property.
+     * @param {JSONValue} value      The value to assign to the property.
+     */
+    setProperty(propertyName: string, value: JSONValue): void;
+    setProperty(componentName: string, propertyName: string, value: JSONValue): void;
+    setProperty(propertyNameOrComponentName: string, valueOrPropertyName: JSONValue, value?: JSONValue): void {
```
BertKleewein, line 46: >propertyNameOrComponentName: [](http://example.com/codeflow?start=16&length=28)

I'm not sure yet and I'm asking the cross-language cabal., but we may want to skip the "component name is optional" overloading and split this method (and `getProperty`) into `setRootProperty` and `setComponentProperty`

device/core/src/pnp/properties.ts
vishnureddy17: Yep, it transpiles down to ES5

device/core/src/pnp/json.ts
vishnureddy17: Changed to `JSONSerializableObject` and `JSONSerializableValue`

device/core/src/pnp/properties.ts
vishnureddy17: Resolved over Teams. Keeping it the way it is.

device/core/src/pnp/commands.ts
```diff
@@ -0,0 +1,62 @@
...
+   */
+  payload: JSONSerializableValue;
+
+  constructor(methodRequest: DeviceMethodRequest) {
+    this.requestId = methodRequest.requestId;
+    this.payload = methodRequest.payload;
+    const splitMethod = methodRequest.methodName.split('*');
+    if (splitMethod.length > 1) {
+      this.componentName = splitMethod[0];
+      this.commandName = splitMethod.slice(1).join('*');
```
anthonyvercolano, line 32: This is unclear.

device/core/src/pnp/properties.ts
```diff
@@ -0,0 +1,92 @@
+import { JSONSerializableValue, JSONSerializableObject } from '.';
```
anthonyvercolano, line 1: '.' ??

device/core/src/pnp/properties.ts
```diff
@@ -0,0 +1,92 @@
...
+ * @param {number}                ackVersion     The av field of the response object. Usually equal to the received version number in the writable property request.
+ */
+export function generateWritablePropertyResponse(value: JSONSerializableValue, ackCode: number, ackDescription: string, ackVersion: number): {
+  value: JSONSerializableValue,
+  ac: number,
+  ad: string,
+  av: number
+} {
+  return {
+    value,
```
anthonyvercolano, line 19: This value seems odd.

device/core/src/pnp/properties.ts
```diff
@@ -0,0 +1,92 @@
...
+   * If no component is specified, the default component is assumed.
+   *
+   * @param {string}                componentName The name of the component.
+   * @param {string}                propertyName  The name of the property.
+   * @param {JSONSerializableValue} value         The value to assign to the property.
+   */
+  setProperty(propertyName: string, value: JSONSerializableValue): void;
+  setProperty(componentName: string, propertyName: string, value: JSONSerializableValue): void;
+  setProperty(propertyNameOrComponentName: string, valueOrPropertyName: JSONSerializableValue, value?: JSONSerializableValue): void {
+    if (value === undefined) { // We cannot just check the truthiness of value because null is a valid argument for value
```
anthonyvercolano, line 56: I think you should describe more the structure that you are creating here.  More explanation on why your normalization is correct.

device/core/src/internal_client.ts
```diff
@@ -162,15 +162,17 @@ export abstract class InternalClient extends EventEmitter {
...
-  sendTelemetry(payload: JSONValue, sendTelemetryCallback: Callback<results.MessageEnqueued>): void;
-  sendTelemetry(payload: JSONValue, componentName: string, sendTelemetryCallback: Callback<results.MessageEnqueued>): void;
-  sendTelemetry(payload: JSONValue, callbackOrComponent?: Callback<results.MessageEnqueued> | string, sendTelemetryCallback?: Callback<results.MessageEnqueued>): Promise<results.MessageEnqueued> | void {
+  sendTelemetry(payload: JSONSerializableValue): Promise<results.MessageEnqueued>;
+  sendTelemetry(payload: JSONSerializableValue, componentName: string): Promise<results.MessageEnqueued>;
+  sendTelemetry(payload: JSONSerializableValue, sendTelemetryCallback: Callback<results.MessageEnqueued>): void;
+  sendTelemetry(payload: JSONSerializableValue, componentName: string, sendTelemetryCallback: Callback<results.MessageEnqueued>): void;
+  sendTelemetry(payload: JSONSerializableValue, callbackOrComponent?: Callback<results.MessageEnqueued> | string, sendTelemetryCallback?: Callback<results.MessageEnqueued>): Promise<results.MessageEnqueued> | void {
     let componentName: string;
```
anthonyvercolano, line 176: This is really confusing.  You have a parameter named componentName and then you have a local named componentName?!!  It would seem as though the local obscures the parameter via scoping rules.

device/core/src/internal_client.ts
```diff
@@ -337,6 +339,68 @@ export abstract class InternalClient extends EventEmitter {
...
+  getClientProperties(): Promise<ClientProperties>;
+  getClientProperties(done: Callback<ClientProperties>): void;
+  getClientProperties(done?: Callback<ClientProperties>): Promise<ClientProperties> | void {
+    return callbackToPromise((_callback) => {
+      this.getTwin((err, twin) => {
+        if (err) {
+          _callback(err);
+          return;
+        }
+        if (typeof twin.properties.reported.update === 'function') {
```
anthonyvercolano, line 396: I think I know what you are doing here but you gotta explain this.

device/core/src/internal_client.ts
```diff
@@ -381,23 +445,26 @@ export abstract class InternalClient extends EventEmitter {
...
+   * @param {string}   commandName   The name of the command.
+   * @param {string}   componentName The name of the component that corresponds with the command.
+   * @param {function} callback      The callback that is called each time a command request for this command is received.
    */
   onCommand(commandName: string, callback: (request: CommandRequest, response: CommandResponse) => void): void;
-  onCommand(commandName: string, componentName: string, callback: (request: CommandRequest, response: CommandResponse) => void): void;
-  onCommand(commandName: string, callbackOrComponent: ((request: CommandRequest, response: CommandResponse) => void) | string, callback?: (request: CommandRequest, response: CommandResponse) => void): void {
+  onCommand(componentName: string, commandName: string, callback: (request: CommandRequest, response: CommandResponse) => void): void;
+  onCommand(commandOrComponent: string, callbackOrCommand: ((request: CommandRequest, response: CommandResponse) => void) | string, callback?: (request: CommandRequest, response: CommandResponse) => void): void {
+    let commandName: string;
```
anthonyvercolano, line 455: Maybe it's the wine, but it looks like you're obscuring the parameter with the local.

anthonyvercolano: I mean are we doing this other places?

device/core/src/internal_client.ts
```diff
@@ -381,23 +445,26 @@ export abstract class InternalClient extends EventEmitter {
...
     let componentName: string;
-    if (typeof callbackOrComponent === 'function') {
-      callback = callbackOrComponent;
-    } else if (typeof callbackOrComponent === 'string') {
-      componentName = callbackOrComponent;
+    if (typeof callbackOrCommand === 'function') {
```
anthonyvercolano, line 457: I generally think I follow normalization, but I'm having trouble going step by step here.

anthonyvercolano (changes requested):

device/core/src/internal_client.ts
```diff
@@ -337,6 +339,68 @@ export abstract class InternalClient extends EventEmitter {
...
+        return;
+      }
+      twin.on('properties.desired', (patch) => {
+        callback(new ClientPropertyCollection(patch));
+      });
+    });
+  }
+
+  /**
+   * Gets the client properties the Azure IoT Hub or Azure IoT Edge Hub service.
```
vishnureddy17, line 381: There's a typo here

vishnureddy17: Fixed

device/core/src/internal_client.ts
YoDaMa: the naming for sure could be better, but it looks like because the previous definitions are overloads, there is not really a parameter named componentName. The user will see componentName as an option but the code itself operates on callbackOrComponent...

device/core/src/internal_client.ts
vishnureddy17: Resolved by adding more comments

device/core/src/internal_client.ts
```diff
@@ -337,6 +339,68 @@ export abstract class InternalClient extends EventEmitter {
...
+  updateClientProperties(propertyCollection: ClientPropertyCollection): Promise<void>;
+  updateClientProperties(propertyCollection: ClientPropertyCollection, done: (err: Error) => void): void;
+  updateClientProperties(propertyCollection: ClientPropertyCollection, done?: (err: Error) => void): Promise<void> | void {
+    return callbackToPromise((_callback) => {
+      this.getTwin((err, twin) => {
+        if (err) {
+          _callback(err);
+          return;
+        }
+        (twin as any)._updateReportedProperties(propertyCollection.backingObject, _callback);
```
YoDaMa, line 358: do we have a type for twin? avoid using `any`

device/core/src/internal_client.ts
vishnureddy17: Discussed over a call, but basically the answer is that `componentName` is not actually initially declared in the function scope.  Otherwise, the `let componentName: string` line would cause a compiler error. The initial variables are `payload`, `callbackOrComponent`, and `sendTelemetryCallback`.

device/core/src/internal_client.ts
```diff
@@ -337,6 +339,68 @@ export abstract class InternalClient extends EventEmitter {
...
+      });
+    }, done);
+  }
+
+  /**
+   * Registers a listener to get invoked with a writable property patch whenever the device receives a writable property request.
+   *
+   * @param {(properties: ClientPropertyCollection) => void} callback The callback to get invoked on a writable property request.
```
YoDaMa, line 366: fix this js doc to include a - between callback and The callback. Like: 

@param {string} title - The title of the book.

device/core/src/internal_client.ts
```diff
@@ -337,6 +339,68 @@ export abstract class InternalClient extends EventEmitter {
...
+        callback(new ClientPropertyCollection(patch));
+      });
+    });
+  }
+
+  /**
+   * Gets the client properties the Azure IoT Hub or Azure IoT Edge Hub service.
+   * This method is only intended for use with Azure IoT Plug and Play.
+   *
+   * @param {Callback<ClientProperties>} done The callback which gets invoked with the ClientProperties object.
```
YoDaMa, line 384: jsdoc format here too

device/core/src/internal_client.ts
vishnureddy17: It's done this way because `_updateReportedProperties` is actually a private member of `Twin` and tsc would complain about it if you took out the `as any`. 

We could have called the `properties.reported.update` function instead, but this is a good opportunity to get rid of the [limitation where having a reported property called `update` would break things](https://github.com/Azure/azure-iot-sdk-node/issues/578). 

And IMO we should keep it private because we expose `Twin` directly to the user with the "legacy" API, and we don't want user's calling that directly. 

I'll add a comment explaining why the `as any` is there.

device/core/src/pnp/properties.ts
vishnureddy17: Resolved over a call

device/core/src/internal_client.ts
vishnureddy17: See the my other reply to a similar comment.

device/core/src/internal_client.ts
vishnureddy17: @YoDaMa , do you have any suggestions on alternate parameter names? Maybe something like `_0`, `_1`, `_2` or `first`, `second`, `third`?

device/core/src/pnp/commands.ts
vishnureddy17: Added comments to make this clearer

device/core/src/pnp/properties.ts
vishnureddy17: Changed it to `value: value`

device/core/src/pnp/properties.ts
vishnureddy17: Added some explanation in the form of comments

device/core/src/internal_client.ts
vishnureddy17: Added a comment to explain this (a whole paragraph was needed to fully explain)

device/core/src/internal_client.ts
vishnureddy17: Fixed here and in a bunch of other places

device/core/src/internal_client.ts
vishnureddy17: Fixed here and in a bunch of other places

device/core/src/internal_client.ts
anthonyvercolano: I wouldn't change the names.

YoDaMa (approved):
lgtm

PR #1014, opened 2021-08-15
Update to 12

anthonyvercolano:


common/transport/http/src/http.ts
```diff
@@ -107,7 +107,7 @@ export class Http {
     // Codes_SRS_NODE_HTTP_13_004: [ Use the request object from the `options` object if one has been provided or default to HTTPS request. ]
     let request = https_request;
     if (requestOptions && requestOptions.request) {
-      request = requestOptions.request;
+      request = requestOptions.request as any;
```
vishnureddy17, line 110: Can you explain why this was needed?

vishnureddy17 (approved):

vishnureddy17 (approved):
Not sure why `node-canary` is failing

common/transport/http/src/http.ts
anthonyvercolano: Because typescript didn't like it otherwise.  I couldn't quite figure out why.  Since I felt that this code worked previously, I hit it with the big hammer.

YoDaMa:
looks like this is an error in a TPM test. Let's rerun and see if that helps.

YoDaMa:
1) tpm
       getEndorsementKey
         returns the endorsement key:
     Error: Cannot find module 'ref'
This is the error FYI.

PR #1017, opened 2021-08-24
fix(azure-iot-device): only fetch twin when necessary, don't fire events when fetching twin

vishnureddy17:


device/core/test/_internal_client_test.js
```diff
@@ -790,6 +873,27 @@ var ClientPropertyCollection = require('../dist/pnp/properties').ClientPropertyC
         assert.deepEqual(twin._updateReportedProperties.args[0][0], {key: "value"});
       });
```
anthonyvercolano, line 875: I don't see any tests where you pass a first parameter to get that ISN'T a function or a boolean type.

anthonyvercolano (approved):
Approving assuming that you will and the parameter 1 type checking for get.

anthonyvercolano:
*you will ADD the ...

PR #1027, opened 2021-09-10
Update create_test_cert to produce PFX files

rido-min:
When creating test certificates, produce the PFX container with the private key to be used in platform/runtimes where PEM files are not available (eg, .NET)


YoDaMa (approved):
lgtm

YoDaMa:
thanks for the contribution @rido-min

PR #1028, opened 2021-09-13
Adding GitHub Codespace support 

danhellem:
# Checklist

# Description of the problem
Add codespace support

# Description of the solution
- Added codespace files (devcontainer.json and docker file)
- Updated the root package.json file and added new script that gets executed after install of the codespace. This runs npm install command in the samples and the samples/pnp folders (we can add more as needed)
- update samples/read.me file


YoDaMa (approved):
lgtm

danhellem:
@YoDaMa looks like some odd node install issue. Should I let that block the merge?

YoDaMa:
> @YoDaMa looks like some odd node install issue. Should I let that block the merge?

yep! if there's a problem in the gate we'll have to investigate. it's in the postinstall you added.

PR #1031, opened 2021-09-15
test: add E2E tests for PnP and missing fromEnvironment() unit tests

vishnureddy17:


device/core/test/_module_client_test.js
```diff
@@ -49,24 +50,39 @@ describe('ModuleClient', function () {
   });
 
   describe('#fromEnvironment', function () {
-    // Tests_SRS_NODE_MODULE_CLIENT_13_033: [ The fromEnvironment method shall throw a ReferenceError if the callback argument is not a function. ]
+    [{}, 20].forEach(function (badArgument) {
+      it('throws if the second argument is neither a string nor a function', function () {
```
BertKleewein, line 54: This makes me think that we should 100% stop overloading functions as soon as we can (like v.next). Do you agree?

BertKleewein: Because I can't read these tests (or the implementations themselves) and know that they're correct without pulling out pencil and paper and giving myself a headache by going through the entire set of possibilities.  If we decide that we do want to overload functions, we should find a better way to do it.

e2etests/test/testUtils.js
```diff
@@ -4,12 +4,12 @@
 'use strict';
 var deviceSdk = require('azure-iot-device');
 
-function createDeviceClient(deviceTransport, provisionedDevice) {
+function createDeviceClient(deviceTransport, provisionedDevice, modelId) {
```
BertKleewein, line 7: >provisionedDevice [](http://example.com/codeflow?start=45&length=17)

I should know this better, but if you're using DPS to create the device, the modelId goes into the provisioning payload somewhere.  I think this is for IoTCentral, so it may not be strictly 100% necessary, but, we should make sure we do in production code, and we should make sure our E2E tests are doing this too.  (At least I _think_ our E2E tests should do this.)

e2etests/test/pnp_connect.js
```diff
@@ -0,0 +1,60 @@
...
+[
+  DeviceIdentityHelper.createDeviceWithSas,
+  DeviceIdentityHelper.createDeviceWithSymmetricKey,
+  DeviceIdentityHelper.createDeviceWithX509SelfSignedCert,
+  DeviceIdentityHelper.createDeviceWithX509CASignedCert
+].forEach((createDeviceMethod) => {
+  [deviceMqtt.Mqtt, deviceMqtt.MqttWs].forEach((deviceTransport) => {
+    pnpConnectTests(deviceTransport, createDeviceMethod);
+  });
+});
```
BertKleewein, line 26: A suggestion (take it or leave it).  I see this same block of code repeated 5 times.  I feel like it could be a helper function.  (or you could share the 2 arrays that make up the matrix between tests.)

e2etests/test/pnp_connect.js
```diff
@@ -0,0 +1,60 @@
...
+    beforeEach(async function () {
+      deviceClient = createDeviceClient(deviceTransport, provisionedDevice, modelId);
+      await deviceClient.open();
+    });
+
+    afterEach(async function () {
+      await deviceClient.close();
+    });
+
+    it('connects to the Hub with the model name', async function () {
```
BertKleewein, line 55: >it('connects to the Hub with the model name', async function () { [](http://example.com/codeflow?start=4&length=65)

Just to make sure I understand, this test file (pnp_connect.js) ends up creating 8 device instances and those devices are only used for this one test.  Is that right?  "Yes" is the correct answer because we don't want to be re-using devices for this specific test.

e2etests/test/pnp_telemetry.js
```diff
@@ -0,0 +1,178 @@
...
+        rdv.imIn('ehClient');
+        return ehClient.getPartitionIds();
+      }).then(function (partitionIds) {
+        partitionIds.forEach(function (partitionId) {
+          ehClient.receive(partitionId, onEventHubMessage, onEventHubError, { eventPosition: EventPosition.fromEnqueuedTime(startAfterTime) });
+        });
+        return new Promise(function (resolve) {
+          setTimeout(resolve, 3000);
+        });
+      }).then(function () {
```
BertKleewein, line 95: Another suggestion:  Every time I see code that iterates over eventhub partitions, I think that this code needs to be refactored.  How often do we do this in E2E tests?  I'm guessing a lot.

e2etests/test/pnp_telemetry.js
```diff
@@ -0,0 +1,178 @@
...
```
BertKleewein, line 156: So.  Much.  Copy.  Paste.  (say it in William Shatner's voice).  

(not really your fault because this is a well-worn pattern in these E2E tests long before you got here, but it's a habit we need to stop.)

e2etests/test/pnp_commands.js
```diff
@@ -0,0 +1,105 @@
...
+
+const hubConnectionString = process.env.IOTHUB_CONNECTION_STRING;
+const commandName = 'fakeCommand';
+const componentName = 'fakeComponent';
+const requestPayload = {this: 'is the payload for the request'};
+const responsePayload = {payload: 'for the response'};
+const statusCode = 200;
+
+[
+  DeviceIdentityHelper.createDeviceWithSas,
```
BertKleewein, line 21: Do we want to expand these tests to also run with module clients?

e2etests/test/pnp_properties_requests.js
```diff
@@ -0,0 +1,75 @@
...
+    beforeEach(async function () {
+      deviceClient = createDeviceClient(deviceTransport, provisionedDevice);
+      await deviceClient.open();
+    });
+  
+    afterEach(async function () {
+      await deviceClient.close();
+    });
+
+    it('receives writable property update requests and calls the listener', async function () {
```
BertKleewein, line 53: No test for the PNP equivelant of getTwin?

BertKleewein (approved):
:shipit:

device/core/test/_module_client_test.js
vishnureddy17: Hmm, I'm not sure if I agree with 100% stopping overloading, but I do agree that it is very problematic here. Here, I think making the callback the last argument no matter what makes the overloading way more painful than it needs to be.

Much of our usage of overloading has to do with the presence or absence of the callback parameter. If and when we get rid of the callback APIs (say, for v next), we would do far less overloading and if we ever do have overloading for something like adding an additional optional parameter it will be less problematic.

e2etests/test/testUtils.js
vishnureddy17: I don't think we do use DPS, actually. We just add the device to the device registry.

vishnureddy17: `provisionedDevice` is poor naming, I guess. I'll change it.

e2etests/test/pnp_connect.js
vishnureddy17: Correct.

e2etests/test/pnp_telemetry.js
vishnureddy17: We do it a lot. There's actually already the `EventHubReceiverHelper` written that I can use that would make things a lot simpler. I'll refactor my tests.

e2etests/test/pnp_telemetry.js
vishnureddy17: Refactored the test to use `EventHubReceiverHelper` to be cleaner and avoid repeating code.

e2etests/test/pnp_telemetry.js
vishnureddy17: Done

e2etests/test/pnp_connect.js
vishnureddy17: Will probably leave it as-is. IMO it's simple enough to where a helper function wouldn't be that useful and might make things a bit less clear

e2etests/test/testUtils.js
vishnureddy17: Done

e2etests/test/pnp_properties_requests.js
vishnureddy17: Added test for that

e2etests/test/pnp_commands.js
vishnureddy17: Yes, I've refactored things so the tests I wrote will work for both `Client` and `ModuleClient`

e2etests/test/pnp_connect.js
vishnureddy17: Never mind, I made a helper function now that `ModuleClients` are in the mix

e2etests/test/pnp_get_properties.js
```diff
@@ -0,0 +1,71 @@
...
+    beforeEach(async function () {
+      await client.open();
+    });
+  
+    afterEach(async function () {
+      await client.close();
+    });
+
+    it('successfully gets the twin from the service', async function () {
+      const updateTwinFunc = registryClient[isModule ? 'updateModuleTwin' : 'updateTwin'].bind(registryClient, deviceId, ...(isModule ? [moduleId] : []));
```
anthonyvercolano, line 53: Should you be using the DigitalTwin api's?  service/src/cl/*
They use a different format for updating the properties.  JSON update? I think.
See service/samples/javascript/update_digital_twin.js

anthonyvercolano (approved):
Strongly consider using the DigitalTwin service API

e2etests/test/pnp_get_properties.js
vishnureddy17: That doesn't work for modules, though.

PR #1032, opened 2021-09-20
fix: update debug logging

BertKleewein:
These changes are inspired by my current ICM.  I think they would be useful to apply across the board.  I'm fixing a few things:

1.  For processes with multiple device clients, there's no way to correlate which azure-iot* level operations correspond to which rhea operations.  Adding the `.toString()` text to `sender_link.ts` and `receiver_link.ts` adds a `[connection-n]` prefix to our logs so it's easier to filter on a specific connection.
2. There are  no consistent "what did the customer call and did it succeed" logs.  `azure-iot-common:RetryOperation` output is part-way there, but it needs more detail.  I added lines here so we can see when an operation starts, when it finishes, if it fails, etc..
3. There is no concept of "debug levels", so we can't set `DEBUG` values to show only errors, for instance.

Since my current runs produce 10-100 GB of log per run, I want to filter output as much as possible.  With this code, I can run this with:
```
export DEBUG="rhea:frames*,azure-iot-common:RetryOperation*,azure*:Errors"
```
This way I get:
1. Rhea traffic (`rhea:frames*`)
2. What is causing that traffic. (`azure-iot-common:RetryOperation*`)
3. What error code gets called as a result of this traffic.  (`azure*:Errors"`)


common/transport/amqp/src/amqp_cbs.ts
```diff
@@ -193,7 +193,7 @@ export class ClaimsBasedSecurityAgent extends EventEmitter {
               callback();
             }
             let tokenOperation = this._putTokensNotYetSent.shift();
-            while (tokenOperation) {
+            while (tokenOperation) {debug
```
vishnureddy17, line 196: Typo here

e2etests/test/device_identity_helper.js
```diff
@@ -16,8 +16,8 @@ var Registry = require('azure-iothub').Registry;
 var hubConnectionString = process.env.IOTHUB_CONNECTION_STRING;
 
 var registry = Registry.fromConnectionString(hubConnectionString);
-var CARootCert = Buffer.from(process.env.IOTHUB_CA_ROOT_CERT, 'base64').toString('ascii');
-var CARootCertKey = Buffer.from(process.env.IOTHUB_CA_ROOT_CERT_KEY, 'base64').toString('ascii');
+// var CARootCert = Buffer.from(process.env.IOTHUB_CA_ROOT_CERT, 'base64').toString('ascii');
+// var CARootCertKey = Buffer.from(process.env.IOTHUB_CA_ROOT_CERT_KEY, 'base64').toString('ascii');
```
vishnureddy17, line 20: Probably don't want to check this in

vishnureddy17:
I'm a fan of this. Since the debug library doesn't provide log levels, having something like this `:Errors` patten would be useful more broadly across our libraries. I also like the changes you made to be able to filter on a specific connection. Could we apply this to our other transports too?

Also, there's a logfile in this PR that doesn't seem like it is meant to be there.

common/core/src/retry_operation.ts
```diff
@@ -7,6 +7,9 @@ import * as errors from './errors';
 import { RetryPolicy } from './retry_policy';
 import * as dbg from 'debug';
 const debug = dbg('azure-iot-common:RetryOperation');
+const debugErrors = dbg('azure-iot-common:RetryOperation:Errors');
```
olivakar, line 10: if this is errors , then why not name the constant as `errors` ?

common/core/src/retry_operation.ts
```diff
@@ -17,18 +20,20 @@ export class RetryOperation {
...
   /**
    * Creates an instance of {@link azure-iot-common.RetryOperation.}
    * @param {RetryPolicy} policy The retry policy to be used for this operation, which determines what error is "retryable" or not and how fast to retry.
    * @param {number} maxTimeout  The maximum timeout for this operation, after which no retry will be attempted.
    */
-  constructor (policy: RetryPolicy, maxTimeout: number) {
-    if (policy && policy.constructor && policy.constructor.name === 'NoRetry') {
-      debug('A RetryOperation is being used with a NoRetry policy. The operation will not be retried on failure.');
-    }
+  constructor (name: string, policy: RetryPolicy, maxTimeout: number) {
```
olivakar, line 31: the inclusion of name is cool

YoDaMa:
I agree with what @vishnureddy17 previously said. The addition of some Errors debugging level would be useful. I also have no issue with finer grained specificity for each link on our RHEA logs. Overall these will make our life easier with debugging.

common/core/src/retry_operation.ts
```diff
@@ -17,18 +20,20 @@ export class RetryOperation {
...
+  private _name: string = '';
+  private _id: number = 0;
 
   /**
    * Creates an instance of {@link azure-iot-common.RetryOperation.}
    * @param {RetryPolicy} policy The retry policy to be used for this operation, which determines what error is "retryable" or not and how fast to retry.
    * @param {number} maxTimeout  The maximum timeout for this operation, after which no retry will be attempted.
    */
-  constructor (policy: RetryPolicy, maxTimeout: number) {
-    if (policy && policy.constructor && policy.constructor.name === 'NoRetry') {
```
anthonyvercolano, line 27: We absolutely have to keep this log in the code.  This is a specific request from CSS.

anthonyvercolano (approved):

PR #1034, opened 2021-09-22
samples typescript

danhellem:
DRAFT ONLY - DO NOT MERGE


YoDaMa (commented):
this is a good change. I support it. not offering any approvals or other comments yet since still in draft mode.

danhellem:
@YoDaMa thanks, I will keep working on this then in a separate branch and will create a PR when ready

PR #1035, opened 2021-09-23
fix: update error logging

BertKleewein:
These are changes I needed to make to debug my current ICM.   Mostly this changes debug logging that says "error is [object Object]" to something more meaningful.

### DIFF
--- a/common/transport/amqp/src/sender_link.ts
+++ b/common/transport/amqp/src/sender_link.ts
@@ -6,6 +6,7 @@ import { EventContext, AmqpError, Session, Sender, SenderOptions } from 'rhea';
 import { results } from 'azure-iot-common';
 import { AmqpMessage } from './amqp_message';
 import { AmqpLink } from './amqp_link_interface';
+import { getErrorName } from './amqp_common_errors';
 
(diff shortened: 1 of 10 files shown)


common/transport/amqp/src/sender_link.ts
```diff
@@ -315,7 +316,7 @@ export class SenderLink extends EventEmitter implements AmqpLink {
             this._indicatedError = undefined;
             this._senderCloseOccurred = true;
             if (error) {
-              debugErrors(this.toString() + ': In sender attached state - close event for ' + context.sender.name + ' already indicated error is: ' + this._getErrorName(error));
+              debugErrors(this.toString() + ': In sender attached state - close event for ' + context.sender.name + ' already indicated error is: ' + getErrorName(error));
```
vishnureddy17, line 319: _very minor_ nit: Template literals would be more readable and slightly shorter than concatenating strings, i.e.,
```suggestion
              debugErrors(`${this.toString()}: In sender attached state - close event for ${context.sender.name} already indicated error is: ${getErrorName(error)}`);
```

vishnureddy17 (approved):

PR #1036, opened 2021-09-24
fix: handle dropped connection failures

BertKleewein:


common/transport/amqp/package.json
```diff
@@ -59,10 +59,10 @@
       ".tsx"
     ],
     "check-coverage": true,
-    "lines": 96,
+    "lines": 95,
     "functions": 94,
-    "branches": 85,
-    "statements": 96
+    "branches": 83,
```
BertKleewein, line 64: >branches [](http://example.com/codeflow?start=5&length=8)

I plan to update unit tests to bring the coverage numbers back up

sdklab/longhaultests/package.json
```diff
@@ -9,7 +9,7 @@
     "lint": "tslint --project . -c ../../tslint.json",
     "build": "tsc",
     "setup_links": "npm link azure-iot-common azure-iot-device azure-iot-device-amqp azure-iot-device-mqtt azure-iot-device-http azure-iothub",
-    "start": "node lib/iothub_longhaul.js"
+    "start": "node dist/iothub_longhaul.js"
```
BertKleewein, line 12: Ignore everything in sdklab for now.  I'm not planning to pull any of this right now.

common/transport/amqp/src/sender_link.ts
```diff
@@ -314,6 +315,22 @@ export class SenderLink extends EventEmitter implements AmqpLink {
...
+            if (!error) {
+              ///
+              // question for reviewers: Is this a Rhea bug?
+              //
+              // https://github.com/amqp/rhea/blob/e826e8bbffd410eaa5a1efc70650f37c14069b23/lib/link.js#L161-L163
+              //
+              // when they fire the link_error and link_close events, I think context.error should be set.
+              // Since it's not set, we have to guess that there's an error by looking at link state here.
+              //
+              // I think my change is OK?  Do you agree?  Should we push for a Rhea fix?
```
vishnureddy17, line 327: This seems reasonable. Is there an issue filed against rhea for this? If so, we could keep an eye on it and remove this if they fix it

common/transport/amqp/src/amqp_common_errors.ts
```diff
@@ -110,6 +110,30 @@ export function translateError(message: string, amqpError: Error): AmqpTransport
...
     error = new Error(errorString);
+  } else if (amqpError.constructor?.name) {
+    //
+    // Questions for reviewers:
+    //
+    // 1. We don't want to translate these errors because we already have the right error,
+    //    but, what do we do with the passed-in message?  Do we lose that info, or do we
+    //    have some sort of "inner exception". e.g. if the passed error is
+    //    "NotConnectedError: SenderLink disconnected", and message is "Could not send",
+    //    What do we want to return?
```
vishnureddy17, line 121: One possibility is to add a property to the error object called `innerError` with the passed in message. We'd have to leave a comment somewhere so the person debugging in the future knows that this message is there.

common/transport/amqp/src/amqp_common_errors.ts
```diff
@@ -110,6 +110,30 @@ export function translateError(message: string, amqpError: Error): AmqpTransport
...
+    switch (amqpError.constructor.name) {
+      case 'ServiceUnavailableError':
+      case 'NotConnectedError':
+        error = amqpError;
+        amqpError = null;
+        break;
+      default:
+        error = new Error(message);
+        break;
+    }
```
vishnureddy17, line 136: For the `ServiceUnavailableError` case, this just ends up in the `default` case. This whole thing would could just be an if-else.

I think you're aware of it and left this unfinished in case you need to deal with more error cases, but pointing it out just in case.

YoDaMa: > For the `ServiceUnavailableError` case, this just ends up in the `default` case. This whole thing would could just be an if-else.

is this true? because there's no break happening between 'ServiceUnavailableError' and 'NotConnectedError' they both will be handled by the no-translation case of 'NotConnectedError'. See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#methods_for_multi-criteria_case

Unless I'm misunderstanding the comment you made?

common/transport/amqp/src/sender_link.ts
```diff
@@ -314,6 +315,22 @@ export class SenderLink extends EventEmitter implements AmqpLink {
...
+              //
+              const senderState = (context.sender as any).state;
```
YoDaMa, line 329: is context.sender unknown? why are we coercing it to any?

YoDaMa:
So the gist of this draft PR seems to be the code in two places: 

-   there should be a context.error set from RHEA when the link_error and link_close events are fired. That is not currently the case. _this is a bug in RHEA._
-  `ServiceUnavailableError` and `NotConnectedError's` should not be packaged inside another Error object. 

In both cases these are bugs in our handling of RHEA errors for dropped connections, and we want to just expose those errors to the user, since there doesn't seem to be any other logic being altered here.

Am I missing something?

common/transport/amqp/src/sender_link.ts
BertKleewein: context.sender is of type `Sender` from `link.d.ts` which doesn't have a state property defined.  I guess that rhea bug, if we open it, could also update the typing to add `Sender.state`.

BertKleewein:
Change #3: link detach errors were being dropped and instead generic `Error` objects were being returned.  Changes made to `amqp_cbs.ts` to pass the errors up, and sender_link `detached._on_enter` changed to return `ServiceUnavailableError` as a fallback error for detached links (was `Error`, which is not "retryable")

---
In reply to: [927946348](https://github.com/Azure/azure-iot-sdk-node/pull/1036#issuecomment-927946348) [](http://example.com/codeflow?ancestors=927946348,927946348)

common/transport/amqp/src/amqp_common_errors.ts
vishnureddy17: @YoDaMa Suppose the value of `amqpError.constructor.name` is `'ServiceUnavailableError'`. Then it will match with the first case of the switch. There are no statements for it (and importantly, that means there is no `break`). Therefore, it will fall it end up in the `default` case.

YoDaMa: hrm but is your assumption of how switch cases correct? I could be wrong, but JS supports fallthrough behavior in the switches. Here's an excerpt from something: 

> JavaScript supports the fall-through behavior in switch. If a case statement does not have a break;, the code for the next cases are also executed until a break; is found. This is why the break; statement is crucial when using the switch statement. If the break; statement does not exist, all the cases after the "correct" case are executed until the last case, including the default case. This defeats the purpose of the switch statement as the statement is used when you have multiple outcomes and only want to pinpoint one outcome, not all of them.

vishnureddy17: Yes, I realized my understanding of `switch` was wrong right after I sent that message. :)

YoDaMa: saw your edit :) cool! That being said, your confusion is justified, and I think linters can get mad about fallthrough.

See: https://eslint.org/docs/rules/no-fallthrough , which is enabled by default. Essentially it says "don't do this whole fallthrough thing, unless you tell us you're doing it with a comment", because it as error prone aspect of Javascript.

common/transport/amqp/src/sender_link.ts
YoDaMa: is there a problem with changing this line to `(context.sender as unknown).state`?

anthonyvercolano (approved):

YoDaMa:
when it's no longer a draft PR let me know.

common/transport/amqp/src/sender_link.ts
BertKleewein: I don't want to coerce this to `unkonwn`.  That would cause a type error.  

`any` is most permissive (meaning we can call any methods and access any properties on it).  `unkonwn` is least permissive (meaning we can't call any methods on it or access any properties on it).  `unknown` is used when we have an object, and we don't konw the type, and we don't care what the type is because we're just holding it or passing it through. 

[https://devblogs.microsoft.com/typescript/announcing-typescript-3-0-rc-2/#the-unknown-type](https://devblogs.microsoft.com/typescript/announcing-typescript-3-0-rc-2/#the-unknown-type)

YoDaMa (approved):
lgtm

PR #1037, opened 2021-09-24
feat(azure-iot-provisioning-service): support TokenCredential auth

vishnureddy17:


anthonyvercolano:
General comment, I see a number of unable to find module messages in imports and then also in the references to the RestApiClient namespace.  This was in code before your changes, but still, it would be nice to know why they are there and can we fix them?

service/test/_client_test.js
```diff
@@ -133,6 +133,7 @@ describe('Client', function () {
       assert.equal(client._transport._config.tokenCredential, fakeTokenCredential);
       assert.equal(client._restApiClient._config.host, 'hub.host.tv');
       assert.equal(client._restApiClient._config.tokenCredential, fakeTokenCredential);
+      assert.equal(client._restApiClient._config.tokenScope, 'https://azure-devices-provisioning.net/.default');
```
anthonyvercolano, line 136: Really want that address?

PR #1044, opened 2021-10-01
stress test to use fault injection with d2c

BertKleewein:


sdklab/stress_fault/src/stress_fault.ts
```diff
@@ -0,0 +1,259 @@
...
+    // Keep track of what faults are injected.
+    const statName = `faultCount-${fault.operationType}`;
+    this.stats[statName] = this.stats[statName] || 0;
+    this.stats[statName]++;
+    this.stats.totalFaultCount++;
+  }
+}
+
+// Top-level task for limiting the length of the test.  This task just sleeps
+// until the time elapses and then raises a sting to actually end the task.
```
anthonyvercolano, line 177: sting?  fancy dancy CS term.

sdklab/stress_fault/src/stress_fault.ts
```diff
@@ -0,0 +1,259 @@
...
+  objectName: string = 'TimeLimit';
+
+  constructor(interval: number) {
+    this._interval = interval;
+    this.stats = {
+      totalTime: utils.getIntervalString(interval),
+    };
+  }
+
+  async run(): Promise<void> {
```
anthonyvercolano, line 193: Two with the same name?

sdklab/stress_fault/src/stress_fault.ts
```diff
@@ -0,0 +1,259 @@
...
+
+  // Helper function to run an async function at a random interval.
+  private async _runAtRandomInterval(
+    func: Function,
+    minInterval: number,
+    maxInterval: number
+  ): Promise<void> {
+    while (!this.done) {
+      await func();
+      const interval = Math.floor(Math.random() * (maxInterval - minInterval + 1)) + minInterval;
```
anthonyvercolano, line 132: Is it possible to seed the random generator with a fixed constant so that we get the same random intervals on every run?

anthonyvercolano (approved):

sdklab/stress_fault/src/stress_fault.ts
BertKleewein: different class.

sdklab/stress_fault/src/stress_fault.ts
BertKleewein: The node Math.random generator is not seedable.  Other RNGs exist that are seedable, but I don't want to add that right now.  If we want repeatability later, we can add that.

PR #1051, opened 2021-10-11
fix(azure-iot-device): fix issues where token wouldn't refresh if stop() was called previously

vishnureddy17:


device/core/test/_sak_authentication_provider_test.js
```diff
@@ -244,6 +244,21 @@ describe('SharedAccessKeyAuthenticationProvider', function () {
...
+    it('clears _currentTokenExpiryTimeInSeconds so that _shouldRenewToken returns true', async function () {
+      const fakeCredentials = {
+        deviceId: 'fakeDeviceId',
+        host: 'fake.host.name',
+        sharedAccessKey: 'fakeKey'
+      };
+      const sakAuthProvider = new SharedAccessKeyAuthenticationProvider(fakeCredentials, 10, 1);
+      await sakAuthProvider.getDeviceCredentials();
+      assert.isFalse(isNaN(sakAuthProvider._currentTokenExpiryTimeInSeconds), 'isNaN() returned true when it should have been false');
+      assert.isFalse(sakAuthProvider._shouldRenewToken(), '_shouldRenewToken() returned true when it should have been true');
```
anthonyvercolano, line 257: should have been false?

PR #1055, opened 2021-10-19
Doc: Sample code cleanup

BertKleewein:
Fixed some comments.  Also, fixed places where the samples incorrectly assume that we always get an on("disconnect") event for every on("connect") event.


device/samples/javascript/device_through_proxy.js
```diff
@@ -53,18 +56,20 @@ function generateMessage () {
...
-function connectCallback () {
+function connectHandler () {
   console.log('Client connected');
   // Create a message and send it to the IoT Hub every two seconds
-  sendInterval = setInterval(() => {
-    const message = generateMessage();
-    console.log('Sending message: ' + message.getData());
-    client.sendEvent(message, printResultFor('send'));
-  }, 2000);
+  if (!!sendInterval) {
```
vishnureddy17, line 66: Shouldn't this be a single bang? Same feedback applies to the other samples.
```suggestion
  if (!sendInterval) {
```

device/samples/javascript/device_through_proxy.js
```diff
@@ -23,7 +23,10 @@ const deviceConnectionString = process.env.DEVICE_CONNECTION_STRING;
 let sendInterval;
 
 function disconnectHandler () {
-  clearInterval(sendInterval);
+  if (!!sendInterval) {
+    clearInterval(sendInterval);
+    sendInterval = null;
+  }
```
vishnureddy17, line 29: The double bang isn't needed here. Same feedback applies to the other samples.
```suggestion
  if (sendInterval) {
    clearInterval(sendInterval);
    sendInterval = null;
  }
```

device/samples/javascript/device_through_proxy.js
```diff
@@ -23,7 +23,10 @@ const deviceConnectionString = process.env.DEVICE_CONNECTION_STRING;
 let sendInterval;
 
 function disconnectHandler () {
-  clearInterval(sendInterval);
+  if (!!sendInterval) {
```
vishnureddy17, line 26: Based on my testing, `clearInterval(null)` and `clearInterval(undefined)` doesn't cause an error, so this `if` could potentially be removed entirely. Same feedback applies to the other samples.

device/samples/javascript/device_through_proxy.js
```diff
@@ -53,18 +56,20 @@ function generateMessage () {
...
-  }, 2000);
+  if (!!sendInterval) {
+    sendInterval = setInterval(() => {
+      const message = generateMessage();
+      console.log('Sending message: ' + message.getData());
+      client.sendEvent(message, printResultFor('send'));
+    }, 2000);
+  }
 
 }
```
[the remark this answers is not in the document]
vishnureddy17: Yeah I think all the `var` usage should be removed in favor of `let` and `const`

device/samples/javascript/simple_sample_device.js
```diff
@@ -47,26 +50,27 @@ function generateMessage () {
...
+function connectHandler () {
   console.log('Client connected');
   // Create a message and send it to the IoT Hub every two seconds
-  sendInterval = setInterval(() => {
-    const message = generateMessage();
-    console.log('Sending message: ' + message.getData());
-    client.sendEvent(message, printResultFor('send'));
-  }, 2000);
-
+  if (!!sendInterval) {
```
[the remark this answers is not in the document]
vishnureddy17: Yep

danhellem:
@BertKleewein @vishnureddy17 is this PR ready to be merged?

vishnureddy17:
> @BertKleewein @vishnureddy17 is this PR ready to be merged?

Not yet, there is still an error.

device/samples/javascript/device_through_proxy.js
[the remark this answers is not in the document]
BertKleewein: Good idea.  Out of scope for this change

device/samples/javascript/device_through_proxy.js
```diff
@@ -63,7 +61,7 @@ function errorHandler (err) {
 function connectHandler () {
   console.log('Client connected');
   // Create a message and send it to the IoT Hub every two seconds
-  if (!!sendInterval) {
+  if (!sendInterval) {
```
vishnureddy17, line 64: Not actionable, but this would be a good place to use the [logical nullish assignment operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_nullish_assignment) if/when we drop support for Node 14.

vishnureddy17 (approved):
nit: usage of `var` instead of `let` and `const`. It was there before but this could be a good time to change it.

Otherwise, LGTM

PR #1057, opened 2021-10-20
samples: Connection string rename

danhellem:
# Checklist

# Description of the problem
PNP samples and the Python repo samples use IOTHUB_DEVICE_CONNECTION_STRING instead of just DEVICE_CONNECTION_STRING. 

# Description of the solution
To make things consistent for customers who work across both SDKs, it makes sense to have the same env variable name. This PR renames it from DEVICE_CONNECTION_STRING to IOTHUB_DEVICE_CONNECTION_STRING


vishnureddy17 (approved):
LGTM, Might want to change the title of the PR so that commit from squashing and merging matches the conventional commits.

PR #1070, opened 2021-11-16
Adding commented protocols to appropriate samples

danhellem:
# Checklist

# Description of the problem
Getting customer feedback on how to use other protocols besides mqtt. Noticed that many of our samples do not have the block for customers to choose.

```
var Protocol = require('azure-iot-device-mqtt').Mqtt;
// const Protocol = require('azure-iot-device-amqp').AmqpWs;
// const Protocol = require('azure-iot-device-http').Http;
// const Protocol = require('azure-iot-device-amqp').Amqp;
// const Protocol = require('azure-iot-device-mqtt').MqttWs;
```

# Description of the solution
Added block to each sample.

### DIFF
--- a/device/samples/javascript/device_methods.js
+++ b/device/samples/javascript/device_methods.js
@@ -5,10 +5,10 @@
 
 // Choose a protocol by uncommenting one of these transports.
 var Protocol = require('azure-iot-device-mqtt').Mqtt;
-// const Protocol = require('azure-iot-device-amqp').AmqpWs;
-// const Protocol = require('azure-iot-device-http').Http;
-// const Protocol = require('azure-iot-device-amqp').Amqp;
-// const Protocol = require('azure-iot-device-mqtt').MqttWs;
+// var Protocol = require('azure-iot-device-amqp').Amqp;
+// var Protocol = require('azure-iot-device-http').Http;
+// var Protocol = require('azure-iot-device-mqtt').MqttWs;
+// var Protocol = require('azure-iot-device-amqp').AmqpWs;
 
 require('es5-shim');
 var Client = require('azure-iot-device').Client;
--- a/device/samples/javascript/device_methods.js
+++ b/device/samples/javascript/device_methods.js
@@ -3,10 +3,15 @@
 
 'use strict';
 
-require('es5-shim');
+// Choose a protocol by uncommenting one of these transports.
 var Protocol = require('azure-iot-device-mqtt').Mqtt;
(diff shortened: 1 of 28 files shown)


device/samples/javascript/device_methods.js
```diff
@@ -3,10 +3,15 @@
 
 'use strict';
 
-require('es5-shim');
+// Choose a protocol by uncommenting one of these transports.
 var Protocol = require('azure-iot-device-mqtt').Mqtt;
```
vishnureddy17, line 7: nit: `var` is used for the first one and `const` is used for the others. We should use `var` or `const` but not both.
`let`/`const` is preferable but a lot of the sample code uses `var` so changing all of the `var`s might not be in scope for this PR.

device/samples/javascript/dmpatterns_reboot_device.js
```diff
@@ -2,9 +2,14 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 'use strict';
 
-var Client = require('azure-iot-device').Client;
+// Choose a protocol by uncommenting one of these transports.
```
vishnureddy17, line 5: Do we want to leave all of them commented so we force the customer to make a choice? I'm not sure about this myself but something to consider.

device/samples/javascript/upload_to_blob.js
```diff
@@ -1,7 +1,8 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-// This example is deprecated. We recommend you follow the 'upload_to_blob_advanced.js' sample.
+// ⚠️THIS EXAMPLE IS DEPRECATED⚠️
```
vishnureddy17, line 4: nit: IMO non-ascii characters shouldn't be in code

vishnureddy17 (approved):

device/samples/javascript/dmpatterns_reboot_device.js
danhellem: i think we should default to having mqtt enabled. If the user wants to try another choice they can comment and uncomment. I would like to push people to use mqtt

device/samples/javascript/device_methods.js
danhellem: I agree. But that would require more clean up and testing. lets make it `var` for now and I will go ahead and do a separate PR later using `const` / `let` instead of `var`.

PR #1071, opened 2021-11-17
test(stresstests): initial stress D2C stress tests without faults

vishnureddy17:
Apologies for the large PR. Should have broken this up into multiple reviews in hindsight.

These tests are flaky, so either the SDK needs some work or the tests need some tweaking.


stresstests/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "azure-iot-stresstests",
+  "description": "Azure IoT Stress Tests",
+  "author": "Microsoft Corporation",
+  "version": "1.0.0",
+  "private": true,
+  "license": "MIT",
+  "//": "TODO: will the bump version pipeline update the dependencies?",
```
BertKleewein, line 8: I was just going to add a comment. This directory needs to be added to internals/release/node-sdk/inputs.js

stresstests/test/config.js
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+
+const deviceHttp = require('azure-iot-device-http');
+const deviceAmqp = require('azure-iot-device-amqp');
+const deviceMqtt = require('azure-iot-device-mqtt'); 
+
+module.exports = {
```
BertKleewein, line 10: >exports [](http://example.com/codeflow?start=7&length=7)

Curious why you used JS intead of TS.  Is this because of mocha?  Is this still the right test framework to use?

vishnureddy17: Mainly because I wasn't sure how TS interacts with Mocha. Sure, I could have figured it out but I didn't think it was worth the effort.

vishnureddy17: Decided to stick with mocha because it seems to work for our scenario and we can avoid our repo having dependencies on different test runners

stresstests/test/test_d2c_stress.js
```diff
@@ -0,0 +1,185 @@
...
+      /* 7 is the number of characters in {"":""} */
+      const payloadSize = Math.max(config.telemetryPayloadSizeInBytes, 7);
+      const keyLength = Math.ceil((payloadSize - 7) / 2);
+      const valLength = payloadSize - keyLength - 7;
+      const message = new Message(JSON.stringify({[randomString(keyLength)]: randomString(valLength)}));
+      message.messageId = messageId;
+      const deferred = eventHubHelper.awaitMessage(messageId);
+      stressMeasurementsRecorder.messageEnqueued();
+      let attempts = 0;
+      while (attempts < 5) {
```
vishnureddy17, line 93: Doing retries here. The SDK is supposed to retry for us, at least for certain errors that are deemed retriable. I added this to reduce the flakiness of the tests. Should we keep this, or are we suppressing failures that are interesting to us by doing this?

BertKleewein: This depends on the API contract.  Right now, the Python contract doesn't have reliable retry, so the longhaul tests need to do it themselves.  Node, on the other hand, does have reliable retry, so this is unnecessary and probably hides failures that we'd want to see.

BertKleewein:
> 		"": {

If you do `lerna bootstrap --hoist`, the delta isn't nearly this big.  Regardless, the way we use this file seems silly -- our updating is ad-hoc and we don't actually use it for it's indented purpose.  What do you think we should do about these files?   What is the industry standards at this point?

---
Refers to: package-lock.json:6 in 291314d. [](commit_id = 291314dd2ef971a26ca562a1cb75b65ca869df32, deletion_comment = False)

stresstests/test/device_client_helper.js
```diff
@@ -0,0 +1,256 @@
...
+      client.close().then(() => debug('client closed')),
+      this._registry.delete(deviceId).then(() => debug('device id removed from registry'))
+    ]);
+  }
+
+  /**
+   * Registers a device with the IoT Hub and then creates a corresponding device
+   * client (Client) that authenticates using a shared access signature.
+   * @param {function} transportCtor - The constructor of the device transport to use.
+   * @param {string} [modelId] - An optional model ID to use for the device.
```
BertKleewein, line 59: >optional model ID to [](http://example.com/codeflow?start=36&length=20)

PNP model id, right?  Maybe use a more specific name for the parameter

stresstests/test/device_client_helper.js
```diff
@@ -0,0 +1,256 @@
...
+    }
+  }
+
+  /**
+   * Sets the token valid time in seconds and token renewal margin in seconds
+   * for the device client. Only supported for clients with symmetric key
+   * authentication.
+   * 
+   * NOTE: This method relies on internal implementation details of the client
+   * and is not guaranteed to work in the future.
```
BertKleewein, line 180: There are public APIs for this.  Maybe add  comment on why you did it this way.

stresstests/test/event_hub_helper.js
```diff
@@ -0,0 +1,153 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+'use strict';
+
+const EventHubClient = require('@azure/event-hubs').EventHubClient;
+const EventPosition = require('@azure/event-hubs').EventPosition;
+const debug = require('debug')('stresstests:EventHubHelper');
+
+/**
+ * A thenable that can be can be resolved or rejected from outside and records
```
BertKleewein, line 11: >thenable [](http://example.com/codeflow?start=5&length=8)

Nice.  Is this a word?

stresstests/test/event_hub_helper.js
```diff
@@ -0,0 +1,153 @@
...
+ * the time it takes to settle.
+ */
+class Deferred {
+  constructor() {
+    this.startTime = Date.now();
+    this.timeToSettle = null;
+    this.promise = new Promise((resolve, reject) => {
```
BertKleewein, line 18: >promise [](http://example.com/codeflow?start=9&length=7)

Should `this.promise` be an instance variable or a closure (hidden) variable?  If we don't want user's touching it, we should either underscore it or hide it.

stresstests/test/stress_measurements_recorder.js
```diff
@@ -0,0 +1,143 @@
...
+    this.peakReconnectTimeInMs = null;
+    this.totalElapsedTime = null;
+    this._startTime = Date.now();
+    this._disconnectedTime = null;
+    this._memorySamplingInterval = setInterval(() => {
+      const rss = process.memoryUsage().rss;
+      this.peakRssInBytes = Math.max(this.peakRssInBytes, rss);
+      debug(`RSS: ${rss / 1024 / 1024} MiB (max ${this.peakRssInBytes / 1024 / 1024} MiB)`);
+    }, memorySamplingIntervalInMs);
+    this._client.on('connect', this._connectHandler);
```
BertKleewein, line 102: >_connectHandler [](http://example.com/codeflow?start=36&length=15)

dumb question: why do you have to bind when assigning to `Deferred.then` but not here?

stresstests/test/stress_measurements_recorder.js
```diff
@@ -0,0 +1,143 @@
...
+      this._disconnectedTime = null;
+    }
+  }
+
+  /**
+   * @private
+   */
+  _disconnectHandler = () => {
+    debug('client emitted a disconnect event')
+    this._disconnectedTime = Date.now()
```
BertKleewein, line 139: >_disconnectedTime [](http://example.com/codeflow?start=9&length=17)

What if `disconnected` gets fired twice before `connected` gets fired?

BertKleewein (approved):
:shipit:

stresstests/test/test_d2c_stress.js
vishnureddy17: Cool. I'll remove the retrying then.

stresstests/test/event_hub_helper.js
vishnureddy17: It's a word in the JavaScript world: https://masteringjs.io/tutorials/fundamentals/thenable

stresstests/test/stress_measurements_recorder.js
vishnureddy17: Bind doesn't work for arrow functions. In arrow functions, `this` gets captured from the scope that the arrow function is defined in. Though the way I've done this is wrong, because `this` in `_connectHandler` and `_disconnectHandler` will be the global object.

stresstests/test/device_client_helper.js
vishnureddy17: Actually, I'll just remove this because we never shipped the native PnP implementation

vishnureddy17:
> > ```
> >   "": {
> > ```
> 
> If you do `lerna bootstrap --hoist`, the delta isn't nearly this big. Regardless, the way we use this file seems silly -- our updating is ad-hoc and we don't actually use it for it's indented purpose. What do you think we should do about these files? What is the industry standards at this point?
> 
> Refers to: package-lock.json:6 in [291314d](https://github.com/Azure/azure-iot-sdk-node/commit/291314dd2ef971a26ca562a1cb75b65ca869df32). [](commit_id = [291314d](https://github.com/Azure/azure-iot-sdk-node/commit/291314dd2ef971a26ca562a1cb75b65ca869df32), deletion_comment = False)

@BertKleewein Not really sure what the standard is TBH, but I've just been committing the `package-lock.json`s any time I change a dependency and omit them otherwise.

PR #1072, opened 2021-11-18
build: emit es2017 commonjs modules

BertKleewein:


device/core/test/_device_client_test.js
```diff
@@ -40,11 +40,11 @@ describe('Device Client', function () {
       var clock = this.clock;
 
       var x509ConnectionString = 'HostName=host;DeviceId=id;x509=true';
-      var client = new Client.fromConnectionString(x509ConnectionString, FakeTransport);
+      var client = Client.fromConnectionString(x509ConnectionString, FakeTransport);
       assert.instanceOf(client, Client);
 
       sinon.stub(client._transport, 'updateSharedAccessSignature').callsFake(function () {
-        clock.restore();
+        clock.restoee();
```
vishnureddy17, line 47: typo?
```suggestion
        clock.restore();
```

common/core/package.json
```diff
@@ -6,6 +6,7 @@
   "license": "MIT",
   "main": "common.js",
   "typings": "common.d.ts",
+  "type": "commonjs",
```
vishnureddy17, line 9: Is this needed?
```suggestion
```

security/tpm/package.json
```diff
@@ -55,10 +55,10 @@
       ".tsx"
     ],
     "check-coverage": true,
-    "lines": 70,
+    "lines": 68,
     "functions": 77,
-    "branches": 55,
-    "statements": 71
+    "branches": 54,
+    "statements": 69
```
vishnureddy17, line 61: Interesting that changing the target reduced our test coverage

sdklab/stress_fault/tsconfig.json
```diff
@@ -1,12 +1,10 @@
...
     "compilerOptions": {
         "outDir": "./dist",
-        "target":"es5",
+        "target":"es2017",
+        "module": "commonjs",
         "sourceMap": true,
         "declaration": true,
-        "lib": [
-            "es6"
-        ],
```
vishnureddy17, line 9: We also have the `lib` field in `service/samples/typescript`, `service`, and `common/transport/mqtt`. Could we get rid of those too?

vishnureddy17 (changes requested):
I think the `clock.restoee()` thing will cause an issue.

security/tpm/package.json
BertKleewein: Probably because there's more (or different) boilerplate code in the generated module that deals with downlevel versions.

vishnureddy17 (approved):

PR #1079, opened 2021-12-01
device-twin-service helper

danhellem:
# Checklist
- [x] I have read the [contribution guidelines](https://github.com/Azure/azure-iot-sdk-node/blob/main/.github/CONTRIBUTING.md).

# Description of the problem
device_twin samples are confusing and not accurate. They require a separate service to post patch updates for desired properties.

# Description of the solution
- added a new helper project that can be used as a backend service
- updated device_twin samples
- added readme's


vishnureddy17:
nit: IMO it'd be better to use async/await rather than callbacks but it might not be worth going back and redoing it.

vishnureddy17 (approved):

device/samples/helpers/device-twin-service/twin-service.js
```diff
@@ -0,0 +1,96 @@
+'use strict';
+
+const iothub = require('azure-iothub');
+const connectionString = process.env.IOTHUB_CONNECTION_STRING;
+const deviceId = 'devDevice';
+
+if (connectionString === undefined) {
+  console.log('IoT Hub connection string not set');
+  process.exit(-1);
```
vishnureddy17, line 9: Can you explain what a negative exit code does? I haven't seen this before.

danhellem: I don't think it does anything special :)

i changed it to zero

danhellem:
> nit: IMO it'd be better to use async/await rather than callbacks but it might not be worth going back and redoing it.

If that is the better pattern then we should most def fix them. But let's do that in a separate PR

device/samples/helpers/device-twin-service/twin-service.js
vishnureddy17: A 0 exit code means success so I think it should be non-zero

device/samples/helpers/device-twin-service/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "iot-hub",
```
BertKleewein, line 2: >"name": "iot-hub", [](http://example.com/codeflow?start=2&length=18)

You should give this a better name and add it to lerna.json and the inputs.js file in the internals repo.  lerna.json ensures that the dependencies get linked for local development, and input.js ensures that the dependency versions get updated.

BertKleewein (approved):
:shipit:

device/samples/helpers/device-twin-service/package.json
danhellem: @BertKleewein I updated the name and added to lerna.json. I can't find an inputs.js file

PR #1081, opened 2021-12-02
fix typo in TemperatureController telemetry

rido-min:
workingset->workingSet

### DIFF
--- a/device/samples/javascript/pnp_temperature_controller.js
+++ b/device/samples/javascript/pnp_temperature_controller.js
@@ -290,7 +290,7 @@ async function main() {
 
 
     intervalToken3 = setInterval(() => {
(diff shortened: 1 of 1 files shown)


danhellem:
@BertKleewein  and @vishnureddy17 this is a PR to check against some tests. Per @rido-min 

> IIRC we were asking all SDK samples to run the certification tests , which should have detect this issue.

BertKleewein (approved):
:shipit:

danhellem:
@rido-min what do you want us to do with this?

rido-min:
merge it? not sure why the pipeline fails since the change is fairly simple.

BertKleewein:
The gates have been unreliable.  I would be OK bypassing the gate for this change.

danhellem:
Minor typo fix but the build keeps failing. I am going to just use my admin privileges to merge.

PR #1085, opened 2021-12-28
Users/danhellem/device samples

danhellem:
# DO NOT PUBLISH THIS, PR ONLY FOR FEEDBACK

# Checklist
- [x] I have read the [contribution guidelines](https://github.com/Azure/azure-iot-sdk-node/blob/main/.github/CONTRIBUTING.md).

These are for our new device samples format and content


doc/device-samples/iot-hub-prerequisites.md
```diff
@@ -0,0 +1,47 @@
+# 🔨 Create an IoT hub
```
vishnureddy17, line 1: Can we just link to some existing documentation on creating an IoT Hub? IMO we shouldn't have multiple places documenting the same thing. There might also be other stuff in the `doc` directory that can also just be an MSDocs link.

device-samples/getting started/README.md
```diff
@@ -0,0 +1,38 @@
...
+In some scenarios you can't easily map the data your devices send into the relatively small device-to-cloud messages that IoT Hub accepts. For example: Large files that contain images or videos. This sample will show you how to upload file to a linked Azure storage account.
+
+# Next Steps
+
+- [Send messages to IoT Hub](./src/send%20messages)
+- [How to guides](src/../../how%20to%20guides)
+
+# Feedback
+
+If you have any feedback or questions about our device samples, please [post it here](https://github.com/Azure/azure-iot-sdk-node/discussions/1042).
```
vishnureddy17, line 38: We need to make sure to keep tabs on the discussion section of the repo

vishnureddy17 (commented):
This looks great! Some nits:
- could the spaces be removed from the directory names (e.g. `getting started` -> `getting_started`)
- In directories with multiple examples (e.g., send message, device twins), I think it'd better to put JS and TS examples in their own subdirectories
- Do we want to have users pull the package from NPM or link the local source code with Lerna? Right now there is no Lerna linking, and `dev-environment.md` just has the user pull from NPM. Maybe we want to keep it that way, but it's something to consider. 
- Minor spelling/grammar issue: In a few places, "setup" should be "set up" and "environmental variable" should be "environment variable"

device-samples/getting started/README.md
```diff
@@ -0,0 +1,38 @@
+# Getting started samples (level 100)
+
+These getting started samples are perfect for someone just getting into the [Azure IoT Device SDK for Node.js](https://github.com/Azure/azure-iot-sdk-node). The samples are targeted, simple, and you should be able to get them up and running in just a couple minutes. For more complicated samples and scenarios, see the [How to guides](../how%20to%20guides) section.
+
+🔔 If you are just starting with Azure IoT device SDK's, we recommend you start with the basic [send messages to IoT Hub](src/send%20messages) example.
+
+## [Connection types](src/connections)
+
+These samples will show you the different connections types you can use to create a client. To keep things simple, all our samples showcase the **connection string** method. However, you can use other connection methods like Shared Access Signature (SAS), X509 Certificates, and web socket proxies. These samples will show you how to create those clients and send a single event to Azure IoT Hub.
```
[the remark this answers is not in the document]
danhellem: Fixed

doc/device-samples/iot-hub-prerequisites.md
danhellem: the existing IoT Hub docs are long and confusing. Thus the reason I wanted to create a shortened and targeted version of them. Something that gets to the point and is easy to follow. We can discuss.

device-samples/getting started/README.md
danhellem: Agreed

danhellem:
> This looks great! Some nits:
> 
> * could the spaces be removed from the directory names (e.g. `getting started` -> `getting_started`)

Done

> * In directories with multiple examples (e.g., send message, device twins), I think it'd better to put JS and TS examples in their own subdirectories

Lets talk about this. I like having the TS files side by side with JS files.

> * Do we want to have users pull the package from NPM or link the local source code with Lerna? Right now there is no Lerna linking, and `dev-environment.md` just has the user pull from NPM. Maybe we want to keep it that way, but it's something to consider.

I think the user should have to pull from NPM. Just like the would in the real world.

> * Minor spelling/grammar issue: In a few places, "setup" should be "set up" and "environmental variable" should be "environment variable"

Fixed, i think :)

BertKleewein:
I love this! We really need this kind of organization and I'm glad that you're taking this on.

> Lets talk about this. I like having the TS files side by side with JS files.

I was about to make a similar comment. I tend to agree with Vishnu, especially since we have hand-written JS, and we have compiler-generated JS, and the compiler-generated JS is more closely related to the TS, but it's further away in the file system.  

Plus the `src` directory naming convention is more closely aligned with TS. though I've also seen it used with JS, especially when using webpack-type tools  

Also, I don't know the answer, but do we want `device_samples/getting_started/src/connections` or do want `device_samples/getting_started/connections/src`? The latter makes more sense to me, but you've though about it more than I have.

---
In reply to: [1005128467](https://github.com/Azure/azure-iot-sdk-node/pull/1085#issuecomment-1005128467) [](http://example.com/codeflow?ancestors=1005128467,1005128467)

device_samples/getting_started/src/connections/connection_string.js
```diff
@@ -0,0 +1,67 @@
...
+const Client = require('azure-iot-device').Client;
+const Message = require('azure-iot-device').Message;
+
+// String containing Hostname, Device Id & Device Key in the following formats:
+//  "HostName=<iothub_host_name>;DeviceId=<device_id>;SharedAccessKey=<device_key>"
+const deviceConnectionString = process.env.IOTHUB_DEVICE_CONNECTION_STRING;
+
+// make sure we have a connection string before we can continue
+if (deviceConnectionString === '' || deviceConnectionString === undefined) {
+  console.error('\x1b[31m%s\x1b[0m', 'Missing device connection string');
```
BertKleewein, line 22: >'\x1b[31m%s\x1b[0m [](http://example.com/codeflow?start=16&length=18)

nit: I _love_ colored output for this kind of things, but shouldn't the `console.error` implementation be adding this?  You're assuming an ANSI console, and that's OK, but I've had to script these characters away before so I see this as a point of annoyance.

device_samples/getting_started/src/connections/connection_string.js
```diff
@@ -0,0 +1,67 @@
...
+// open client connection
+client.open().catch((err) => {
+  console.error('\x1b[31m%s\x1b[0m', 'Could not connect: ' + err.message);
+  process.exit(0);
+});
+console.log('Client connection: Open');
+
+// send a messge
+console.log('Sending message: ' + message.getData());
+client.sendEvent(message, printResultFor('Send'));
```
BertKleewein, line 48: If I'm reading this right, `client.sendEvent` returns a `Promise` object, but this sample never waits for this Promise to complete.  And looking above, shouldn't there be a `.then` after the `client.open().catch()`? I think the `console.log` on line 44 _could_ execute before the client is actually open.  There's so many ways to write in JS now that I'm not even sure anymore.

BertKleewein: Sorry, I'm wrong, it doesn't return a `Promise` because we're passing in a callback via `printResultsFor`. But, the concern stands that code after this _could_ execute before the message is actually sent.

I personally think this `printResultsFor` is an artifact of a coding shortcut written in ancient times based on the limitations of the day.  This pattern should be stricken from our samples, and we should decide if we want our JS to use:
1. callbacks
2. .Promise chains (then.then.then.catch)
or
3. await

It seems like we're living with a combination of all 3, and I'm guilty of using all 3 together, so I'm not sure if there's a better option, but we should talk about it.

device_samples/getting_started/src/connections/connection_string.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+function generateMessage() {
+  const data = JSON.stringify({
+    deviceId: 'my-first-device',
+    temperature: 20 + Math.random() * 10,
+  });
+  
+  return new Message(data);
+}
```
BertKleewein, line 33: nit: this function is copy/pasted all over our samples.  We could probably delete hundreds of lines if we just hardcode messages data.

device_samples/getting_started/src/send_messages/send_messages_in_batch_http.ts
```diff
@@ -0,0 +1,40 @@
...
+  { id: 1, message: 'message one' },
+  { id: 2, message: 'message two' },
+  { id: 2, message: 'message three' }
+];
+
+const messages: any[] = [];
+
+data.forEach(function (value: { id: number; message: string }): void {
+  messages.push(new Message(JSON.stringify(value)));
+});
```
BertKleewein, line 30: nit `id` is unused, plus id==2 is used twice, and the comment says "two messages" but there's actually 3.

having 2 arrays isn't necessary.  This could be shortened to skip the `data` array altogether as: 
```
const messages: Message[] = [
  new Message(JSON.stringify("message one"),
  new Message(JSON.stringify("message two"),
  new Message(JSON.stringify("message three"),
]
```

device_samples/getting_started/src/send_messages/send_messages_to_central.js
```diff
@@ -0,0 +1,127 @@
+// Copyright (c) Microsoft. All rights reserved.
```
BertKleewein, line 1: nit: do you want to call this `send_message_to_iot_central.js`? also `send_mesage_to_iot_hub.js`

device_samples/getting_started/src/send_messages/send_messages_to_central.js
```diff
@@ -0,0 +1,127 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
BertKleewein, line 2: >Licensed [](http://example.com/codeflow?start=3&length=8)

I LOVE, love, love how you organized the `send_messages` sample here.  "How can I send a message to iot central?" feels like a very common question, and it's way better to have a single level-100 sample that shows this together instead of forcing the reader to take part of a DPS sample and part of a sendEvent sample and munge them together.

Related side question: should there be samples under `connections` for connecting to central and for connecting using DPS and using the Edge environment?  I know that DPS is a different library, but it's really just a different way to connect your device.

device_samples/getting_started/src/send_messages/send_messages_to_central.js
```diff
@@ -0,0 +1,127 @@
...
+
+if (deviceId === '' || deviceId === undefined) {
+  console.error('\x1b[31m%s\x1b[0m', 'Missing Device Id string');
+  process.exit(0);
+}
+
+if (deviceKey === '' || deviceKey === undefined) {
+  console.error('\x1b[31m%s\x1b[0m', 'Missing Device Key string');
+  process.exit(0);
+}
```
BertKleewein, line 44: nit: this is 20 lines of code that doesn't add any value to the samples.  It could be replaced with 4 asserts

device/samples/javascript/simple_sample_device_twin.js
```diff
@@ -198,6 +198,7 @@ client.open(function(err) {
               }
             }
           });
+          console.log();
```
anthonyvercolano, line 201: Not sure about this.

device_samples/getting_started/src/connections/connection_string.js
```diff
@@ -0,0 +1,67 @@
...
+  console.error('\x1b[31m%s\x1b[0m', 'Could not connect: ' + err.message);
+  process.exit(0);
+});
+console.log('Client connection: Open');
+
+// send a messge
+console.log('Sending message: ' + message.getData());
+client.sendEvent(message, printResultFor('Send'));
+
+client.removeAllListeners();
```
anthonyvercolano, line 50: Since this example didn't add any listeners, we should probably not go through the exercise of removing them.
Bert you are correct about line 44.

device_samples/getting_started/src/connections/connection_string.ts
```diff
@@ -0,0 +1,62 @@
...
+client.open().catch((err: Error) => {
+  console.error('\x1b[31m%s\x1b[0m', 'Could not connect: ' + err.message);
+  process.exit(0);
+});
+console.log('Client connection: Open');
+
+// send a messge
+console.log('Sending message: ' + message.getData());
+client.sendEvent(message, printResultFor('Send'));
+client.removeAllListeners();
```
anthonyvercolano, line 45: Pretty dangerous.

device_samples/getting_started/src/connections/connection_string.js
vishnureddy17: Agreed, and I'd vote for async/await-ifying wherever reasonable.

device_samples/getting_started/src/send_messages/send_messages_to_central.js
vishnureddy17: +1 on samples under `connections` for Central, DPS, and Edge.

device_samples/getting_started/tslint.json
```diff
@@ -0,0 +1,13 @@
+{
```
vishnureddy17, line 1: We want to move everything over to ESLint at some point, so maybe we should use ESLint here. But also it might be better to keep everything consistent and move everything over to ESLint in one go.

device_samples/getting_started/src/connections/connection_string.ts
```diff
@@ -0,0 +1,62 @@
...
+// import { Amqp as Protocol } from 'azure-iot-device-amqp';
+// import { Http as Protocol } from 'azure-iot-device-Http';
+// import { MqttWs as Protocol } from 'azure-iot-device-mqtt';
+// import { AmqpWs as Protocol } from 'azure-iot-device-amqp';
+
+import { Client, Message } from 'azure-iot-device';
+
+// String containing Hostname, Device Id & Device Key in the following formats:
+//  "HostName=<iothub_host_name>;DeviceId=<device_id>;SharedAccessKey=<device_key>"
+const deviceConnectionString: string = process.env.IOTHUB_DEVICE_CONNECTION_STRING || '';
```
vishnureddy17, line 15: nit: we could use the nullish coalescing operator (??) if we are using a reasonable new TypeScript compiler
```suggestion
const deviceConnectionString: string = process.env.IOTHUB_DEVICE_CONNECTION_STRING ?? '';
```

vishnureddy17: Also, in some places it's `process.env.IOTHUB_DEVICE_CONNECTION_STRING || ''` and other places it's just `process.env.IOTHUB_DEVICE_CONNECTION_STRING`. Better to be consistent

device_samples/getting_started/src/connections/proxy.js
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) Microsoft. All rights reserved.
```
vishnureddy17, line 1: I don't see a TypeScript version of this sample

device_samples/getting_started/src/device_twins/service/Readme.md
```diff
@@ -0,0 +1,59 @@
+# Device Twin Backend Service
```
vishnureddy17, line 1: nit: filename not in caps

device_samples/getting_started/src/device_twins/service/reported_properties.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
```
vishnureddy17, line 1: no typescript for the samples in `service`

device_samples/getting_started/src/receive_messages/receive_message.js
```diff
@@ -0,0 +1,48 @@
...
+        console.log('Ready to recieve messages...');
+    }
+}
+
+function handleMessage(msg) {
+    console.log('\x1b[33m%s\x1b[0m', `\nReceived C2D message:`);
+    console.log(` message: ${msg.data.toString().trim()}`);
+    console.log(` properties: ${JSON.stringify(msg.properties.propertyList)}`);   
+}
+
```
vishnureddy17, line 46: One issue here is that for Amqp and Http the user is supposed to "complete" the C2D message and for Mqtt is not.
Maybe check the transport constructor name and complete the message if appropriate?

device_samples/getting_started/src/send_messages/README.md
```diff
@@ -0,0 +1,68 @@
+# Send messages
+
+Sending telemetry messages is the first and most basic thing to do using the SDK. There are multiple examples below and each example covers multipe protocols. We suggest you start with the [send messages to IoT Hub](#send-messages-to-hub) to ensure you successfull connection and message send using MQTT. Then you can work work your way through the different protocols.
```
vishnureddy17, line 3:
```suggestion
Sending telemetry messages is the first and most basic thing to do using the SDK. There are multiple examples below and each example covers multiple protocols. We suggest you start with the [send messages to IoT Hub](#send-messages-to-hub) to ensure you successful connection and message send using MQTT. Then you can work your way through the different protocols.
```

device_samples/getting_started/src/send_messages/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+Also note, that all of these examples use the [connection string](../connections/connection_string.js) method to create a connection client. See the [connections](../connections) section if you want use a different connection method.
+
+# 🦉 Getting setup
```
vishnureddy17, line 7:
```suggestion
# 🦉 Getting set up
```

device_samples/getting_started/src/send_messages/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+> For a more in-depth example of plug and play capabilies, see [this how to sample](../../../how_to_guides/src/plug_and_play)
+
+#### Running the sample
+
+From the `getting_started/src/send_messages` directory, run `node send_messages_to_central.js`
+
+### Send messages in batch
+
+Batch up several messages and send them in a single HTTP request. |
```
vishnureddy17, line 55:
```suggestion
Batch up several messages and send them in a single HTTP request.
```

device_samples/getting_started/README.md
```diff
@@ -0,0 +1,38 @@
+# Getting started samples (level 100)
+
+These getting started samples are perfect for someone just getting into the [Azure IoT Device SDK for Node.js](https://github.com/Azure/azure-iot-sdk-node). The samples are targeted, simple, and you should be able to get them up and running in just a couple minutes. For more complicated samples and scenarios, see the [How to guides](../how_to_guides) section.
```
vishnureddy17, line 3: I think we should emphasize that these level 100 samples are not intended to be the basis of a production application

vishnureddy17:
> Also, I don't know the answer, but do we want `device_samples/getting_started/src/connections` or do want `device_samples/getting_started/connections/src`? The latter makes more sense to me, but you've though about it more than I have.
> 
> In reply to: [1005128467](https://github.com/Azure/azure-iot-sdk-node/pull/1085#issuecomment-1005128467)\

+1

device_samples/getting_started/src/connections/proxy.js
danhellem: yeah the TS version still needs to be created. We had trouble getting it working a few months ago remember?

YoDaMa:
reviewing this... I agree with the comments already made. I'll see what else can be said. Once this is merged into a branch on the main repo I can also contribute some changes.

device_samples/getting_started/tslint.json
danhellem: I will leave this to you if you want to move to ESList later

device_samples/getting_started/src/send_messages/send_messages_to_central.js
danhellem: @BertKleewein do you have an example on how we could do that?

device_samples/getting_started/src/connections/connection_string.js
danhellem: I fixed this in the connections samples, but it looks like it still adds value in the send messages samples. I guess we could simplify the telemetry we send. that would make it easier to remove that function and just hard code.

@BertKleewein  thought?

danhellem:
> > Also, I don't know the answer, but do we want `device_samples/getting_started/src/connections` or do want `device_samples/getting_started/connections/src`? The latter makes more sense to me, but you've though about it more than I have.
> > In reply to: [1005128467](https://github.com/Azure/azure-iot-sdk-node/pull/1085#issuecomment-1005128467)
> 
> +1

I fixed this in the branch I am going to push in a bit

PR #1143, opened 2022-07-28
Remove sinon types from package.json

BertKleewein:


vishnureddy17:
Are the typings shipped with sinon now?

anthonyvercolano (approved):

BertKleewein:
@vishnureddy17  - No, but all of our tests are in javascript so we don't need them.

PR #1145, opened 2022-08-28
add event listener tracker for mqtt base client

vishnureddy17:


common/transport/mqtt/src/external_event_tracker.ts
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+import { EventEmitter } from 'events';
```
anthonyvercolano, line 4: You can leave it as is.  I probably would have put more comments in describing the structure in words the the scenarios using each method.

anthonyvercolano (approved):

PR #1149, opened 2022-09-21
Improve handling of the client closed error.

anthonyvercolano:
Better systematic handling of enabling disabling features.
Better handling of the client closing error on a subscribe/unsubscribe.


device/transport/mqtt/src/mqtt.ts
```diff
@@ -1020,6 +1020,16 @@ export class Mqtt extends EventEmitter implements DeviceTransport {
       });
     }
   }
+
+  private _ignoreConnectionClosedInErrorCallback(callback: (err?: Error, ...args: any[]) => void): (err?: Error, ...args: any[]) => void {
```
BertKleewein, line 1024: >_ignoreConnectionClosedInErrorCallback [](http://example.com/codeflow?start=10&length=38)

This needs to be explained in both places, otherwise it will become a Chesterson’s Fence. (one of my new favorite obscure terms).

service/package.json
```diff
@@ -10,7 +10,7 @@
     "async": "^3.2.3",
     "es5-ext": "0.10.53",
     "@azure/core-http": "1.2.3",
-    "@azure/identity": "1.2.5",
+    "@azure/identity": "2.0.0",
```
BertKleewein, line 13: > "@azure/identity": "2.0.0", [](http://example.com/codeflow?start=3&length=28)

Intentional change?

device/core/src/device_client.ts
```diff
@@ -96,7 +96,7 @@ export class Client extends InternalClient {
       if (err && this._retryPolicy.shouldRetry(err)) {
         debugErrors('reconnect policy specifies a reconnect on error');
         /*Codes_SRS_NODE_DEVICE_CLIENT_16_097: [If the transport emits a `disconnect` event while the client is subscribed to c2d messages the retry policy shall be used to reconnect and re-enable the feature using the transport `enableC2D` method.]*/
-        if (this._c2dFeature) {
+        if (this._userRegisteredC2dListener) {
```
BertKleewein, line 99: >_userRegisteredC2dListener [](http://example.com/codeflow?start=17&length=26)

<no action required> It hurts my soul to see `this._enableXXX` code inside the `_deviceDisconnectHandler` function

BertKleewein (approved):
:shipit:

device/core/src/device_client.ts
anthonyvercolano: I'm going to leav it.  I'm assuming it hurts your soul because it should have resided in some renablefeature class?

device/core/src/module_client.ts
```diff
@@ -53,24 +53,33 @@ export class ModuleClient extends InternalClient {
...
+          } else {
+            debug('removeListener successfully disabled input messages.');
           }
         });
       }
     });
 
     this.on('newListener', (eventName) => {
       if (eventName === 'inputMessage') {
+        this._userRegisteredInputMessageListener = true;
```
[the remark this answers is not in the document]
anthonyvercolano: Done.

device/core/src/device_client.ts
BertKleewein: I have no problem leaving it as-is.

It hurts my soul because it looks like this code is trying to enable C2D as part of the disconnect process. This is odd. Plus, it's not obvious that `this._enableC2D` is reconnecting the transport.

In other words, enabling C2D is a side-effect of being disconnected, and reconnecting is a side-effect of Enabling C2D. 

If C2D is not enabled, then the reconnect happens in a completely different place.  I hope.

PR #1151, opened 2022-10-12
Fix 1141 - support dt-subject for https

anthonyvercolano:


common/transport/amqp/src/amqp_message.ts
```diff
@@ -120,12 +120,23 @@ export class AmqpMessage {
...
           ensureApplicationPropertiesCreated();
         }
+        const DT_SUBJECT = 'dt-subject';
         for (let index = 0; index < propsCount; index++) {
           const item = props.getItem(index);
           if (!!item) {
-            /*Codes_SRS_NODE_IOTHUB_AMQPMSG_16_013: [If one of the property key is `IoThub-status`, this property is reserved and shall be forced to an `int` `rhea` type.]*/
-            const val = (item.key === 'IoThub-status') ? rheaTypes.wrap_int(parseInt(item.value)) : item.value;
-            amqpMessage.application_properties[item.key] = val;
+            if (item.key === DT_SUBJECT) {
```
BertKleewein, line 127: >if (item.key === DT_SUBJECT) { [](http://example.com/codeflow?start=12&length=30)

I don't know if this matters, but `ensureApplicationPropertiesCreated()` gets called even if there is a single `dt-subject` properties (and thus no application properties.) Do you want to move the call to `ensureApplicationPropertiesCreated` down below where it actually sets the property?

olivakar (approved):

device/transport/mqtt/src/mqtt.ts
```diff
@@ -1002,7 +1003,9 @@ export class Mqtt extends EventEmitter implements DeviceTransport {
     if (message.properties && message.properties.count() > 0) {
       for (let i = 0; i < message.properties.count(); i++) {
         if (i > 0 || sysPropString) topic += '&';
-        topic += encodeURIComponent(message.properties.propertyList[i].key) + '=' + encodeURIComponent(message.properties.propertyList[i].value);
+        const keyName = message.properties.propertyList[i].key;
+        topic += encodeURIComponent(((keyName === DT_SUBJECT) ? ('$.sub') : (keyName)))
```
BertKleewein, line 1007: >'$.sub') [](http://example.com/codeflow?start=65&length=8)

It's strange that `$.sub` is treated differently than `$.cid`, etc., above.

BertKleewein (approved):
:shipit:

common/transport/amqp/src/amqp_message.ts
anthonyvercolano: Done

device/transport/mqtt/src/mqtt.ts
anthonyvercolano: These things happen over time.  Lot of this code in here could probably be refactored to use more modern apis.

PR #1180, opened 2023-03-09
Fix flaky retry operation unit test

vishnureddy17:


common/core/test/_retry_operation_test.js
```diff
@@ -26,9 +26,9 @@ describe('RetryOperation', function () {
       testOperation.retry(function (callback) {
         actualOperation(callback);
       }, function (finalErr) {
-        assert.isBelow(actualOperation.callCount, 3);
-        assert.isBelow(testPolicy.nextRetryTimeout.callCount, 3);
-        assert.isBelow(testPolicy.shouldRetry.callCount, 3);
+        assert.isBelow(actualOperation.callCount, 4);
+        assert.isBelow(testPolicy.nextRetryTimeout.callCount, 4);
+        assert.isBelow(testPolicy.shouldRetry.callCount, 4);
```
vishnureddy17, line 31: Sometimes it manages to get called 3 times and when that happens the test fails. Bumping up this number should make the test less flaky. 

As an aside, I'm not sure if we should even be having this test becuause these tests really depend on execution time which can make things flaky. However, that is out of the scope of this PR.

cartertinney: I like to inject blocking to deal with these kind of execution time problems

cartertinney (approved):

PR #1181, opened 2023-04-05
fix(azure-iot-device-mqtt): add naive D2C message size checking

vishnureddy17:
This adds naive D2C message size checking. This won't catch every message that hits the size limit because properties (including system properties) count toward the limit. However, the formula for this seems a tricky to derive from simply testing against a hub.

### DIFF
--- a/device/transport/mqtt/src/mqtt.ts
+++ b/device/transport/mqtt/src/mqtt.ts
@@ -277,6 +277,12 @@ export class Mqtt extends EventEmitter implements DeviceTransport {
               topic += '/';
             }
 
+            // This will not catch all messages that exceed IoT Hub limits because properties contribute the size as well.
+            if ((message?.data?.length ?? 0) > 256 * 1024) {
+              sendEventCallback(new errors.MessageTooLargeError('Message size is greater than 256KiB'));
+              return;
+            }
+
             /*Codes_SRS_NODE_COMMON_MQTT_BASE_16_010: [** The `sendEvent` method shall use QoS level of 1.]*/
             this._mqtt.publish(topic, message.data, { qos: 1, retain: false }, (err, result) => {
               if (err) {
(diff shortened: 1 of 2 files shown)


device/transport/mqtt/src/mqtt.ts
```diff
@@ -277,6 +277,12 @@ export class Mqtt extends EventEmitter implements DeviceTransport {
               topic += '/';
             }
 
+            // This will not catch all messages that exceed IoT Hub limits because properties contribute the size as well.
```
BertKleewein, line 280: >properties contribute the size as well. [](http://example.com/codeflow?start=83&length=39)

You can account for properties by adding len(topic). Then you're checking everything except the MQTT packet overhead (which might be a constant size)

device/transport/mqtt/src/mqtt.ts
```diff
@@ -277,6 +277,12 @@ export class Mqtt extends EventEmitter implements DeviceTransport {
...
+            if ((message?.data?.length ?? 0) > 256 * 1024) {
```
BertKleewein, line 281: >message?.data?.length  [](http://example.com/codeflow?start=17&length=22)

`message.data` is typed as `any`, but not all objects have a `length` attribute. JSON objects don't,, but I can't see anywhere below here that would serialize an object. 

Regardless, `publish()` in mqtt-packet has code to handle Buffer objects when calculating packet length.  You may or may not want to replicate that code here.

BertKleewein (approved):
:shipit:

device/transport/mqtt/src/mqtt.ts
vishnureddy17: Nope, it's more complicated than that. From my testing, everything up to and including `/events/` in the topic string doesn't count. In the key-value pairs in the topic string for properties, the `=` characters don't count. I think that's enough to know that the service calculation isn't so simple.

device/transport/mqtt/src/mqtt.ts
vishnureddy17: message.data must be a string or Buffer according to the MQTT.js typing. I think whoever wrote the types here made a mistake.

PR #1182, opened 2023-04-05
log error if outgoing circular buffer is full

BertKleewein:


common/transport/amqp/test/_sender_link_test.js
```diff
@@ -75,6 +75,7 @@ describe('SenderLink', function () {
       fakeRheaLink.sendable = sinon.stub();
       fakeRheaLink.sendable.onSecondCall().returns(true);
       fakeRheaLink.sendable.returns(false);
+      fakeRheaLink.has_credit = sinon.stub().returns(true)
```
vishnureddy17, line 78: nit: why add this stub if the tests don't use it?

vishnureddy17 (approved):

common/transport/amqp/test/_sender_link_test.js
BertKleewein: Because the test calls send() on the transport and that requires this method to exist on the sender link.

PR #1184, opened 2023-04-27
chore: add module tests backand remove unused tests

BertKleewein:
Two .js files weren't being called. `file_upload_advanced.js` had had obviously never passed (the code was passing an integer into a `deviceClient` parameter).

module tests (`phase0_modules`) had previously been removed for being flakey.  I am unable to get these tests to fail locally and am re-enabling them so they can run inside our devops pipelines and flag any errors that might still be there.


vishnureddy17 (dismissed):

e2etests/test/configurations.js
```diff
@@ -13,6 +13,13 @@ const getErrorDetailString = require('./testUtils').getErrorDetailString;
 const hubConnectionString = process.env.IOTHUB_CONNECTION_STRING;
 const registry = Registry.fromConnectionString(hubConnectionString);
 
+
+function waitForEventualConsistency(callback) {
+  // Some of these tests occasionally fails, presumably because the registry doesn't update quickly
+  // enough. Sleep for an arbitrary 2 seconds to account for this.
+  setTimeout(() => callback(), 2000);
```
vishnureddy17, line 20: Do we have a guarantee that it's eventually consistent by 2 seconds, or is this just some number you chose that seemed to work?

vishnureddy17 (approved):
nit: put a delay function in somewhere like testUtils.js instead of repeating the same pattern in different places.

e2etests/test/configurations.js
BertKleewein: 2 seconds is totally arbitrary.

PR #1196, opened 2023-07-21
fix: don't send error string from azure storage on the wire

vishnureddy17:


device/core/src/blob_upload/blob_upload_result.ts
```diff
@@ -52,7 +52,7 @@ export class BlobUploadResult {
     let uploadResult: BlobUploadResult;
     if (err) {
       const statusCode = ((Object.prototype.hasOwnProperty.call(err, 'statusCode')) ? (err.statusCode) : (-1));
-      const statusDescription = ((Object.prototype.hasOwnProperty.call(err, 'response')) ? (err.response) : (err.message));
+      const statusDescription = "failed to upload to blob";
```
minune-msft, line 55: consider concatenating the StatusCode here.

vishnureddy17: done @minune-msft

PR #1198, opened 2023-11-27
fix(e2etests): fix tests that check for object equality

vishnureddy17:


lev-i (approved):
Looks good

vishnureddy17:
force pushed an empty commit to trigger pipelines, github had issues yesterday and that seems to have put this pr in a weird state.

vishnureddy17:
I don't remember what the purpose of this PR was. I will close it as it is stale.

PR #1234, opened 2025-08-20
Have forceDisconnect as an option

olivakar:
When experiencing message loss due to high throughput, the force flag in disconnectClient can be set to true to cause messages to be properly dropped and thus re-sent. 

Original ICM ID 567171345

Now default is False otherwise set to True


avishekpant (approved):

BertKleewein (approved):
:shipit:

